PR #2, opened 2020-10-13
HAI-20: Github Actions -ci-skripti

umeetiusbaar:


muhassin (approved):
Building the artefact seems to be now in two places. This is known, and needs to considered at some point again.

PR #39, opened 2020-11-12
HAI-202 Initial Liquibase and half-start for JPA

muhassin:
The renaming of HankeEntity to OldHankeEntity was just to allow old code to continue working as before, without having to do lots of refactoring at this point.  The OldHankeEntity will of course be dropped as part of next steps.

### DIFF
--- a/services/hanke-service/build.gradle.kts
+++ b/services/hanke-service/build.gradle.kts
@@ -43,7 +43,10 @@ plugins {
 	id("org.springframework.boot") version "2.3.4.RELEASE"
 	id("io.spring.dependency-management") version "1.0.10.RELEASE"
 	kotlin("jvm") version "1.4.10"
+	// Gives kotlin-allopen, which auto-opens classes with certain annotations
 	kotlin("plugin.spring") version "1.4.10"
+	// Gives kotlin-noarg for @Entity, @Embeddable
+	kotlin("plugin.jpa") version "1.4.10"
 	idea
 }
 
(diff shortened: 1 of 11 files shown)


services/hanke-service/src/main/resources/application.properties
```diff
@@ -11,3 +21,13 @@ management.endpoint.health.probes.enabled=liveness,readiness
 info.java-version = ${java.specification.version}
 info.java-vendor = ${java.specification.vendor}
 server.forward-headers-strategy=FRAMEWORK
+
+# LIQUIBASE
+# For Liquibase, until we can finalize the DataSource setup. Once that is done, Liquibase should
+# be able to use the (primary) data source of Spring Boot directly.
+# TODO: If testing will at some point need database "haitaton_test", that will need to handled
+# with e.g. Spring profiles
+spring.liquibase.url=jdbc:postgresql://${HAITATON_HOST:localhost}:${HAITATON_PORT:5432}/haitaton
```
umeetiusbaar, line 30: These will be removed in future and liquibase will use spring.datasource properties?

umeetiusbaar (approved):

services/hanke-service/src/main/resources/application.properties
umeetiusbaar: Yes, now I noticed your comment above that.

services/hanke-service/build.gradle.kts
```diff
@@ -54,8 +57,14 @@ dependencies {
 	implementation("io.github.microutils:kotlin-logging:1.12.0")
 	implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
 	implementation("de.grundid.opendatalab:geojson-jackson:$geoJsonJacksonVersion")
-    runtimeOnly("org.postgresql:postgresql")
+	implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+	implementation("org.liquibase:liquibase-core")								// TODO: do we need org.springframework:spring-dao ?
+
+	runtimeOnly("org.postgresql:postgresql")
+	// H2 is used as embedded db for some simple low level Entity and Repository class testing
+	runtimeOnly("com.h2database:h2")
```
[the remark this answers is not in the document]
muhassin: Would require not using the (at)DataJpaTest annotation, and thus would require using the normal datasource (i.e. postgresql), which is at least order of magnitude slower (and would require that such service would be available in the build environment, which apparently is not the case yet).

PR #48, opened 2020-11-23
HAI-161 enum for vaihe and few small bugfixes

muhassin:


tintotti (approved):
Good change to use enum. Seems like good validation rules to be added for the update functionality.

PR #52, opened 2020-11-27
HAI-248 contact information

tintotti:
Saving, updating, fetching HankeYhteystiedot and small tests for it.
Missing some validation steps and ContantType needs to be removed from Rest API interface.


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -78,11 +90,15 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
         logger.info {
             "Saving  Hanke step 2 for: ${entity.hankeTunnus}"
         }
+
```
muhassin, line 93: Add a comment here, about needing a transaction around these (or the whole request-reponse chain). Maybe the framework is already doing it for us (or can do it if configured so), but we haven't checked that yet.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -59,11 +69,13 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
         // Special fields; handled "manually".. TODO: see if some framework could handle (some of) these for us automatically
         entity.version = 0
         // TODO: will need proper stuff derived from the logged in user.
-        entity.createdByUserId =  1
+        entity.createdByUserId = 1
         entity.createdAt = getCurrentTimeUTCAsLocalTime()
         entity.modifiedByUserId = null
         entity.modifiedAt = null
 
+        //    entity.listOfHankeYhteystieto =
```
muhassin, line 77: TODO? or leftover?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -108,19 +125,41 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
         logger.info {
             "Saved  Hanke ${hanke.hankeTunnus}"
         }
+        // yhteystiedot
+
```
muhassin, line 129: And transaction wrapping check/add later here, too.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
```diff
@@ -129,4 +132,42 @@ class HankeControllerITests(@Autowired val mockMvc: MockMvc) {
...
+        //HankeYhteystieto Omistaja added
+        hankeToBeMocked.omistajat = arrayListOf(
+                HankeYhteystieto(null, ContactType.TOTEUTTAJA, "Pekkanen", "Pekka",
+                        "pekka@pekka.fi", "3212312", null,
+                        "Kaivuri ja mies", null, null, null,
+                        null, null))
+
+        //changing some return values
+        val expectedHanke = hankeToBeMocked
+                .apply { hankeTunnus = mockedHankeTunnus }
```
umeetiusbaar, line 151: "apply" does not need to be chained like this - just one is enough and all these separate clauses can be inside it

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -40,6 +46,10 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
 
         // TODO: Find out all savetype matches and return the more recent draft vs. submit.
         val entity = hankeRepository.findByHankeTunnus(hankeTunnus) ?: throw HankeNotFoundException(hankeTunnus)
+
+        val listOfHankeYhteystiedot = entity.id?.let { hankeYhteystiedotRepository.findByHankeId(it) };
```
umeetiusbaar, line 53: ';' sneaked in ;-)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -191,6 +251,74 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
...
+                        hankeYht.etunimi,
+                        hankeYht.email,
+                        hankeYht.puhelinnumero,
+                        hankeYht.organisaatioId,
+                        hankeYht.organisaatioNimi,
+                        hankeYht.osasto,
+                        1, //TODO: real user
+                        hankeYht.createdAt?.toLocalDateTime(),
+                        null, //TODO: real user
+                        getCurrentTimeUTCAsLocalTime(),  //TODO: if changed?
```
muhassin, line 286: I left these created and modified -things out of the entity-creation, specifically so that the calling method (create vs. update) can choose what to do. However, for the main Hanke, there is only one entity, so it is easy. Here one request could end up having 10+ Yhteystieto-entities, so it might be easier to just give the desired created/modified values as parameters to the function call. Especially since we do not want to resolve the user-value 10+ times during the same request.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -157,6 +197,27 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
             return h
         }
 
+        /**
+         * createSeparateYhteystietolistsFromEntityData splits entity's one list to three different contact information
+         * lists and adds them for Hanke domain object
+         */
+        private fun createSeparateYhteystietolistsFromEntityData(hanke: Hanke, hankeEntity: HankeEntity) {
+            var it = hankeEntity.listOfHankeYhteystieto?.iterator()
```
umeetiusbaar: maybe something like.. 
```
hankeEntity.listOfHankeYhteystieto?.forEach { hankeYhteysEntity ->
  var hankeYhteystieto = createHankeYhteystietoDomainObjectFromEntity(hankeYhteysEntity)
...
}
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -191,6 +251,74 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
...
+
+            addHankeYhteystietoEntitysToList(h.omistajat, hankeEntity)
+            addHankeYhteystietoEntitysToList(h.arvioijat, hankeEntity)
+            addHankeYhteystietoEntitysToList(h.toteuttajat, hankeEntity)
         }
+
+        private fun addHankeYhteystietoEntitysToList(listOfHankeYhteystiedot: List<HankeYhteystieto>, hankeEntity: HankeEntity) {
+            val iterator = listOfHankeYhteystiedot.iterator()
+
+            while (iterator.hasNext()) {
```
umeetiusbaar, line 273: In here also something like
```
listOfHankeYhteystiedot.forEach { hankeYht ->
...
}
```
Or even like this:
```
hankeEntity.listOfHankeYhteystieto!!.addAll(
  listOfHankeYhteystiedot.map { hankeYht ->
    HankeYhteystietoEntity(
      hankeYht.contactType,
      ...
    )
  }
)
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Persistence.kt
```diff
@@ -114,7 +115,13 @@ class HankeEntity (
         // can be a performance problem if there is a need to do bulk inserts.
         // Using SEQUENCE would allow getting multiple ids more efficiently.
         @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
-        var id: Int? = null
+
```
muhassin, line 118: No extra newline between annotations and related field

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -33,11 +35,12 @@ data class Hanke(
         var modifiedAt: ZonedDateTime?,
 
         // Default for machine API's. UI should always give the save type.
+
```
muhassin, line 38: And no extra newline between a comment that is specific for the next code line.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeYhteystietoEntity.kt
```diff
@@ -0,0 +1,45 @@
...
+enum class ContactType {
+    OMISTAJA, //owner
+    ARVIOIJA, //planner or person to do the planning of hanke
+    TOTEUTTAJA // implementator or builder
+}
+
+
+@Entity
+@Table(name = "hankeyhteystieto")
+class HankeYhteystietoEntity (
```
umeetiusbaar, line 15: All these fields are not really necessary inside constructor. Are any of them needed inside constructor? Maybe for testing some are but not all?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -33,11 +35,12 @@ data class Hanke(
...
         var saveType: SaveType? = SaveType.SUBMIT) {
 
-    var listOfOmistaja: List<HankeYhteystiedot> = arrayListOf()
-    var listOfArvioija: List<HankeYhteystiedot> = arrayListOf()
-    var listOfToteuttaja: List<HankeYhteystiedot> = arrayListOf()
+    var omistajat: MutableList<HankeYhteystieto> = arrayListOf()
```
umeetiusbaar, line 41: These lists can be initialized more simply:
```
var omistajat = mutableListOf<HankeYhteystieto>()
```

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeControllerTest.kt
```diff
@@ -94,29 +94,38 @@ class HankeControllerTest {
...
                 version = 1, createdBy = "Tiina", createdAt = getCurrentTimeUTC(), modifiedBy = null, modifiedAt = null, saveType = SaveType.DRAFT)
 
+
+        hanke.omistajat = arrayListOf(
+                HankeYhteystieto(null, ContactType.TOTEUTTAJA, "Pekkanen", "Pekka",
+                        "pekka@pekka.fi", "3212312", null,
+                        "Kaivuri ja mies", null, null, null,
+                        null, null))
+
+        var mockedHanke = hanke.copy()
```
umeetiusbaar, line 112: What is the meaning of this `mockedHanke`?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
tintotti: Left over. Removed

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
tintotti: This is a bit tricky in many ways. Do we want to update the modifiedat and by also when they maybe did not change? We need to come back to this when we have the User and we are actually implementing update functionality by another user.

I'll add a TODO-comment for it

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Persistence.kt
tintotti: Yes. I guess it came from merge.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
tintotti: Thanks.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeYhteystietoEntity.kt
tintotti: Hmm.. this was how the example seemed to be written by Markku. If the instance is created usually it will contain these. 
Is there a good reason why NOT put them to constructor?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeControllerTest.kt
tintotti: I want to add the ids for the returned object but not have them in the content that is sent for actual logic.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
tintotti: Fixed all lists in Hanke

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeYhteystietoEntity.kt
umeetiusbaar: Well.. if we want to create this kind of object, it would need all the fields (without default value) to be filled also. And these are Entity classes meaning that JPA/Hibernate will construct them without parameters and use setters to populate it (afaik). In principle, the constructor should only hold the fields that are mandatory in order to construct the object. If all these are mandatory, then it is okay. But maybe they are not all mandatory and some of them could be moved outside constructor? Those with default values for example?

muhassin: My previous example should not be used as a template; I had some issues while trying to create the initial entity, so it ended up like that, at the time. Indeed, probably many of the constructor parameters should be moved out to body also in HankeEntity.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -139,11 +175,12 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
                     hankeEntity.modifiedByUserId?.toString(),
                     if (hankeEntity.modifiedAt != null) ZonedDateTime.of(hankeEntity.modifiedAt, TZ_UTC) else null,
 
-                    hankeEntity.saveType
+                    hankeEntity.saveType,
```
muhassin, line 178: extra comma

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -158,39 +195,119 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
...
+        // Assuming the incoming date, while being zoned date and time, is in UTC and time value can be simply dropped here.
+        // Note, .toLocalDate() does not do any time zone conversion.
+        hanke.alkuPvm?.let { entity.alkuPvm = hanke.alkuPvm?.toLocalDate() }
+        hanke.loppuPvm?.let { entity.loppuPvm = hanke.loppuPvm?.toLocalDate() }
+        hanke.vaihe?.let { entity.vaihe = hanke.vaihe }
+        hanke.suunnitteluVaihe?.let { entity.suunnitteluVaihe = hanke.suunnitteluVaihe }
+
+        hanke.saveType?.let { entity.saveType = hanke.saveType }
+        hanke.tyomaaKatuosoite?.let { entity.tyomaaKatuosoite = hanke.tyomaaKatuosoite }
+        hanke.tyomaaTyyppi.let { entity.tyomaaTyyppi = hanke.tyomaaTyyppi }
```
muhassin, line 263: Maybe: (the "let" has no use for tyyppi at this point, and tyomaakoko can (I think) still be null).
entity.tyomaaTyyppi = hanke.tyomaaTyyppi
hanke.tyomaaKoko?.let { entity.tyomaaKoko = hanke.tyomaaKoko }

tintotti: I just fixed things that gradle build was vaining about. If there is underlying problem in some other way, it is best if you fix it?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
tintotti: Fixed like that

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -158,39 +195,119 @@ class HankeServiceImpl (@Autowired val hankeRepository: HankeRepository) : Hanke
...
+                    etunimi = hankeYhteystietoEntity.etunimi,
+                    sukunimi = hankeYhteystietoEntity.sukunimi,
+                    email = hankeYhteystietoEntity.email,
+                    puhelinnumero = hankeYhteystietoEntity.puhelinnumero,
+                    organisaatioId = hankeYhteystietoEntity.organisaatioid,
+                    organisaatioNimi = hankeYhteystietoEntity.organisaationimi,
+                    osasto = hankeYhteystietoEntity.osasto,
+                    createdBy = hankeYhteystietoEntity.createdByUserId?.toString() ?: "",
+                    modifiedBy = hankeYhteystietoEntity.modifiedByUserId?.toString() ?: "",
+                    createdAt = createdAt,
```
muhassin, line 236: Shorter, simpler:
if (hankeYhteystietoEntity.createdAt != null) ZonedDateTime.of(hankeYhteystietoEntity.modifiedAt, TZ_UTC) else null,

and same for modifiedAt..

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeYhteystietoEntity.kt
tintotti: We don't have clear understanding of what is mandatory fields and what not :( I would maybe do this later if we have time... If there is no clear "you shall not do this because there will be massive bad consequences" rule why this should be changed immediately?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
tintotti: Well this would be the one reason not to have all the fields in the constructor.. but since they are and this is in fact contructor initialization, the example you have there won't work as it is.

muhassin (approved):

PR #53, opened 2020-11-27
HAI-208 - geometrian tallennus

umeetiusbaar:
# Description

Saving HankeGeometriat coming from UI into the PostgreSQL/PostGIS database using direct JDBC/SQL approach instead of JPA/Hibernate (for speed of development at this point).

### Jira Issue: 
https://helsinkisolutionoffice.atlassian.net/browse/HAI-208

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Regular unit and integration tests with gradle:
- HankeGeometriatServiceImplTest (unit test)
- HankeGeometriatDaoImplITest (integration test using [TestContainers](https://www.testcontainers.org/))
- HankeGeometriatServiceImplITest (integration test using [TestContainers](https://www.testcontainers.org/))

Manual testing also possible with e.g. docker-compose BUT IT NEEDS A LITTLE TWEAKING ATM: After the first page on UI a Hanke row is stored into the database. It has a service-generated `hanketunnus`. In UI on the other hand, when drawing the geometries and clicking "Tallenna geometria" it sends the message with a hard-coded `hanketunnus` `1234567`. So, after going to the map page in UI and before clicking "Tallenna geometria" one has to manually update the database (we assume there is only one Hanke at the database at this point):
```
UPDATE Hanke
SET hanketunnus = '1234567'
WHERE id = 1;
```
After this one can save geometries in UI. But the results have to be checked also directly in the database:
```
SELECT * FROM HankeGeometriat;
SELECT id, hankeGeometriatId, ST_AsGeoJson(geometria), parametrit FROM HankeGeometria;
```
There should be only one row in `HankeGeometriat` and as many rows and one draws geometry shapes in UI in `HankeGeometriat`. Every time one clicks "Tallenna geometria" old `HankeGeometria` rows are deleted and new ones created (therefore the `id` of the rows increase).

# Checklist:

 or other location: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/1068236801/Geometriat (VERY SIMPLE)

# Other relevant info

## Use of Spring Profiles
The main `@Configuration` class `fi.hel.haitaton.hanke.Configuration` uses profile "default" which is the default (duh) when Spring Boot Application starts. This contains all the Service/DAO beans which are no longer annotated `@Service`or such in order to prevent them to coming into tests when not needed. For integration tests is a separete `fi.hel.haitaton.hanke.IntegrationTestConfiguration` that defines all beans as mocks (using mockk library). This configuration uses profile "itest". Now, in a integration test that needs mocks two annotation are needed (yes, they both are needed):
```
@Import(IntegrationTestConfiguration::class)
@ActiveProfiles("itest")
```
And if the test does not need mocks - e.g. `HankeGeometriatServiceImplITest` which uses database and all - only one annotation is needed (yes, it is still needed even though it is "default" because when running e.g. integration tests Spring automatically uses profile "integrationTest" instead of "default" - this might be something to consider in the future: how to benefit from this default test profile picking of Spring. But for now, there is the "itest" profile):
```
@ActiveProfiles("default")
```

## TestContainers
I chose to use [TestContainers](https://www.testcontainers.org/)  which launch a database instance inside docker container for the tests. The usage is quite simple. In e.g. HankeGeometriatServiceImplITest I have this:
```
@Testcontainers
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@ActiveProfiles("default")
internal class HankeGeometriatServiceImplITest {

    companion object {
        @Container
        var container: HaitatonPostgreSQLContainer = HaitatonPostgreSQLContainer
                .withExposedPorts(5433) // use non-default port
                .withPassword("test")
                .withUsername("test")

        @JvmStatic
        @DynamicPropertySource
        fun postgresqlProperties(registry: DynamicPropertyRegistry) {
            registry.add("spring.datasource.url", container::getJdbcUrl)
            registry.add("spring.datasource.username", container::getUsername)
            registry.add("spring.datasource.password", container::getPassword)
            registry.add("spring.liquibase.url", container::getJdbcUrl)
            registry.add("spring.liquibase.user", container::getUsername)
            registry.add("spring.liquibase.password", container::getPassword)
        }
    }
...
```
`@TestContainers`, `@Container` and `@DynamicPropertySource` (with kotlin's `@JvmStatic`) are needed and some configuration. `HaitatonPostgreSQLContainer` is a singleton in order to launch the database only once for all tests (I haven't actually checked if this is really the case). Using TestContainers does not seem overly slow so I think it is a good way to test against a real database. One should just remember to configure it to use other port than what docker-possu or some other PostgreSQL running on same host uses.

## Postgis extension in Docker PostgreSQL
I moved the extension into liquibase script because it is needed per database and Docker database initialization script had it for "postgres" database where it already was as default. Also, it is needed for testing with TestContainers so liquibase is a better place for it.

## Use of SQL
I chose to use SQL in both Liquibase script and also in DAO kotlin code. This was for me much easier than trying to figure out how JPA/Hibernate could work with PostGIS.


muhassin:
For the case of "HaitatonPostgreSQLContainer is a singleton in order to launch the database only once for all tests (I haven't actually checked if this is really the case)."
Could the "@TestInstance(Lifecycle.PER_CLASS)" perhaps work on that integration test? I used it on the HelloApiController unit test as an example, but I haven't checked if it works on integration tests as expected.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/IntegrationTestConfiguration.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Profile
+import org.springframework.jdbc.core.JdbcOperations
+
+@Configuration
+@Profile("itest")
+class IntegrationTestConfiguration {
+
+    @Bean
+    fun jdbcOpeations(): JdbcOperations = mockk()
```
muhassin, line 16: typo with "Opeations"

services/hanke-service/src/main/resources/db/changelog/changesets/create-hankegeometriat-tables.sql
```diff
@@ -0,0 +1,68 @@
...
+CREATE EXTENSION IF NOT EXISTS postgis;
+
+CREATE TABLE IF NOT EXISTS HankeGeometriat
+(
+    id              integer GENERATED BY DEFAULT AS IDENTITY,
+    hankeId         integer,
+    version         integer,
+    createdByUserId integer,
+    createdAt       timestamp,
+    updatedByUserId integer,
```
muhassin, line 13: "updated" vs. "modified". This has gone two ways at some point, see:
https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/928809010/Sovittuja+kehitysk+yt+nteit

services/hanke-service/src/main/resources/db/changelog/changesets/create-hankegeometriat-tables.sql
```diff
@@ -0,0 +1,68 @@
+--liquibase formatted sql
+--changeset teemu:create-hankegeometry-tables comment:Create Hanke Geometry tables
+
+CREATE EXTENSION IF NOT EXISTS postgis;
+
+CREATE TABLE IF NOT EXISTS HankeGeometriat
```
muhassin, line 6: IF NOT EXISTS  is dangerous. This could e.g. in some situations lead to old geometry tables in the database, while Liquibase thinks the changeset was run correctly (with newer structure).  To match how the Liquibase's own changes would work, there should be no IF NO EXISTS  (except maybe on that postgis extension thing, since that is simply on/off).
To be polite about it, one could add some kind of pre-check, but I don't know how to make that into a pure SQL changeset file.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Persistence.kt
```diff
@@ -134,8 +134,20 @@ class HankeEntity (
         var meluHaitta: Haitta13? = null
         var polyHaitta: Haitta13? = null
         var tarinaHaitta: Haitta13? = null
-}
 
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is HankeEntity) return false
+
+        if (hankeTunnus != other.hankeTunnus) return false
```
muhassin, line 142: Hmm.. must be careful here.  From e.g. database point of view, there can be (currently/soon) up to 3 different rows each with the same hankeTunnus. Also, there could be a row with no hankeTunnus (for autosave-purposes in future).
I do not know which all parties will be using these equals() and hashCode(), so I'd check also fields "id" and "saveType" to be on the safe side.

services/hanke-service/src/integrationTest/resources/fi/hel/haitaton/hanke/hankeGeometriat.json
```diff
@@ -35,6 +35,6 @@
     ]
   },
   "version": 1,
-  "createdAt": "2020-11-09T15:53:23.1234567+02:00[Europe/Helsinki]",
-  "updatedAt": "2020-11-09T16:53:23.1234567+02:00[Europe/Helsinki]"
+  "createdAt": "2020-11-09T15:53:23.123+02:00[Europe/Helsinki]",
```
muhassin, line 38: Wasn't the time zones supposed to be UTC to/from front (and usually better to stick to UTC in database, too, just to avoid the hazzle of daylight savings messups; calculation logic in between doesn't care as long as timezones are either implicitly same or known).
Like Hanke-API is now passing back ""createdAt": "2020-11-20T09:55:08.2493325Z"... though, maybe with less sub-second digits.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hello/HelloApiControllerITests.kt
```diff
@@ -1,13 +1,16 @@
 package fi.hel.haitaton.hanke.hello
 
+import com.ninjasquad.springmockk.MockkBean
```
muhassin, line 3: Three first imports seem to be unused.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImplITest.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    @Autowired
+    private lateinit var hankeGeometriatDao: HankeGeometriatDao
+
+    @Autowired
+    private lateinit var jdbcOperations: JdbcOperations
+
+    @Test
+    fun `CRUD testing`() {
+        val hankeGeometriat = OBJECT_MAPPER.readValue(Files.readString(Paths.get("src/integrationTest/resources/fi/hel/haitaton/hanke/hankeGeometriat.json")), HankeGeometriat::class.java)
```
muhassin, line 62: In order to not mix project directory hierarchy into the code/classpath hiearchy, following can be used:
`val hankeGeometriat = OBJECT_MAPPER.readValue(this::class.java.getResource("/fi/hel/haitaton/hanke/hankeGeometriat.json").readText(Charsets.UTF_8), HankeGeometriat::class.java)`
Alternately, if the json-file is moved to the same package (as it should, ...hanke.geometria/hankeGeometriat.json), it can be made shorter like:
`val hankeGeometriat = OBJECT_MAPPER.readValue(this::class.java.getResource("hankeGeometriat.json").readText(Charsets.UTF_8), HankeGeometriat::class.java)`
(I tested both using the HankeGeometriatServiceImplTest)

muhassin (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImpl.kt
```diff
@@ -0,0 +1,155 @@
...
+                        parametrit
+                    FROM
+                        HankeGeometria
+                    WHERE
+                        hankeGeometriatId = ?
+                """.trimIndent(), { rs, _ ->
+                    val geojson = rs.getString(1)
+                    val paramjson = rs.getString(2)
+                    hankeGeometriat.featureCollection!!.features.add(Feature().apply {
+                        geometry = OBJECT_MAPPER.readValue(geojson)
```
tintotti: This is a bit hard to read. Could it be split into smaller (maybe even separately testable?) pieces.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImpl.kt
```diff
@@ -0,0 +1,155 @@
...
+                hankeId,
+                version,
+                createdByUserId,
+                createdAt,
+                modifiedByUserId,
+                modifiedAt
+            FROM HankeGeometriat WHERE hankeId = ?            
+        """.trimIndent(), { rs, _ ->
+                HankeGeometriat(
+                        rs.getInt(1),
```
tintotti, line 120: Should we use local constants for readability... numbered "hihavakio"s are always a bit ugly.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImpl.kt
```diff
@@ -0,0 +1,155 @@
...
+                        rs.getInt(2),
+                        null,
+                        rs.getInt(3),
+                        rs.getInt(4),
+                        rs.getTimestamp(5).toInstant().atZone(TZ_UTC),
+                        rs.getInt(6),
+                        rs.getTimestamp(7).toInstant().atZone(TZ_UTC)
+                )
+            }, hankeId).getOrNull(0)
```
tintotti, line 129: Whaat? What are we going here? :D

tintotti (commented):
Some observations to be considered

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImpl.kt
muhassin: Those .getInt(1) etc, are pretty standard way to access the results.  The alternate version gets quite verbose.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriaControllerITests.kt
```diff
@@ -1,102 +1,106 @@
...
                 .andExpect(MockMvcResultMatchers.jsonPath("$.errorCode").value("HAI1001"))
 
-        verify { hankeGeometriaService.saveGeometria(hankeId, any()) }
+        verify { hankeGeometriatService.saveGeometriat(hankeTunnus, any()) }
     }
 
     @Test
     fun `create Geometria with service failure`() {
-        val hankeGeometriat = OBJECT_MAPPER.readValue(Files.readString(Paths.get("src/integrationTest/resources/fi/hel/haitaton/hanke/hankeGeometriat.json")), HankeGeometriat::class.java)
+        val hankeGeometriat = "/fi/hel/haitaton/hanke/hankeGeometriat.json".asJsonResource(HankeGeometriat::class.java)
```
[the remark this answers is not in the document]
umeetiusbaar: functions starting with "to" (e.g. toString()) or "as" (asList) are quite common

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImpl.kt
umeetiusbaar: what do you mean by "constants"? rs.getInt(1)? They are not constants, they are indexes of ResultSet values.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImpl.kt
umeetiusbaar: what do you mean?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatDaoImpl.kt
umeetiusbaar: I splitted reading the HankeGeometria rows into separate function

tintotti (approved):

PR #61, opened 2020-12-04
HAI-247 and HAI-249 tests and changed yhteystieto persistence

muhassin:
yhteystieto persistence changed to use Cascade.ALL and not using own Repository.
Fixed outgoing yhteystieto createdAt-field getting value from saved modifiedAt field.
Changed yhteystieto audit-fields to use internal info only, and along with cascade-all usage, updates to fields are copied to existing entity objects, instead of creating new ones.

Effects are contained mostly internally into the HankeServiceImpl-class.
Tested manually in swagger that the persistency and yhteystietohandling code changes should not affect REST API.
Did not cause changes to database.

Integration tests run.

TODO: some more tests to service and controller -levels.

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
@@ -255,7 +255,7 @@ class HankeServiceImpl(private val hankeRepository: HankeRepository) : HankeServ
         // preserve createdBy and createdAt field values without having to rely on the client-side to hold
         // the values for us (bad design), which would also require checks on those (to prevent tampering).
 
-        // Create temporary id-to-existingyhteystieto -map
+        // Create temporary id-to-existingyhteystieto -map, used to find quickly whether an incoming YT is new or exists already.
         val existingYTs: MutableMap<Int, HankeYhteystietoEntity> = mutableMapOf()
         for (existingYT in entity.listOfHankeYhteystieto) {
             if (existingYT.id == null) {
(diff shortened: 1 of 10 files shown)


umeetiusbaar (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -271,38 +245,97 @@ class HankeServiceImpl(private val hankeRepository: HankeRepository,
...
+
+        for (hankeYht in listOfHankeYhteystiedot) {
+            // Is the incoming YT new (does not have id, create new) or old (has id, update existing):
+            if (hankeYht.id == null) {
+                // New YT, check if it is fully empty; if anything set, create new one.
+                // Note, validator should have enforced that at least the four main fields are all set (or none of them set).
+                if (hankeYht.sukunimi.isNotBlank() || hankeYht.etunimi.isNotBlank()
+                        || hankeYht.email.isNotBlank() || hankeYht.puhelinnumero.isNotBlank()
+                        || !hankeYht.organisaatioNimi.isNullOrBlank() || !hankeYht.osasto.isNullOrBlank()) {
+                    val hankeYhtEntity = HankeYhteystietoEntity(
```
tintotti, line 293: This is now allowing partially OK data to go to database. We need all 4 of them. Otherwise some other level data check has failed.
Osasto and organisaatioNimi can be blank for private citizen. I don't understand why did you add rules for them.

muhassin: This will need a bigger discussion about the logic on these (and also on other) validations; if this part works in any other way, there will too many state combinations for an easy condition expression here.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -271,39 +245,160 @@ class HankeServiceImpl(private val hankeRepository: HankeRepository,
...
+            //   would not be sent to backend as a group of ""-fields, so the condition here is to make
+            //   such fully-empty entry considered as non-existent (i.e. skip it).
+            //   Also, for now, if anything is given, it is checked that all 4 mandatory fields are given, or we log an error and skip it.
+            // Note, validator should have enforced that at least the four main fields are all set (or all of them are empty).
+            // TODO: Currently validator does allow through an entry with four mandatory fields empty, but organisation field(s) non-empty!
+            // If anything field is given (not empty and not only whitespace)...
+            if (hankeYht.sukunimi.isNotBlank() || hankeYht.etunimi.isNotBlank()
+                    || hankeYht.email.isNotBlank() || hankeYht.puhelinnumero.isNotBlank()
+                    || !hankeYht.organisaatioNimi.isNullOrBlank() || !hankeYht.osasto.isNullOrBlank()) {
+                someFieldsSet = true
```
tintotti, line 309: This seems to do the trick. It might have been a place for a isYTValid() method even..? But consider yourself if you want to do that... This is hopefully temporary code.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -271,39 +245,160 @@ class HankeServiceImpl(private val hankeRepository: HankeRepository,
...
-                        hankeYht.osasto,
-                        1, //TODO: real user , make sure you don't update by another user but only once
-                        hankeYht.createdAt?.toLocalDateTime(),
-                        null, //TODO: real user , make sure updated only if real changes? or what is the design decision?
-                        getCurrentTimeUTCAsLocalTime(),  //TODO: only if changed?  do we always want to change the date? how do we know has subobject really been updated?
-                        hankeYht.id,
-                        hankeEntity)
-
-                hankeEntity.listOfHankeYhteystieto!!.add(hankeYhtEntity)
+    private fun processUpdateYT(hankeYht: HankeYhteystieto, existingYTs: MutableMap<Int, HankeYhteystietoEntity>,
```
tintotti, line 357: We might loose some readability when we shorten Yhteystieto to YT in method name.. :)

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeControllerTest.kt
```diff
@@ -129,4 +129,9 @@ class HankeControllerTest {
         Assertions.assertThat(responseHanke.body?.nimi).isEqualTo("hankkeen nimi")
     }
 
+    // TODO: test that checks that tyomaa- and haitat-fields can be sent to controller and come back in the response
```
tintotti, line 132: Are we going to do this? Isn't this the part which is under implementation in the original Jira item? So maybe either implement the test or remove the TODO comment? You can of course push this to dev and continue, if otherwise the set is OK.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
muhassin: Yes, depending on how the controller validations and/or frontend handling of empty fields could change, these may still change. However, that "some field not empty" -situation is something we may need to be ready to handle always (one way or another). I'll add separate methods for both condition groups, make it more readable.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeControllerTest.kt
muhassin: Yes, still going to do it, as part of the ongoing ticket. The PR was just getting too big.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
muhassin: Will change. The "YT" in the method is a left-over; the method names were much longer earlier, and shortening the YT improved readability at that time.

PR #62, opened 2020-12-09
HAI-247 more tests and configuration fix for one test

muhassin:
Note especially the HankeServiceITests, it had bad configuration, causing stuff to go into the real database. Oops.  For some reason it can not find the service bean with the DataJpaTest (even if adding Service-annotation to the service), so the explicit Import-annotation helps with that.

The addition of equals() and hashcode() to YhteystietoEntity is to ensure the collection operations will work correctly (they worked correctly before, because the existing use case needed only the "same instance"-comparison, which the default equals() implementation did).

Adding the addYhteystieto() and removeYhteystieto() functions to HankeEntity is as per recommendations, and ensures things get done correctly (and is more convenient to use).


umeetiusbaar (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeYhteystietoEntity.kt
```diff
@@ -41,4 +41,35 @@ class HankeYhteystietoEntity (
...
+
+        if (sukunimi != other.sukunimi) return false
+        if (etunimi != other.etunimi) return false
+        if (email != other.email) return false
+        if (puhelinnumero != other.puhelinnumero) return false
+        if (organisaatioId != other.organisaatioId) return false
+        if (organisaatioNimi != other.organisaatioNimi) return false
+        if (osasto != other.osasto) return false
+
+        return true
```
tintotti, line 60: Id is the one which tells if there are the same.

muhassin: Between new entries, id is null on all of them. However, id comparison is still needed here (for updates), and contactType is missing (required). Coming...

tintotti (approved):

PR #63, opened 2020-12-10
HAI-247 and HAI-249 more tests and couple test fixes

muhassin:
and small (Kotlin style) changes as suggested by Idea, to reduce the number of warnings.

The datetime changes are there to be prepared for possible "can not start before this moment" validations, i.e. ensuring the start date is always in future, and end date is the same or later than that.


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeControllerTest.kt
```diff
@@ -46,21 +48,21 @@ class HankeControllerTest {
...
-                        getCurrentTimeUTC(), getCurrentTimeUTC(), Vaihe.OHJELMOINTI, null,
+                        getDatetimeAlku(), getDatetimeLoppu(), Vaihe.OHJELMOINTI, null,
                         1, "Risto", getCurrentTimeUTC(), null, null, SaveType.DRAFT))
 
         val response: ResponseEntity<Any> = hankeController.getHankeByTunnus(mockedHankeTunnus)
 
         Assertions.assertThat(response.statusCode).isEqualTo(HttpStatus.OK)
         Assertions.assertThat(response.body).isNotNull
-        Assertions.assertThat((response.body as Hanke).nimi).isNotEmpty()
+        Assertions.assertThat((response.body as Hanke).nimi).isNotEmpty
```
tintotti, line 58: This gives an error in my environment. Maybe ( ) should be returned to avoid problems in other environments too?

tintotti (approved):
If you return the problematic ( ) the HankeControllerTest class otherwise seems to be OK. 

Some tests I only checked cursively.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeControllerTest.kt
muhassin: Returning () with a comment so that they don't get accidentally removed again.

PR #66, opened 2020-12-22
HAI-427 backend for hanke list: get all hanke items from database

tintotti:


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeControllerTest.kt
```diff
@@ -59,6 +59,36 @@ class HankeControllerTest {
...
+                    getDatetimeAlku(), getDatetimeLoppu(), Vaihe.SUUNNITTELU, SuunnitteluVaihe.KATUSUUNNITTELU_TAI_ALUEVARAUS,
+                    1, "Paavo", getCurrentTimeUTC(), null, null, SaveType.SUBMIT)
+        )
+
+        Mockito.`when`(hankeService.loadListOfHanke()).thenReturn(listOfHanke)
+
+        val response: ResponseEntity<Any> = hankeController.getAllHankeItems()
+
+        //basic checks for getting a response
+        Assertions.assertThat(response.statusCode).isEqualTo(HttpStatus.OK)
```
umeetiusbaar, line 81: How about getting rid of "Assertions.assertThat" here and import it instead and use just "assertThat"?

tintotti: Ah.. I again forgot this clearer way.. :)

umeetiusbaar (approved):

PR #72, opened 2021-01-12
Hai 430 hankkeet kartalla rajapinta

umeetiusbaar:
* /hankkeet?geometry=true (possibly with other search parameters as well) includes geometry in results
* Added Hanke.geometriat
* Added HankeSearch.geometry
* NOTICE TODO in HankeController.getHankeList for moving search logic into service instead of controller - I think it would be better in service (checking what kind of search it is and returning results accordingly).

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
@@ -2,8 +2,11 @@ package fi.hel.haitaton.hanke
 
 import fi.hel.haitaton.hanke.domain.Hanke
 import fi.hel.haitaton.hanke.domain.HankeYhteystieto
+import fi.hel.haitaton.hanke.geometria.HankeGeometriat
+import fi.hel.haitaton.hanke.geometria.HankeGeometriatService
 import io.mockk.every
 import io.mockk.verify
+import org.geojson.FeatureCollection
 import org.junit.jupiter.api.Test
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest
@@ -34,6 +37,9 @@ class HankeControllerITests(@Autowired val mockMvc: MockMvc) {
     @Autowired
     lateinit var hankeService: HankeService  //faking these calls
 
+    @Autowired
+    lateinit var hankeGeometriatService: HankeGeometriatService
+
     @Test
     fun `When hankeTunnus is given then return Hanke with it (GET)`() {
 
@@ -53,7 +59,7 @@ class HankeControllerITests(@Autowired val mockMvc: MockMvc) {
     }
 
     @Test
-    fun `When calling get without parameters then return all Hanke data`() {
+    fun `When calling get without parameters then return all Hanke data without geometry`() {
 
         // faking the service call with two returned Hanke
         every { hankeService.loadAllHanke() }.returns(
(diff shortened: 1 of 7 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
```diff
@@ -53,7 +59,7 @@ class HankeControllerITests(@Autowired val mockMvc: MockMvc) {
     }
 
     @Test
-    fun `When calling get without parameters then return all Hanke data`() {
+    fun `When calling get without parameters then return all Hanke data without geometry`() {
```
tintotti, line 62: Good addition for naming!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -80,23 +75,9 @@ class HankeController(@Autowired private val hankeService: HankeService) {
...
-            ResponseEntity.status(HttpStatus.OK).body(hankeList)
-
-        } catch (e: Exception) {
-            logger.error(e) {
-                HankeError.HAI1004.toString()
-            }
-            ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(HankeError.HAI1004)
+    private fun includeGeometry(hankeList: List<Hanke>) {
+        hankeList.forEach { hanke ->
+            hanke.geometriat = hankeGeometriatService.loadGeometriat(hanke.hankeTunnus!!)
```
tintotti, line 80: Did you consider getting all geometrys in one call? Now service is called separately for all hanke. If this is thought process then OK :)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -47,31 +51,22 @@ class HankeController(@Autowired private val hankeService: HankeService) {
...
-
-        if (hankeSearch == null || hankeSearch.periodBegin == null || hankeSearch.periodEnd == null ) {
-            return getAllHankeItems()
-        }
         return try {
-            //  Get all hanke datas within time period (= either or both of alkuPvm and loppuPvm are inside the requested period)
-            // TODO: user token  from front?
-            //  TODO: do we limit result for user own hanke?
-            val hankeList = hankeService.loadAllHankeBetweenDates(hankeSearch.periodBegin!!, hankeSearch.periodEnd!!)
+            // TODO it might be better to move this logic with different search types into service and keep controller as clean and lean as possible
```
tintotti, line 55: That is not bad idea. Should have thought that when controller started to get more complex.

tintotti (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
umeetiusbaar: I did not consider changing into that. But I did change what you suggested elsewhere about loading geometries without also loading the Hanke first (as was done for the geometry loading with only hankeTunnus before).
Loading geometries for multiple Hanke at once would need more changes into database queries etc. Something for future development perhaps...

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatServiceImpl.kt
```diff
@@ -3,6 +3,7 @@ package fi.hel.haitaton.hanke.geometria
 import fi.hel.haitaton.hanke.HankeNotFoundException
 import fi.hel.haitaton.hanke.HankeRepository
 import fi.hel.haitaton.hanke.TZ_UTC
+import fi.hel.haitaton.hanke.domain.Hanke
```
muhassin, line 6: unused import?

umeetiusbaar: fixed

muhassin (approved):

PR #87, opened 2021-02-04
HAI-558: Sentry hanke-servicelle

umeetiusbaar:


services/hanke-service/src/main/resources/application.properties
```diff
@@ -38,4 +38,14 @@ spring.liquibase.user=${HAITATON_USER:haitaton_user}
 spring.liquibase.password=${HAITATON_PASSWORD:haitaton}
 logging.level.liquibase = INFO
 
-
+# SENTRY
+sentry.dsn=https://a4148023aaeb4e88a37cb41e938dc076@o394401.ingest.sentry.io/5617535
```
muhassin, line 42: Add a comment explaining why revealing this dsn here is not a mistake (i.e. it will be public in the users' browsers anyway).

muhassin (approved):

PR #90, opened 2021-02-10
Hai 635 geometriahakuja

tintotti:


tintotti:
Azure environment deploy steps failed so I forced the merge to go forward with development even though the build seems to have failed.

PR #100, opened 2021-02-15
Hai 583 tila liput backendista

muhassin:
Added state flags to Hanke.

Note, different flags are updated in different way; geometry flag is updated whenever geometry is modified (its service methods called), while certain other flags are simply recalculated when the Hanke is loaded from its service.

PR contains also some notes on possible bugs (to be checked, fixed and tested separately), and refactored certain methods around to make the logic checks clearer (and less likely to have mismatching code implementation in different places).

Integration tests finally go through, too.

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
@@ -50,9 +50,11 @@ open class HankeServiceImpl(private val hankeRepository: HankeRepository, privat
     }
 
     // WITH THIS ONE CAN AUTHORIZE ONLY THE OWNER TO LOAD A HANKE: @PostAuthorize("returnObject.createdBy == authentication.name")
-    override fun loadHanke(hankeTunnus: String): Hanke {
+    override fun loadHanke(hankeTunnus: String): Hanke? {
         // TODO: Find out all savetype matches and return the more recent draft vs. submit.
-        val entity = hankeRepository.findByHankeTunnus(hankeTunnus) ?: throw HankeNotFoundException(hankeTunnus)
+        val entity = hankeRepository.findByHankeTunnus(hankeTunnus)
+        if (entity == null)
+            return null
 
         if (entity.id == null) {
             throw DatabaseStateException(hankeTunnus)
@@ -108,6 +110,9 @@ open class HankeServiceImpl(private val hankeRepository: HankeRepository, privat
         val entity = HankeEntity()
         copyNonNullHankeFieldsToEntity(hanke, entity)
         copyYhteystietosToEntity(hanke, entity, userid)
+        // NOTE: flags are NOT copied from incoming data, as they are set by internal logic.
+        hanke.updateStateFlagTiedotLiikHaittaIndeksille()
+        hanke.updateStateFlagOnKaikkiPakollisetLuontiTiedot()
         // Special fields; handled "manually".. TODO: see if some framework could handle (some of) these for us automatically
         entity.version = 0
         entity.createdByUserId = userid
@@ -143,6 +148,9 @@ open class HankeServiceImpl(private val hankeRepository: HankeRepository, privat
         // Transfer field values from domain object to entity object, and set relevant audit fields:
         copyNonNullHankeFieldsToEntity(hanke, entity)
         copyYhteystietosToEntity(hanke, entity, userid)
+        // NOTE: flags are NOT copied from incoming data, as they are set by internal logic.
+        hanke.updateStateFlagTiedotLiikHaittaIndeksille()
+        hanke.updateStateFlagOnKaikkiPakollisetLuontiTiedot()
         // Special fields; handled "manually".. TODO: see if some framework could handle (some of) these for us automatically
         entity.version = entity.version?.inc() ?: 1
         // (Not changing createdBy/At fields.)
@@ -162,6 +170,20 @@ open class HankeServiceImpl(private val hankeRepository: HankeRepository, privat
         return createHankeDomainObjectFromEntity(entity)
     }
 
+    override fun updateHankeStateFlags(hanke: Hanke) {
+        if (hanke.hankeTunnus == null)
+            error("Somehow got here with hanke without hanke-tunnus")
+        // Both checks that the hanke already exists, and get its old fields to transfer data into
+        val entity = hankeRepository.findByHankeTunnus(hanke.hankeTunnus!!)
+                ?: throw HankeNotFoundException(hanke.hankeTunnus)
+        copyStateFlagsToEntity(hanke, entity)
+        hankeRepository.save(entity)
+    }
+
+
+    // ======================================================================
+
+    // --------------- Helpers for data transfer from database ------------
 
     companion object Converters {
         internal fun createHankeDomainObjectFromEntity(hankeEntity: HankeEntity): Hanke {
@@ -181,8 +203,11 @@ open class HankeServiceImpl(private val hankeRepository: HankeRepository, privat
                     //   Do it below, outside this construction call.
                     hankeEntity.createdByUserId ?: "",
                     // From UTC without timezone info to UTC with timezone info
-                    if (hankeEntity.createdAt != null) ZonedDateTime.of(hankeEntity.createdAt, TZ_UTC) else null, hankeEntity.modifiedByUserId,
-                    if (hankeEntity.modifiedAt != null) ZonedDateTime.of(hankeEntity.modifiedAt, TZ_UTC) else null, hankeEntity.saveType
+                    if (hankeEntity.createdAt != null) ZonedDateTime.of(hankeEntity.createdAt, TZ_UTC) else null,
+                    hankeEntity.modifiedByUserId,
+                    if (hankeEntity.modifiedAt != null) ZonedDateTime.of(hankeEntity.modifiedAt, TZ_UTC) else null,
+
+                    hankeEntity.saveType
             )
             createSeparateYhteystietolistsFromEntityData(h, hankeEntity)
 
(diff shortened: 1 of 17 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -181,8 +199,11 @@ open class HankeServiceImpl(private val hankeRepository: HankeRepository, privat
                     //   Do it below, outside this construction call.
                     hankeEntity.createdByUserId ?: "",
                     // From UTC without timezone info to UTC with timezone info
-                    if (hankeEntity.createdAt != null) ZonedDateTime.of(hankeEntity.createdAt, TZ_UTC) else null, hankeEntity.modifiedByUserId,
-                    if (hankeEntity.modifiedAt != null) ZonedDateTime.of(hankeEntity.modifiedAt, TZ_UTC) else null, hankeEntity.saveType
+                    if (hankeEntity.createdAt != null) ZonedDateTime.of(hankeEntity.createdAt, TZ_UTC) else null,
+                    hankeEntity.modifiedByUserId,
+                    if (hankeEntity.modifiedAt != null) ZonedDateTime.of(hankeEntity.modifiedAt, TZ_UTC) else null,
+
```
tintotti, line 205: Why extra row?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -56,5 +64,38 @@ data class Hanke(
...
+        // All mandatory fields have been given... (though their validity should be checked elsewhere)
+        //  and saveType is submit, not just draft?
+        tilat.onKaikkiPakollisetLuontiTiedot = !nimi.isNullOrBlank()
+                && !kuvaus.isNullOrBlank()
+                && (alkuPvm != null) && (loppuPvm != null)
+                && (vaihe != null) && (vaihe != Vaihe.SUUNNITTELU || suunnitteluVaihe != null)
+                && !tyomaaKatuosoite.isNullOrBlank()
+                && (kaistaHaitta != null) && (kaistaPituusHaitta != null)
+                && tilat.onGeometrioita == true
+                && saveType == SaveType.SUBMIT
```
tintotti, line 90: Why this rule?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -56,5 +64,38 @@ data class Hanke(
...
+        updateStateFlagTiedotLiikHaittaIndeksille()
+    }
+
+    fun updateStateFlagOnKaikkiPakollisetLuontiTiedot() {
+        // All mandatory fields have been given... (though their validity should be checked elsewhere)
+        //  and saveType is submit, not just draft?
+        tilat.onKaikkiPakollisetLuontiTiedot = !nimi.isNullOrBlank()
+                && !kuvaus.isNullOrBlank()
+                && (alkuPvm != null) && (loppuPvm != null)
+                && (vaihe != null) && (vaihe != Vaihe.SUUNNITTELU || suunnitteluVaihe != null)
```
tintotti, line 86: I would prefer a bit human readable expression here... maybe even separater method to answer if "vaihe" checks are OK? ...because this no needs paper and pen and some drawing to understand :)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -56,5 +64,38 @@ data class Hanke(
...
+                && !tyomaaKatuosoite.isNullOrBlank()
```
tintotti, line 87: Have you checked all these from Pihla? I don't think this is necessary for doing liikennehaittaindeksi

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatServiceImpl.kt
```diff
@@ -62,4 +68,11 @@ open class HankeGeometriatServiceImpl(
         hankeGeometriat?.includeHankeProperties(hanke)
         return hankeGeometriat
     }
+
+    private fun isGeometryNonEmpty(hankeGeometriat: HankeGeometriat): Boolean {
+        // TODO: might check deeper, and if so, there could be some multi-thread unsafe things hidden (though mostly theoretical)...
+        if (hankeGeometriat.featureCollection?.features.isNullOrEmpty()) return false
+        return true
+    }
```
tintotti, line 76: Is hankegeometriatId enough for the check? Not sure, but isn't it created only after we have successful insert to database?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/HankeValidator.kt
```diff
@@ -98,17 +98,11 @@ class HankeValidator : ConstraintValidator<ValidHanke, Hanke> {
...
-                // TODO: is that property node correct?
-                // TODO: Does not currently matter, though, as the node information does not get through to error response.
-                context.buildConstraintViolationWithTemplate(HankeError.HAI1002.toString()).addPropertyNode("YhteysTiedot").addConstraintViolation()
-                ok1 = false
-            }
+        if (yhteystieto.isAnyMandatoryFieldSet() && !yhteystieto.isValid()) {
+            // TODO: is that property node correct?
+            // TODO: Does not currently matter, though, as the node information does not get through to error response.
+            context.buildConstraintViolationWithTemplate(HankeError.HAI1002.toString()).addPropertyNode("YhteysTiedot").addConstraintViolation()
+            ok1 = false
```
tintotti, line 105: I'm not sure how this was part of this "flag" task... But are you now making sure this works with the current front solution and is not stricter than front?

services/hanke-service/src/main/resources/db/changelog/changesets/add-hanke-flag-fields.yml
```diff
@@ -0,0 +1,12 @@
...
+      id: add-hanke-flag-fields
+      comment: Add new fields for various state/flag info e.g. does it have geometries or have the traffic index values been calculated and still valid
+      author: Markku Hassinen
+      changes:
+        - addColumn:
+            tableName: hanke
+            columns:
+              - column: { name: tilaongeometrioita, type: boolean }
+              - column: { name: tilaonviereisiahankkeita, type: boolean }
+              - column: { name: tilaonasiakasryhmia, type: boolean }
```
tintotti, line 12: I think having these in database is good. Checking them from real situation might cost too much so having the flag to check them seems reasonable.

tintotti (approved):
Some questions for you to consider if they make sense and need attention.

Maybe in the HankeControllerITests.kt there should test for the new flags so that they are correctly shown in API?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
muhassin: To be consistent with the local style of one row = one parameter; the two params when in the end of the row (where they ended during one of Teemu's PRs, probably due to autoformatting) could be easily missed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
muhassin: Will do something. I have to admit, it is a bit unclear even for me, takes a moment or two to convince myself the logic works.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
muhassin: Pihla was uncertain about that field.  I'll comment it for now..

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
muhassin: I thought we were already using that submit mode (in the end of the creation "wizard"), and thus, that prevents starting index analysis with something that is still in draft state, even if all the fields have been saved (with possibly placeholder values).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/HankeGeometriatServiceImpl.kt
muhassin: Not sure. Not really sure if that is enough either. I will happily let someone who knows the geometry side better to improve that function.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/HankeValidator.kt
muhassin: I was looking at the various related logics when calculating the flags, and also added helper functions to another data classes, so, for improved clarity and consistency, I relocated that logic to relevant data class, too.  The logic itself was not changed, only relocated (exactly to avoid having to test it).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
muhassin: Unless you referred to the empty line; which is just to group the audit-fields from other fields.

umeetiusbaar (approved):

PR #114, opened 2021-02-25
Hai-688 tormays api

tintotti:
Draft to see

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
@@ -11,6 +11,7 @@ import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestMapping
+
 private val logger = KotlinLogging.logger { }
 
 @RestController
@@ -21,7 +22,7 @@ class TormaystarkasteluController(@Autowired private val laskentaService: Tormay
     @GetMapping("/{hankeTunnus}/tormaystarkastelu")
     fun getTormaysTarkastelu(@PathVariable(name = "hankeTunnus") hankeTunnus: String?): ResponseEntity<Any> {
         logger.info {
-            "Creating tormaystarkastelu for hanke: $hankeTunnus"
+            "Fetching existing tormaystarkastelu for hanke: $hankeTunnus"
         }
 
         if (hankeTunnus == null) {
@@ -31,9 +32,13 @@ class TormaystarkasteluController(@Autowired private val laskentaService: Tormay
         return try {
             // call service to get tormaystarkastelu
             val tormaysResults = laskentaService.getTormaystarkastelu(hankeTunnus)
-
-            logger.info { "tormaystarkastelu created for Hanke $hankeTunnus." }
-            ResponseEntity.status(HttpStatus.OK).body(tormaysResults)
+            if (tormaysResults == null) {
+                logger.info { "Tormaystarkastelu does not exist for Hanke $hankeTunnus." }
+                ResponseEntity.status(HttpStatus.NOT_FOUND).body(HankeError.HAI1007)
+            } else {
+                logger.info { "Tormaystarkastelu fetched for Hanke $hankeTunnus." }
+                ResponseEntity.status(HttpStatus.OK).body(tormaysResults)
+            }
         } catch (e: Exception) {
             logger.error(e) { HankeError.HAI1006.toString() }
             ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(HankeError.HAI1006)
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
@@ -1,0 +1,98 @@
+package fi.hel.haitaton.hanke.tormaystarkastelu
+
+import fi.hel.haitaton.hanke.HankeError
+import fi.hel.haitaton.hanke.domain.Hanke
+import fi.hel.haitaton.hanke.validation.ValidHanke
+import mu.KotlinLogging
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.validation.annotation.Validated
+import org.springframework.web.bind.annotation.*
+
+
+private val logger = KotlinLogging.logger { }
+
+@RestController
+@RequestMapping("/tormays")
+@Validated
+class TormaystarkasteluController {
+
+    @GetMapping("/{hankeTunnus}")
+    fun getTormaysTarkastelu(@PathVariable(name = "hankeTunnus") hankeTunnus: String?): ResponseEntity<Any> {
+        logger.info {
+            "Creating tormaystarkastelu for hanke: ${hankeTunnus}"
+        }
+
+        if (hankeTunnus == null) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(HankeError.HAI1002) //TODO: own error code?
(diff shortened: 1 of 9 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
```diff
@@ -0,0 +1,106 @@
...
+                HankeError.HAI1006.toString()
+
+            }
+            ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(HankeError.HAI1006)
+        }
+    }
+
+    private fun getDummyTormaystarkasteluTulos(): TormaystarkasteluTulos {
+        val tormaysResults = TormaystarkasteluTulos()
+        tormaysResults.hankeId = 2 //TODO: selvitettävä jostain
```
muhassin, line 52: Outgoing messages should have at least hanketunnus, not only the id.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
```diff
@@ -0,0 +1,106 @@
...
+        if (hanke == null) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(HankeError.HAI1002)
+        }
+
+        return try {
+
+            //if(hanke.tilat.onTiedotLiikenneHaittaIndeksille)
+            // then calculate:
+            // - call service to create tormaystarkastelu -> you will get TormaystarkasteluTulos
+            // - service has saved tormaystulos to database and saved hanke.onLiikenneHaittaIndeksi=true (or is it services job to do that?)
```
muhassin, line 81: One takes a Hanke-object from the HankeService (if not already have it from there, to ensure the flag states are current), manipulates the flag(s) as desired, then calls HankeService.updateHankeStateFlags(hanke).
See begin of HankeGeometriatServiceImpl.saveGeometriat() for an example how it is done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
```diff
@@ -0,0 +1,106 @@
...
+        tormaysResults.liikennehaittaIndeksi!!.indeksi = 4.2
+        tormaysResults.liikennehaittaIndeksi!!.type = IndeksiType.PYORAILYINDEKSI
+
+        //TODO: in the future this would contain "Viereiset hankkeet" too
+
+        return tormaysResults
+    }
+
+    @PostMapping
+    fun createTormaysTarkasteluForHanke(@ValidHanke @RequestBody hanke: Hanke?): ResponseEntity<Any> {
```
muhassin, line 67: The incoming data for Hanke should be ignored here; any changes that are desired to be actually saved (or used as the latest) should be saved using HankeController directly before requests to this service. Therefore, this request here would only need hanke tunnus as input.
This request should then fetch the Hanke from HankeService with that given hanke tunnus.

muhassin: Ooops, I meant the above activities to happen in Service, not in controller; see HankeGeometriatController/Service for examples, they already work in the way I tried to explain.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        return try {
+            //call service
+
+            val tormaysResults = getDummyTormaystarkasteluTulos()
+
+
+            // has saved tormaystulos to database
+            //return hanke with tormaystulos
+            logger.info { "tormaystarkastelu created for Hanke ${hankeTunnus}." }
```
umeetiusbaar, line 38: No need for `{}` in `${hankeTunnus}` here

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
```diff
@@ -0,0 +1,22 @@
+package fi.hel.haitaton.hanke.tormaystarkastelu
+
+class TormaystarkasteluTulos {
```
umeetiusbaar, line 3: Why not using data classes here? It could have only `hankeId` (and `hankeTunnus`) as constructor parameter?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
```diff
@@ -0,0 +1,22 @@
...
+    var hankeId : Int = 0
+    var liikennehaittaIndeksi: LiikennehaittaIndeksiType? = null
```
umeetiusbaar, line 5: Do we need to have this calculated result here at all? Would it be the same if we had only the three different values and the client (UI) would see from they which is the biggest and what type it is?

So, instead of this:
```
{
  "hankeId" : 1,
  "liikennehaittaIndeksi" : {
    "indeksi" : 2.3,
    "type" : "JOUKKOLIIKENNEINDEKSI"
  },
  "perusIndeksi" : 2.2,
  "pyorailyIndeksi" : 1.7,
  "joukkoliikenneIndeksi" : 2.3
}
```
there would be this:
```
{
  "hankeId" : 1,
  "perusIndeksi" : 2.2,
  "pyorailyIndeksi" : 1.7,
  "joukkoliikenneIndeksi" : 2.3
}
```
Plus perhaps `hankeTunnus` also.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluController.kt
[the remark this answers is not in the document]
umeetiusbaar: No, it would not actually call hankeService here but in service. Controller should only call a single service function with `hankeTunnus`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
```diff
@@ -0,0 +1,22 @@
...
+    var perusIndeksi: Double? = null
```
muhassin, line 6: Do these really need to be Doubles (64-bit) instead of Float (32-bit). Even Floats have about 6 digits of precision (and we're only using less than two).  Not that there is much of a penalty using Doubles, as there amount of data and calculations are quite low, but using Floats would be a better match for the use case.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
tintotti: I think that leaves too much decisions for ui? We also need the one number for Hanke to decide the color quickly in map.

muhassin: The decision for color would still be a trivial calculation, that is not a good reason. However, I also think the main index to be calculated in the backend; if backend does some calculations/decisions, it should do them all, and not split this logic between backend and fronted. (Also, I expect that the main index calculation will get more complex over time..)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
tintotti: Changed to float

umeetiusbaar (approved):

PR #115, opened 2021-02-25
HAI-628 authentication & authorization tests for controllers

muhassin:
Only tests the controller endpoints and their settings, not the actual authentication handling (JWT-token etc.) handling in front of it all. Also, if service-classes get method-level security later, these tests do not cover those.

Integration tests run.

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/security/HankeControllerSecurityTests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/security/HankeControllerSecurityTests.kt
@@ -1,0 +1,179 @@
+package fi.hel.haitaton.hanke.security
+
+import fi.hel.haitaton.hanke.HankeController
+import fi.hel.haitaton.hanke.HankeService
+import fi.hel.haitaton.hanke.IntegrationTestConfiguration
+import fi.hel.haitaton.hanke.IntegrationTestResourceServerConfig
+import fi.hel.haitaton.hanke.SaveType
+import fi.hel.haitaton.hanke.TZ_UTC
+import fi.hel.haitaton.hanke.Vaihe
+import fi.hel.haitaton.hanke.domain.Hanke
+import fi.hel.haitaton.hanke.domain.HankeSearch
+import fi.hel.haitaton.hanke.getCurrentTimeUTC
+import fi.hel.haitaton.hanke.toJsonString
+import io.mockk.every
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest
+import org.springframework.context.annotation.Import
+import org.springframework.http.MediaType
+import org.springframework.security.test.context.support.WithMockUser
+import org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors
+import org.springframework.security.test.web.servlet.response.SecurityMockMvcResultMatchers.unauthenticated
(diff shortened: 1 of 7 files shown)


muhassin:
Notes:
* IntegrationTestResourceServerConfig having TestConfiguration-annotation (should IntegrationTestConfiguration also use it?)
* Should IntegrationTestResourceServerConfig have Profile("itest") like IntegrationTestConfiguration has?
* ResourceServerConfig.configure() should be refactored to call the AccessRules.configure(), but I left that out of this PR (not part of these tests).

umeetiusbaar (approved):

muhassin:
Added adjustments for integration test configurations/annotations so that they are done the same way for all similar tests. (Related to TestConfiguration, Profile, Import)

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/security/HankeControllerSecurityTests.kt
```diff
@@ -0,0 +1,179 @@
...
+
+    @Test
+    @WithMockUser(username = "test7358", roles = ["bad role"])
+    fun `status forbidden (403) with authenticated user with bad role`() {
+        performGetHankkeet().andExpect(status().isForbidden)
+        performPostHankkeet().andExpect(status().isForbidden)
+        performPutHankkeetTunnus().andExpect(status().isForbidden)
+        performGetHankkeetTunnus().andExpect(status().isForbidden)
+    }
+
```
tintotti, line 96: Is it really testing that user is authenticated like it says in test name? Status.isOk-check is missing so I'm just wondering.. Maybe isForbidden() does that too?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/AccessRules.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.http.HttpMethod
+import org.springframework.security.config.annotation.web.builders.HttpSecurity
+
+class AccessRules {
+    companion object {
+        fun configureHttpAccessRules(http: HttpSecurity) {
+            http.anonymous().and()
+                    .authorizeRequests()
+                    .mvcMatchers(HttpMethod.GET, "/organisaatiot").permitAll()
+                    .mvcMatchers(HttpMethod.POST, "/hankkeet", "/hankkeet/**").hasRole("haitaton-user")
```
tintotti, line 12: Why do we have different access control for organisaatiot-api?

tintotti (commented):
Just couple questions.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/security/HankeControllerSecurityTests.kt
muhassin: Authenticated means the user is recognized (but does not necessarily have the required permissions to what he is requesting). Having bad role or missing role leads to unauthorized situation, which is shown as the forbidden (403) -response.
The status can be (among other things) one of ok (200), forbidden (403) or unauthorized (401, slightly incorrect term used by HTTP). That is, if the status matches isForbidden, it can not match isOk.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/AccessRules.kt
muhassin: No idea, but I do remember that we did talk about these restrictions and it was mentioned that organisaatiot-list access does not need the same tighter restrictions as the other stuff. Also, this list was copied from the main configuration, so this isn't new.. just in a new place.

PR #135, opened 2021-03-09
Hai 743 tormtark tietokanta

muhassin:
# Description

Adds Liquibase, JPA entities, Repository for törmaystarkastelu results, and changes in related existing places.

Not tested with fully running application, due to not yet connected to actual service functionality. Coming next...

### Jira Issue: 

https://helsinkisolutionoffice.atlassian.net/browse/HAI-537

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Currently the code is not yet connected to actual service-side functionality; integration tests do some tests for the repository and as a side effect, with the liquibase and entity mappings.

# Checklist:

 or other location: 
https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/1468661795/T+rm+ystarkastelun+tietokantamallit

# Other relevant info
NOTE: changed field name from English to Finnish, which may affect REST API
tormaystarkasteluTulos.liikennehaittaIndeksi.type -> .tyyppi

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
@@ -51,10 +51,6 @@ data class Hanke(
     constructor(id: Int, hankeTunnus: String) : this(id, hankeTunnus, null, null, null, null, null, null, null, null, null, null, null, null)
     constructor() : this(null, null, null, null, null, null, null, null, null, null, null, null, null, null)
 
-    // -------------- Tormaystarkastelu -------------
-    var tormaystarkasteluTulos: TormaystarkasteluTulos? = null
-    var liikennehaittaindeksi: LiikennehaittaIndeksiType? = null
-
     // --------------- Yhteystiedot -----------------
     var omistajat = mutableListOf<HankeYhteystieto>()
     var arvioijat = mutableListOf<HankeYhteystieto>()
@@ -91,6 +87,21 @@ data class Hanke(
         } else {
             null
         }
+
+    // -------------- Tormaystarkastelu -------------
+    /**
+     * Liikennehaittaindeksi is always available here if it has been calculated and is still valid.
+     * Might be around even if no longer valid.
+     * TODO: should the validity-state be part of this indeksitype, too? (I.e. would know that even
+     *  when not getting the full tulos-data (which does have that state).
+     */
+    var liikennehaittaindeksi: LiikennehaittaIndeksiType? = null
+
+    /**
+     * These are currently available only when Hanke is returned via Tormaystarkastelu controller/service.
+     * And if they have been calculated and are still valid.
+     */
+    var tormaystarkasteluTulos: TormaystarkasteluTulos? = null
 
     // --------------- State flags -------------------
     var tilat: HankeTilat = HankeTilat()
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/Persistence.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/Persistence.kt
@@ -1,6 +1,7 @@
 package fi.hel.haitaton.hanke.tormaystarkastelu
 
 import fi.hel.haitaton.hanke.HankeEntity
+import fi.hel.haitaton.hanke.getCurrentTimeUTCAsLocalTime
 import org.springframework.data.jpa.repository.JpaRepository
 import java.time.LocalDateTime
 import javax.persistence.Embedded
@@ -23,6 +24,7 @@ class TormaystarkasteluTulosEntity {
     var joukkoliikenne: Float? = null
 
     var tila: TormaystarkasteluTulosTila? = null
+    var tilaChangedAt: LocalDateTime? = null
     var createdAt: LocalDateTime? = null
 
     @Id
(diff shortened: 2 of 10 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
```diff
@@ -37,10 +42,25 @@ data class TormaystarkasteluTulos(val hankeTunnus: String) {
...
     PERUSINDEKSI,
     PYORAILYINDEKSI,
     JOUKKOLIIKENNEINDEKSI
+    // TODO: will likely need a new option "MONTA" or "USEITA".
 }
+
+enum class TormaystarkasteluTulosTila {
+    VOIMASSA,
+    EI_VOIMASSA
+}
```
umeetiusbaar, line 66: No newline at end of file?

umeetiusbaar (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -92,6 +88,21 @@ data class Hanke(
...
         }
 
+    // -------------- Tormaystarkastelu -------------
+    /**
+     * Liikennehaittaindeksi is always available here if it has been calculated and is still valid.
+     * Might be around even if no longer valid.
+     * TODO: should the validity-state be part of this indeksitype, too? (I.e. would know that even
+     *  when not getting the full tulos-data (which does have that state).
+     */
+    var liikennehaittaindeksi: LiikennehaittaIndeksiType? = null
```
tintotti, line 98: Maybe  change also here Type = Tyyppi   ?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/Persistence.kt
```diff
@@ -0,0 +1,98 @@
...
+        // For persisted entities, checking id match is enough
+        if (id != null && id != other.id) return false
+
+        // For non-persisted, have to compare almost everything
+        if (hanke != null && hanke != other.hanke) return false
+        if (liikennehaitta != other.liikennehaitta) return false
+        if (perus != other.perus) return false
+        if (pyoraily != other.pyoraily) return false
+        if (joukkoliikenne != other.joukkoliikenne) return false
+        if (tila != other.tila) return false
```
tintotti, line 73: I'm not totally sure if these are part of the object's identity and should be here... but maybe equals in this case is used so that these are also needed? 

It's just that if we happen to check "Tormaystarkastelu.equals( tormaysToBeInvalidated) "let's get out of here bacause we are not handling correct tormays" " before saving to database the new state, we get the result that we are actually handling another tormaystarkastelu and not the one we wanted.

services/hanke-service/src/main/resources/db/changelog/changesets/add-tormaystarkastelu-results.yml
```diff
@@ -0,0 +1,30 @@
...
+            columns:
+              - column: { name: id, type: int, autoIncrement: true, constraints: { primaryKey: true, nullable: false } }
+              - column: { name: hankeid, type: int, constraints: { nullable: false, foreignKeyName: fk_hanke_tormaystarkastelutulos, references: hanke(id) } }
+              - column: { name: liikennehaitta, type: float }
+              - column: { name: liikennehaittatyyppi, type: varchar(30) }
+              - column: { name: perus, type: float }
+              - column: { name: pyoraily, type: float }
+              - column: { name: joukkoliikenne, type: float }
+              - column: { name: tila, type: varchar(20) }
+              - column: { name: tilachangedat, type: timestamp }
```
tintotti, line 29: Why limit this to tila and have it differently from other database tables? I would like for us to have one way of finding created/creator and modified/modifier -type of data from tables.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
muhassin: Oops.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
muhassin: Since that is class name, it doesn't show up anywhere else, and I thought it was used as "this is a type/class"-part of the class name (since the content of that class is not only about the type of chosen index), in which case it would be ok to stay in English.  I'd rather just drop the Type-word out completely, since we don't use it in that purpose elsewhere in the code.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/Persistence.kt
muhassin: equals() is not about object's identity; it is about which information matters when two objects are to be considered equal/interchangeable. This class was a difficult case, because the future needs would dictate certain requirements, which can not be met, yet. The well-working equals is needed in certain collection operations (where the id-field might not be available).
The example check is, afaik, not needed; there are exactly 3 operations possible with these results: calculate a new one; change state later to invalid; delete old ones (when a new results is calculated). Invalidation and deletion can only be done to a result that has already been saved to database, in which case the actions on the result-objects can be pointed with the 'id' field (without using equals). The equals is needed in the Hanke-side of the relation (a collection); current implementation should use the "if (id != null && id != other.id) return false" -part of the equals() for that.

muhassin:
Merged via another branch, along with the rest of the implementation. Thus, this PR is closed.

PR #150, opened 2021-03-26
HAI-805 support for personal data processing restriction

muhassin:
# Description

Backend support for preventing changes to personal data (yhteystietos). W.I.P.

This version works (almost) as expected with integration tests, but not quite as wanted via UI. Specifically, it _does_ prevent changes to the "locked" yhteystietos, but it does _not_ prevent changes to the rest of Hanke data in the same request, and since there is no feedback in UI (in fact, UI has some trouble processing the response, leading to timeout), the behavior is not consistent.
Also, with integration tests, there will be entries in audit log (as expected) about the prevention, but no entries when used through UI.

Planned change: check and handle restrictions as a first separate step when doing hanke update, and only if there were no restrictions effects at all, continue with the normal update process (the same way as it used to go before these changes). This should keep the changes separate in such a way, that no matter how one returns from the update function, only the intended changes will be saved to database.

### Jira Issue: 

HAI-805

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing

Start creating a new hanke with one owner yhteystieto and another, e.g. arvioija yhteystieto (and save as draft). Manually set the "locked" flag for the non-owner yhteystieto in the database. Change e.g. first name on the non-owner yhteystieto in UI and save as draft... there might not be a popup about successfully saving, so wait a while, refresh the page (or return to the page via hanke list or such, forcing reload of hanke data), and check that the changed field has actually not changed.

# Checklist:

 or other location: 

# Other relevant info

The response to trying to change or delete a locked yhteystieto is currently 451; frontend might not be prepared to handle that in a sensible way.  Minor problem as long as the UI can somehow get back to show current/correct hanke data.

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeRepositoryITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeRepositoryITests.kt
@@ -168,6 +168,8 @@ class HankeRepositoryITests @Autowired constructor(
             1,
             "org1",
             "osasto1",
+            false,
+            null,
             "1",
             datetime,
             "11",
@@ -184,6 +186,8 @@ class HankeRepositoryITests @Autowired constructor(
             2,
             "org2",
             "osasto2",
+            false,
+            null,
             "2",
             datetime,
             "22",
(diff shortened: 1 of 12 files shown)


muhassin:
Closing branch for re-push (after rebasing to current dev) and new PR.

PR #203, opened 2022-09-02
HAI-1180 Refactor audit logging to use unified database table

corvidian:
# Description

Unify all existing audit logging to log to a single table. This is done
in preparation for transferring the logs to a centralized storage.

The logging principles are described at
https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/7990181897/Lokituksista

The old logs are simply deleted, since we don't have full audit logs
anyway, and we're not really in production yet. I'm still waiting on
confirmation if this is ok.

Also fix some issues that stopped the service working on a Silicon-based
mac.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Please describe tests how this change or new feature can be tested.

# Checklist:

- [X] I have written new tests (if applicable)
- [X] I have ran the tests myself (if applicable)
- [x] I have made necessary changes to the documentation, link to confluence
 or other location: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/7990181897/Lokituksista

PR #205, opened 2022-09-13
HAI-1180 Don't remove old log tables

corvidian:
# Description

When deploying #203, the deployment failed. The database migration for removing the old log tables failed because of missing privileges to the `personaldatalogs` schema.

As a temporary fix, remove the migration so that the deployment can be run. The old log tables will have to be removed later.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other


corvidian:
This wasn't need after all.

PR #212, opened 2022-09-30
HAI Fix status endpoint

corvidian:
# Description

Change the status endpoint to check for existence of any rows in the `tormays_central_business_area_polys` table, instead of checking that the table has one row.

Currently, there are 116 rows in that table at least on dev and staging environments. The check for one row fails, so the status endpoint always return 500 Internal Error. This stops us from using it as a health check endpoint. Old availability checks are calling `/organisaatiot/`, but that has required logging in for a while now, so they all fail.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Checklist:

 or other location:


esaostring (approved):
Fix seems OK.

Is merge to main intentional in this PR?

corvidian:
> Is merge to main intentional in this PR?

No it was an accident. Thank you for noticing.

PR #219, opened 2022-11-02
HAI 1176/Process HSL GTFS material to tormays polygons

esaostring:
# Description

Implemented processing of HSL GTFS to Haitaton tormays polygons.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

Prerequisite: fetch `hsl`and `hki` source materials, as per instructions in the README

Go to directory: haitaton-backend/scripts/gis-material-update/process
* (create and) activate Python virtual environment
* run `python -m unittest discover -v`

# Checklist:

 or other location: 

# Other relevant info

-


scripts/gis-material-update/process/modules/config.py
```diff
@@ -1,36 +1,111 @@
...
+            output_dir - directory for gis output files."""
+        deployment_profile = self.deployment_profile()
+        directory = self._cfg.get("storage").get(deployment_profile, {}).get(storage)
+
+        if deployment_profile == "local_development":
+            directory_name = Path(__file__).parent.parent.parent / directory
+        elif deployment_profile == "local_docker_development":
+            directory_name = Path(directory)
+        else:
+            raise ValueError("Storage type not detected!")
```
corvidian, line 61: Is there any way around checking the environment here again? It feels like the config file should already define the directories.

scripts/gis-material-update/config.yaml
```diff
@@ -28,4 +40,26 @@ polkupyora_paareitti:
...
+    host: localhost
+    username: haitaton
+    password: haitaton
+    database: haitaton
+  local_docker_development:
+    port: 5432
+    host: gis-db
+    username: haitaton
+    password: haitaton
+    database: haitaton
```
corvidian, line 57: Instead of grouping by value, like
```
key_a:
  env_a:
    ...
  env_b:
    ...

key_b:
  env_a:
    ...
  env_b:
    ...
```
Would these be better when grouped by environment?
```
env_a:
  key_a:
    ...
  key_b:
    ...
env_b:
  key_a:
    ...
  key_b:
    ...
```
This way, it's easier to see the values of a single environment.

The different environments could even be split into different files, like `config.local_development.yaml`, but then there would need to be logic for reading default values from `config.yaml`, since many values don't change between environments.

corvidian: On the other hand, when considering running this on the cloud at some point, we don't want to change the repository every time we need to change some configuration value in a cloud environment. So there should be a method to override these values from environment variables. Then we wouldn't need the different environments in this file at all. This could just hold the local development settings and config for the local docker development could be set as environment variables in the docker compose config.

scripts/gis-material-update/fetch/fetch_data.sh
```diff
@@ -34,22 +34,36 @@ cfg_layer () {
...
 hsl|osm)
     wget -O "$local_file" "$addr"
     ;;
+# plain WFS fetch
 ylre_katualue|ylre_katuosat|maka_autoliikennemaarat)
     ogr2ogr -f GPKG "$local_file" WFS:"$addr" "$layer"
     ;;
+# WFS fetch with layer renaming and where condition
+hki)
+    ogr2ogr -f GPKG "$local_file" -nln "$dest_layer" -where "$where_query" WFS:"$addr" "$layer"
```
corvidian, line 65: Could this be rewritten as:
```
ogr2ogr -f GPKG "$local_file" "$extra_params" WFS:"$addr" "$layer"
```
and configured like
```
hki:
  extra_params: "-nln alue -where kunta='091'"
...
ylre_katualue:
  extra_params: ""
...
```

Then the same case could be used for all ogr2ogr commands.

The motivation here is to reduce the number of configuration variables that are used for only some of the cases.

scripts/gis-material-update/process/modules/hsl.py
```diff
@@ -1,14 +1,381 @@
...
+    def __init__(
+        self, cfg: Config, validate_gtfs: bool = True, week_of_transit: int = 2
+    ):
+        self._cfg = cfg
+
+        # check that Helsinki city area file is available
+        if not path.exists(self._cfg.local_file("hki")):
+            raise FileNotFoundError("Helsinki city area polygon not found")
+
+        # TODO: how to obtain this string automatically?
```
corvidian, line 36: Which string is this TODO referring to?

scripts/gis-material-update/process/modules/hsl.py
```diff
@@ -1,14 +1,381 @@
...
+            for h in range(4):
+                for m in [0, 15, 30, 45]:
+                    # construct list of tuples (hour_start, minute_start, hour_end, minute_end)
+                    start_times.append((s + h, m, s + h + 1, m))
+            # pick first 13 values, according to spec commented above
+            start_times = start_times[0:13]
+            hours += start_times
+
+        hour_range = []
+        for h_s, m_s, h_e, m_e in hours:
```
corvidian, line 132: This would be a lot clearer, if the variable names were words instead of abbreviations.

scripts/gis-material-update/process/modules/hsl.py
```diff
@@ -1,14 +1,381 @@
...
+        # 9.00 - 10.00
+        #
+        # and evening rush hours (13 ranges):
+        # 15.00 - 16.00
+        # 15.15 - 16.15
+        # ...
+        # 17.45 - 18.45
+        # 18.00 - 19.00
+        for s in [6, 15]:
+            start_times = []
```
corvidian, line 122: This variable is called `start_times`, but it also holds the end times.

scripts/gis-material-update/process/modules/hsl.py
```diff
@@ -1,14 +1,381 @@
...
+            for h in range(4):
+                for m in [0, 15, 30, 45]:
+                    # construct list of tuples (hour_start, minute_start, hour_end, minute_end)
+                    start_times.append((s + h, m, s + h + 1, m))
+            # pick first 13 values, according to spec commented above
+            start_times = start_times[0:13]
```
corvidian, line 128: It's not easy for the reader to see why taking the 13 first values is necesssary and why it produces the right set of times. 

It might be easier to add the last interval separately, if using `for h in range(3):` earlier.
```
            # Add 9.00 - 10.00 or 18.00 - 19.00
            start_times.append((s + 3, 0, s + 4, 0))
```
This would also mean that the start_times variable is not needed, since the values can be appended directly to hours.

scripts/gis-material-update/process/modules/hsl.py
```diff
@@ -1,14 +1,381 @@
...
+        """Check if route is trunk route."""
+        is_trunk_route = {
+            r"^1500.*": "yes",
+            r"^2200.*": "yes",
+            r"^2510.*": "yes",
+            r"^2550.*": "yes",
+            r"^4560.*": "yes",
+            r"^1018.*": "almost",
+            r"^1039.*": "almost",
+            r"^1040.*": "almost",
```
corvidian, line 199: These are proper trunk lines these days. 18 Has been replaced by 20 and 39 by 30. There's also 570. https://www.hsl.fi/hsl/runkoverkko

If I remember correctly, the trunk lines had a different `route_type` then other HSL buses. This would be a more robust way to check for trunk lines than a list of ever-changing route ids.

scripts/gis-material-update/process/modules/hsl.py
```diff
@@ -1,14 +1,381 @@
...
+        shapes_trips_routes = shapes_with_trips.merge(
+            self.feed().routes, on="route_id", how="left"
+        )
+        shapes_trips_routes = shapes_trips_routes[~shapes_trips_routes.route_id.isna()]
+
+        # Exclude route types
+        # 0 - trams
+        # 1 - subway
+        # 4 - Suomenlinna ferries
+        # 109 - trains
```
corvidian, line 285: Would be possible to whitelist the route types for buses, i.e. the types we are interested in? In general, whitelisting is more robust than blacklisting. Like, we don't want to include the Raide-Jokeri here if they decide to use route type like "903 Regional Tram Service".

scripts/gis-material-update/fetch/fetch_data.sh
esaostring: Thought this earlier, but didn't implement it (due to difficulties in bash script end). Solved bash problems, and implemented as suggested.

scripts/gis-material-update/process/modules/config.py
esaostring: I am still struggling with this, how to (robustly) refer to directory in configuration file, several levels up.

I think I am keeping this implementation currently as-is, and will refactor it later if satisfactory solution is found.

scripts/gis-material-update/process/modules/hsl.py
esaostring: To be done in https://helsinkisolutionoffice.atlassian.net/browse/HAI-1327

scripts/gis-material-update/process/modules/hsl.py
esaostring: To be done in https://helsinkisolutionoffice.atlassian.net/browse/HAI-1327

scripts/gis-material-update/process/modules/hsl.py
esaostring: It refers to next line in code.

PR #222, opened 2022-11-10
HAI-1323 Refactor db containers in tests

corvidian:
# Description

Move the PostgreSQL container definitions to an abstract class so that they are not repeated in every test class that needs to use the database.

This has the added benefit of reusing the database container for all tests. Even after adding a script to clean the database between each test, this sped up running the integration tests from 2.5 minutes to 50 seconds on my machine. That's 1/3 of the time they used to take.

Some other, smaller, refactoring was done while touching the classes anyway, e.g. fixing the compilation warnings from
`TormaystarkasteluLaskentaServiceImplITest`.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Checklist:

 or other location:


esaostring (approved):
Looking good!

PR #253, opened 2023-01-24
HAI-1376 Prevent orphaned geometriat

corvidian:
# Description

Add a database trigger to delete geometriat of deleted hankealue. This ensures there are no orphaned geometriat left behind after a hankealue is removed.

I tried other ways of tackling this, but they proved ineffective or very complex. The removal of hankealueet is more implicit than explicit, leveraging JPA's orphan removal. The geometriat is not an JPA entity, however, so the orphan removal doesn't reach them.

One of these abandoned ways was changing how geometries are deleted in GeometriatDao(Impl). Deleting is changed to actually delete the geometriat-table row when asked to. It used to make sense to always have a geometria for a hanke, but with the introduction of hankealueet, this no longer feels like the case.

There's also some small refactoring. These were done for abandoned approaches, but seemed useful overall.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Previously, if you added two hankealue to a hanke and removed one, an orphaned row would remain in geometriat-table. This should be fixed now.
1. Create a hanke.
2. Add two hankealue.
3. Change page so they are saved.
4. Check the number of geometriat in DB: `select count(*) from geometriat ;`, note the count.
5. In the UI, remove one hankealue.
6. Check the number of geometriat in DB: `select count(*) from geometriat ;`, should be one less than before.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
@@ -22,6 +22,7 @@ import fi.hel.haitaton.hanke.test.TestUtils
 import fi.hel.haitaton.hanke.test.TestUtils.nextYear
 import java.time.format.DateTimeFormatter
 import java.time.temporal.ChronoUnit
+import javax.persistence.EntityManager
 import net.pwall.mustache.Template
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatExceptionOfType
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -838,6 +845,12 @@ class HankeServiceITests : DatabaseTest() {
         assertThat(alue.polyHaitta).isEqualTo(Haitta13.KOLME)
         assertThat(alue.tarinaHaitta).isEqualTo(Haitta13.KOLME)
         assertThat(alue.geometriat).isNotNull
+        entityManager.flush()
+        entityManager.clear()
+        val hankeFromDb = hankeRepository.getOne(updatedHanke.id!!)
+        assertThat(hankeFromDb.listOfHankeAlueet).hasSize(1)
+        assertThat(hankealueCount()).isEqualTo(1)
+        assertThat(geometriatCount()).isEqualTo(1)
```
corvidian, line 853: This failed before adding the trigger.

markohaarni (approved):
Hyvältä näyttää ja vaikuttaa toimivan kuten pitää. Varmistin tietokannasta, että geometria poistuu kun alue poistetaan.

PR #257, opened 2023-02-02
HAI-1347 Get application identifier and status after send

corvidian:
# Description

After sending the application to Allu, query for the application status and identifier, and save them to the database.

In the unlikely event that uploading the application succeeds, but the status query fails, still save the Allu ID for the application so we know it's been uploaded. The identifier and status are left null in this case.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create an application
2. Send it
3. Check in the database, that the application has `applicationidentifier` and `allustatus` defined

# Checklist:

 or other location: 

# Other relevant info
First merge HAI-1410 and this on top of that.


markohaarni (approved):
Toimii kuten pitää.

PR #259, opened 2023-02-02
HAI-1347 Rewrite getting current application status

corvidian:
# Description

We need to get the current application status from Allu to know if we can still update an application. Previously, this has been done using the application history query. This has problems, though. It requests more information than we need, since it gets the complete history from Allu. Also, there's a need to deduce the current status from the list of histories. And, perhaps most importantly, the history doesn't return anything after the application has been added for the first time. It only starts returning results after the application has been updated.

Replace the history query with the get application data query. This only returns the current status of the application and it returns the status for all applications, whether or not they've been updated.

This seems like both more efficient and more reliable.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Sending and updating applications still work.

# Checklist:

 or other location: 

# Other relevant info
First merge #257 and this on top of that.


markohaarni (approved):
Vaikuttaa hyvältä muutokselta.

PR #260, opened 2023-02-03
HAI-1457 Delete endpoint for applications

corvidian:
# Description

Add an endpoint for deleting applications. Before deleting the application, check if it's in Allu. If it is, check it's status there. If it's still pending, cancel the application. If not, refuse to delete the application.

Also improve error documentation in OpenAPI.

Hide the generic error messages from exception handlers from the OpenAPI documentation.  Every exceptionhandler in the codebase was leaking into every response everywhere.

Add documented responses for 401 and 500 errors via exception handlers. These errors can happen everywhere in the code, and this way they don't have to be repeated everywhere. Technically, there are a couple of methods which don't require login, like `/status` and `/public-hankkeet`, so 401 is not possible for them. I don't think it's a huge issue if we list 401 in their API documentation.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create an application and save it. Don't send. (http://localhost:3001/fi/johtoselvityshakemusvanha)
2. Call the delete endpoint with cURL or Postman.
3. Check from DB it was deleted.
4. Check from DB that there's a line in the audit table about the delete.
5. Create an application and save it.
6. Send it to Allu.
7. Call the delete endpoint with cURL or Postman.
8. Check from logs that it was canceled in Allu.
9. Check from DB it was deleted.
10. Check from DB that there's a line in the audit table about the delete.

I don't know if there's a way to move the application beyond pending in Allu to test that scenario.

# Checklist:

 or other location:


markohaarni (approved):
Hyvältä näyttää ja vaikuttaa toimivan kuten pitää.

PR #262, opened 2023-02-09
HAI-1454 Add support for named application areas

corvidian:
# Description

Add `areas` field to internal APIs for application data. Areas includes the area name and its geometry, which has to be a polygon. Each polygon should have the correct coordinate reference system (CRS) specified. Namely, they should use `urn:ogc:def:crs:EPSG::3879`.

The `areas` values are transformed to a `GeometryCollection` when sending the data to Allu. The names are ignored here, since Allu doesn't use them. Coordinate reference system is taken from the first area arbitrarily, but they're all checked to use the one Haitaton understands.

Converting application data from the internal API version to Allu version is refactored to a separate object. It contains more and more actual business logic, so it should be easily testable.

To make the transition seamless, support the older `geometry` field in application data for now. If both `areas` and `geometry` are present, use `areas`. `geometry` should be removed after frontend has switched to using the new `areas` field. The removal is described in task HAI-1488.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location: 

# Other relevant info
#261 should be merged before this one.


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
```diff
@@ -0,0 +1,74 @@
...
+
+object ApplicationDataMapper {
+
+    fun toAlluData(applicationData: CableReportApplicationData): AlluCableReportApplicationData =
+        AlluCableReportApplicationData(
+            applicationData.name,
+            applicationData.customerWithContacts.toAlluData("applicationData.customerWithContacts"),
+            getGeometry(applicationData),
+            applicationData.startTime
+                ?: throw AlluDataException("applicationData.startTime", "Can't be null"),
```
pitkni, line 16: The error string "Can't be null" is repeated at least three times. A constant could provide consistency and protection from typos.

services/hanke-service/src/test/resources/fi/hel/haitaton/hanke/geometria/toinen_polygoni.json
```diff
@@ -0,0 +1,33 @@
+{
```
pitkni, line 1: These polygon json files could have more descriptive names to explain themselves / express purpose of use. Feel free to ignore.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
corvidian: Yes, thank you. The exception is called with just two different error descriptions, so I created an enum for them.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapperTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        private val otherPolygon: Polygon =
+            "/fi/hel/haitaton/hanke/geometria/toinen_polygoni.json".asJsonResource()
+        private val thirdPolygon: Polygon =
+            "/fi/hel/haitaton/hanke/geometria/kolmas_polygoni.json".asJsonResource()
+
+        @Test
+        fun `uses areas if they are present`() {
+            val applicationData =
+                AlluDataFactory.createCableReportApplicationData(
+                    areas = listOf(ApplicationArea("test area", polygon)),
```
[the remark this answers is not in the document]
corvidian: Yes, thank you. Replaced with a constant.

services/hanke-service/src/test/resources/fi/hel/haitaton/hanke/geometria/toinen_polygoni.json
corvidian: I tried to think of better names, but they are just three different polygons. I thought about using the coordinates of famous places in Helsinki and names after those, but I'm not sure having names like `polygon_stadikka.json`, `polygon_bollis.json` and `polygon_norris.json` will help much here. They might just raise more questions.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/AlluDataFactory.kt
```diff
@@ -101,20 +103,22 @@ class AlluDataFactory(val applicationRepository: ApplicationRepository) {
...
-                geometry,
-                startTime,
-                endTime,
-                pendingOnClient,
-                identificationNumber,
-                clientApplicationKind,
-                workDescription,
-                contractorWithContacts,
-                false
+                applicationType = ApplicationType.CABLE_REPORT,
```
pitkni, line 109: https://sonarcloud.io/project/issues?resolved=false&severities=MAJOR&sinceLeakPeriod=true&types=CODE_SMELL&pullRequest=262&id=City-of-Helsinki_haitaton-backend&open=AYY2viW8HPUzc9biennP

Sonar complains about too many functions parameters. Would it be possible (or make sense) to wrap some of these parameters to a separate class?

pitkni (approved):

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/AlluDataFactory.kt
corvidian: I don't think it makes much sense to wrap the parameters of a factory function to an object. The function exists to set defaults to the created object.

PR #263, opened 2023-02-14
HAI-1462 API for downloading decisions from Allu

corvidian:
# Description

Add an API for downloading decisions from Allu. We ask Allu if there's a decision available to be downloaded as long as the application has an Allu ID. We don't try to be smart with the statuses.

The endpoint returns 404 if there's no application or if there's no decision for the application. The response have different codes in the body to distinguish them, if necessary.

We never save the decisions, because a collection of e.g. cable reports can be a security risk even if single reports are not.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create an application and send it to Allu.
2. Go to Allu and make a decision for the application. (https://testi.allu.kaupunkiymparisto.fi/login)
3. Call GET http://localhost:3001/api/hakemukset/22/paatos with cURL or Postman.
4. Check that you got the decision as a response.

# Checklist:

 or other location:


corvidian:
Lisäsin myös testausohjeet nyt kun osataan mennä Alluun ja tehdä siellä päätös.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
```diff
@@ -204,18 +205,28 @@ class CableReportServiceAllu(
...
+                { httpStatus -> httpStatus.value() == 404 },
+                {
+                    Mono.error(
+                        ApplicationDecisionNotFoundException(
+                            "Decision not found in Allu. alluid=$applicationId"
+                        )
+                    )
+                }
+            )
+            .bodyToMono(ByteArrayResource::class.java)
```
pitkni, line 226: Can we be sure that the request will indeed return a pdf? Would a check for content type in the response headers make any sense?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAlluITests.kt
```diff
@@ -80,15 +80,66 @@ class CableReportServiceAlluITests {
...
+        }
+
+        @Test
+        fun `throws ApplicationDecisionNotFoundException on 404`() {
+            addStubbedLoginResponse()
+            mockWebServer.enqueue(
+                MockResponse()
+                    .setResponseCode(404)
+                    .setHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+                    .setBody("Error message")
```
pitkni, line 122: This body could be a constant. Feel free to ignore.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAlluITests.kt
```diff
@@ -155,6 +202,12 @@ class CableReportServiceAlluITests {
         }
     }
 
+    fun Assert<RecordedRequest>.isValidLoginRequest() = given { loginRequest ->
```
pitkni, line 205: 👍

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
corvidian: Allu API description says it should respond with `application/pdf`. https://testi.allu.kaupunkiymparisto.fi/external/swagger-ui/index.html?urls.primaryName=v2#/Cable%20reports/getDecision_5

We tell Allu that we `Accept: application/pdf`, so it shouldn't send us anything else, at least if the response is Ok.

Then again, checking the headers doesn't really hurt us, so I added it.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAlluITests.kt
corvidian: I don't think aliasing these would make this test more readable. I changed the very generic error messages to be slightly more descriptive.

pitkni:
Manual testing done. Seems to work as expected. I'll approve.
image

pitkni (approved):

PR #264, opened 2023-02-15
HAI-1470 Send application data as a PDF attachment to Allu

corvidian:
# Description

After sending an application to Allu, create a PDF with the data as it's
represented in Haitaton and add it as an attachment to the application.

This is done so that the people in Allu can see how the data was entered
in Haitaton. The data models and concepts differ slightly between
Haitaton and Allu. Also, the attached PDF is archived for at least some
time along the application, possibly outlasting the data in Haitaton.

The attachment isn't saved to Haitaton and it isn't shown to the user in
any way.

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
@@ -25,7 +25,6 @@ import org.springframework.stereotype.Service
  * to Allu. It will be archived along the application and decision to show how the user inputted the
  * application in Haitaton, as the data models of Haitaton and Allu differ slightly.
  */
-@Service
 object ApplicationPdfService {
     private val titleFont = Font(Font.HELVETICA, 18f, Font.BOLD)
     private val sectionFont = Font(Font.HELVETICA, 15f, Font.BOLD)
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
@@ -18,7 +18,6 @@ import java.io.ByteArrayOutputStream
 import java.time.ZoneId
 import java.time.ZonedDateTime
 import java.time.format.DateTimeFormatter
-import org.springframework.stereotype.Service
 
 /**
  * Transform an application to a PDF. The PDF is added as an attachment when sending the application
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
@@ -1,0 +1,184 @@
+package fi.hel.haitaton.hanke.allu
+
+import com.lowagie.text.Chunk
+import com.lowagie.text.Document
+import com.lowagie.text.Element
+import com.lowagie.text.Font
+import com.lowagie.text.PageSize
+import com.lowagie.text.Paragraph
+import com.lowagie.text.Phrase
+import com.lowagie.text.Rectangle
+import com.lowagie.text.pdf.PdfPTable
+import com.lowagie.text.pdf.PdfWriter
+import fi.hel.haitaton.hanke.application.CableReportApplicationData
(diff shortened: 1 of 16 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
```diff
@@ -0,0 +1,184 @@
...
+import java.time.format.DateTimeFormatter
+import org.springframework.stereotype.Service
+
+/**
+ * Transform an application to a PDF. The PDF is added as an attachment when sending the application
+ * to Allu. It will be archived along the application and decision to show how the user inputted the
+ * application in Haitaton, as the data models of Haitaton and Allu differ slightly.
+ */
+@Service
+object ApplicationPdfService {
```
pitkni, line 29: It seems that other spring annotated services are classes. Is there a reason why class is not used here?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/AlluUpdateService.kt
```diff
@@ -21,7 +21,7 @@ class AlluUpdateService(
 
     internal val lockName = "alluHistoryUpdate"
 
-    @Scheduled(fixedDelay = 1000 * 60, initialDelay = 1000 * 60)
+    @Scheduled(fixedDelay = 1000 * 60, initialDelay = 1000000 * 60)
```
pitkni, line 24: Why so long initial delay? Could a constant improve some readability here? Something like:

```
private const val X_MINUTES =  value here
private const val Y_MINUTES = value here

@Service
class AlluUpdateService(
...

    @Scheduled(fixedDelay = Y_MINUTES, initialDelay = X_MINUTES)
    fun checkApplicationStatuses() {
    
```

Where x and y are the duration.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfServiceTest.kt
```diff
@@ -0,0 +1,331 @@
...
+                contains("0502222222")
+            }
+        }
+    }
+
+    private fun getPdfAsText(pdfData: ByteArray): String {
+        val reader = PdfReader(pdfData)
+        val pages = reader.numberOfPages
+        val textExtractor = PdfTextExtractor(reader)
+        File("pdfTest.pdf").writeBytes(pdfData)
```
pitkni, line 328: image

This seems to actually create a pdf in hanke-service. I imagine this is not wanted.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfService.kt
corvidian: Thank you for noticing. This isn't supposed to be `@Service` anymore. It was at some point during development, but I refactored this so that it has no dependencies, so it can be an object.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/AlluUpdateService.kt
corvidian: I changed the initial delay to get clearner logs at some in the development and accidentally committed it. I reverted this change.

I think it would be a good idea to read the values from the configuration, so they can be overridden. Outside this PR, though.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/allu/ApplicationPdfServiceTest.kt
corvidian: I used this as a way to see what the PDF looks like in different situations while developing and forgot to remove it. Removed it now. Thank you for noticing.

Clearly I didn't check my own code when creating the PR. I'll do a full read-through to see if there are any other bits that were committed by accident.

corvidian: There was one more thing: wiretapping and logging all HttpClient requests. I used at some point to verify the communication with Allu and forgot to remove it. Removed in https://github.com/City-of-Helsinki/haitaton-backend/pull/264/commits/61895f0e30daabc1087907e00748f2bcf7601048.

pitkni (approved):

PR #265, opened 2023-02-21
HAI-1496 Add name for Hanke.Hankealue

pitkni:
# Description

This commit enables adding a name (nimi) for a hankealue.  

**NB.** name length is restricted to var(100). Same as in Hanke.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Local env:
1. localhost:3001/ -> sign in as nordea demo user.
2. Create a new hanke -> copy hankeTunnus
3. GET http://localhost:3001/api/hankkeet/{hankeTunnus} 
  - see that the alueet[x].name is null
4. PUT http://localhost:3001/api/hankkeet/{hankeTunnus} 
  - add a name for hankealue in the request body
5. Verify response and the previous GET request that the hankealue has the given name.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hankealue.kt
```diff
@@ -24,4 +21,5 @@ data class Hankealue(
     var meluHaitta: Haitta13? = null,
     var polyHaitta: Haitta13? = null,
     var tarinaHaitta: Haitta13? = null,
+    var nimi: String? = null
```
corvidian, line 24: You're a bit inconsistent with the trailing commas. I think it's better to add them whenever editing the last row of a list to make future change logs cleaner.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -854,6 +854,40 @@ class HankeServiceITests : DatabaseTest() {
...
+                this.alueet = mutableListOf(hankealue.copy(nimi = "Changed Name"))
+            }
+
+        val updateHankeResult = hankeService.updateHanke(modifiedHanke)
+
+        assertThat(updateHankeResult)
+            .usingRecursiveComparison()
+            .ignoringFields("modifiedAt", "modifiedBy", "version", "alueet")
+            .isEqualTo(createdHanke)
+        assertEquals(createdHanke.alueet.size, updateHankeResult.alueet.size)
```
corvidian, line 878: Mixing `assertThat` and `assertEquals` in the same test is a bit confusing, since they have different order for expected and actual values.

The expected value here should be always 1, since you reset the alueet in the update call. I.e. here `this.alueet = mutableListOf(hankealue.copy(nimi = "Changed Name"))`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hankealue.kt
pitkni: Good point! I added trailing commas.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
pitkni: Assertion is now assertThat and expected value is 1.

corvidian: Writing this as `assertThat(createdHanke.alueet).hasSize(1)` gives a more descriptive error message if the assert fails.

There's another `assertEquals` for the nimi.

corvidian (approved):

markohaarni (approved):

PR #266, opened 2023-02-22
HAI-1470 Update applications only if changed

corvidian:
# Description

Add check to application update to see if it has actually changed. Do the related actions like updating Allu only when updating.

Only send applications if they are drafts. After the first time it's been marked as non-draft, the sending endpoint doesn't change anything anymore. changes are uploaded with the update endpoint after the application has been uploaded.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Change pages in the cable report form in UI (http://localhost:3001/fi/johtoselvityshakemus?hanke=HAI23-1). This will send update requests. See backend logs to see that nothing much is happening.

Pressing the send button on the last page of the cable report application for a second or a third time shouldn't do much on the backend side.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
@@ -73,7 +73,7 @@ class ApplicationController(
     @Operation(
         summary = "Create a new application",
         description =
-            "Creates a new application. The new application is created as a draft, " +
+            "Returns the created application. The new application is created as a draft, " +
                 "i.e. with true in pendingOnClient. The draft is not sent to Allu."
     )
     @ApiResponses(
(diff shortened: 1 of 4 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -95,12 +95,13 @@ class ApplicationController(
     @Operation(
         summary = "Update an application",
         description =
-            "Updates an application. The application can be updated until it has started " +
-                "processing on Allu, i.e. it's still pending. If the application has been " +
-                "sent to Allu, it will be updated there as well. If an Allu-update is " +
-                "required, but it fails, the local version will not be updated. The " +
-                "pendingOnClient value can't be changed with this endpoint. Use POST " +
-                "/hakemukset/{id}/send-application for that."
+            """Updates an application.
```
pitkni, line 98: First sentence may be redundant as it is already mentioned in the summary. Same thing with sending the application description.

pitkni (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
corvidian: Thank you for the suggestion. I changed the first sentences to describe the return value instead.

PR #268, opened 2023-02-24
HAI-1345 hakemus kytkenta hankkeeseen

nkala:
# Description

Creation of new applications requires existing hanke. When posting a new application `hankeTunnus`-field must be set.

### Jira Issue: 

HAI-1345

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

Make sure the corresponding frontend branch is in use when testing with frontend.

# Checklist:

 or other location: 

# Other relevant info
Automatic hanke creation is not in the scope of this PR.


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Persistence.kt
```diff
@@ -196,6 +197,9 @@ class HankeEntity(
     )
     var tormaystarkasteluTulokset: MutableList<TormaystarkasteluTulosEntity> = mutableListOf()
 
+    @OneToMany(fetch = FetchType.LAZY, mappedBy = "hanke", cascade = [])
```
pitkni, line 200: What does cascade = [] mean? No cascade is a default and it does not need to be defined at all (I think).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -65,6 +68,16 @@ class ApplicationController(
     fun getById(@PathVariable(name = "id") id: Long): Application {
         val userId = currentUserId()
         val application = service.getApplicationById(id, userId)
+        val hankeTunnus = application.hankeTunnus ?: throw HankeNotFoundException("")
+        val hankeId = hankeService.getHankeId(hankeTunnus)
+
+        if (
+            hankeId == null ||
+                !permissionService.hasPermission(hankeId, userId, PermissionCode.EDIT_APPLICATIONS)
```
corvidian, line 76: I think the permission for viewing an application should be just `VIEW`. There's no permission specifically for viewing an applications, though there could be. The current specs say that everyone with any kind of permissions has also the permission to view applications. (https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8212906098/K+ytt+oikeustasojen+toimintojen+m+rittely)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -45,6 +48,12 @@ open class ApplicationService(
             "Invalid geometry received when creating a new application for user $userId, reason = ${validationError.reason}, location = ${validationError.location}"
         }
 
+        val hanke =
+            hankeRepository.findByHankeTunnus(
+                application.hankeTunnus ?: throw HankeNotFoundException("")
```
pitkni, line 53: Could some other exception be more suitable here? In my opinion, trying to search with a null tunnus is more like a bad request than 404. Furthermore, the calling method (controller) already checks for this, so the error cannot actually happen here.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -89,7 +102,18 @@ class ApplicationController(
...
     fun create(@RequestBody application: Application): Application {
         val userId = currentUserId()
+        val hankeTunnus = application.hankeTunnus ?: throw HankeNotFoundException("")
+        val hankeId = hankeService.getHankeId(hankeTunnus)
+
+        if (
+            hankeId == null ||
+                !permissionService.hasPermission(hankeId, userId, PermissionCode.EDIT_APPLICATIONS)
+        ) {
+            throw HankeNotFoundException("")
```
pitkni, line 112: Would 403 be better if user does not have permission for this operation?

corvidian: Nope, it's better if the user can't differentiate between hanke that don't exist and hanke they don't have permission to.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
```diff
@@ -1,12 +1,10 @@
 package fi.hel.haitaton.hanke.application
 
 import com.fasterxml.jackson.databind.node.ObjectNode
-import fi.hel.haitaton.hanke.ControllerTest
-import fi.hel.haitaton.hanke.IntegrationTestConfiguration
-import fi.hel.haitaton.hanke.OBJECT_MAPPER
-import fi.hel.haitaton.hanke.andReturnBody
+import fi.hel.haitaton.hanke.*
```
corvidian, line 4: I don't think there's any reason to use wildcard imports these days. The only benefit they give is less lines of code, but IDEs hide them by default. When not using and IDE, like here in Github, it's better to see clearly where classes are imported from. I think this is further highlighted in Kotlin, where there's no difference between importing classes and importing functions and extension functions.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/Application.kt
```diff
@@ -13,5 +13,6 @@ data class Application(
     val alluStatus: ApplicationStatus?,
     val applicationIdentifier: String?,
     val applicationType: ApplicationType,
-    val applicationData: ApplicationData
+    val applicationData: ApplicationData,
+    val hankeTunnus: String?,
```
corvidian, line 17: Why is the hankeTunnus nullable? Shouldn't an application be always linked to a hanke?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
```diff
@@ -24,7 +26,16 @@ data class ApplicationEntity(
     var userId: String?,
     @Enumerated(EnumType.STRING) val applicationType: ApplicationType,
     @Type(type = "json") @Column(columnDefinition = "jsonb") var applicationData: ApplicationData,
+    @ManyToOne @JoinColumn(updatable = false, nullable = false) var hanke: HankeEntity?,
```
corvidian, line 29: This says nullable=false, but the foreign key column doesn't have NOT NULL. This is also a nullable type, so should it be nullable or not?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -280,4 +314,11 @@ class ApplicationController(
         logger.warn(ex) { ex.message }
         return HankeError.HAI2006
     }
+
+    @ExceptionHandler(HankeNotFoundException::class)
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    fun missingApplication(ex: HankeNotFoundException): HankeError {
+        logger.warn(ex) { ex.message }
+        return HankeError.HAI1001
+    }
```
corvidian, line 323: The same handler is already defined in `ControllerExceptionHandler`, which is used if there's no controller-specific handler for an exception.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -31,7 +32,9 @@ private val logger = KotlinLogging.logger {}
 @RequestMapping("/hakemukset")
 class ApplicationController(
```
corvidian, line 33: Should the other application endpoints have some access control as well?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
```diff
@@ -103,15 +103,20 @@ class ApplicationControllerITest(@Autowired override val mockMvc: MockMvc) : Con
     @Test
     @WithMockUser(username)
     fun `getById with known ID returns application`() {
+        val hankeTunnus = "HAI-1234"
```
corvidian, line 106: This value could be moved to class level to avoid repetition.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
```diff
@@ -333,6 +358,40 @@ class ApplicationControllerITest(@Autowired override val mockMvc: MockMvc) : Con
...
+            .andExpect(jsonPath("$.hankeTunnus").value(hankeTunnus))
+
+        verify { applicationService.getApplicationById(1234, username) }
+        verify { hankeService.getHankeId(hankeTunnus) }
+        verify { permissionService.hasPermission(42, any(), any()) }
+    }
+
+    @Test
+    @WithMockUser(username)
+    fun `Creating an application without hanke permissions fails`() {
```
corvidian, line 381: The other endpoints should be tested as well.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -994,7 +1068,46 @@ class ApplicationServiceITest : DatabaseTest() {
...
+        assertThrows<HankeNotFoundException> {
+            applicationService.create(
+                AlluDataFactory.createApplication(id = null, hankeTunnus = "1234"),
+                username
+            )
+        }
+    }
+
+    @Test
+    fun `hankeTunnus cannot be changed on update`() {
```
corvidian, line 1093: Should there be an update call here? Or what is this testing?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
corvidian: Well, we could return 403 if we checked that the person a VIEW permission for that hanke.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Persistence.kt
nkala: Yes it is the default. Redundancy removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
nkala: Thanks for clarification. Changed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
nkala: This is the way other endpoints handle this by not giving information about existing hankkeet. Same point point as in the next one.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
nkala: Agreed. My auto import settings for Kotlin were not configured correctly.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
nkala: Out of scope for this PR.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
nkala: Removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
nkala: Not nullable. Added the missing constraint.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
nkala: Added.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
corvidian: This exception isn't throw when a hanke is not found, it's thrown when the caller has null in `application.hanketunnus`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Persistence.kt
pitkni: I do not see any changes in this.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
nkala: Forced hankeTunnus to be non null.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
nkala: Update does not send hanke infromation again.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -28,19 +32,25 @@ private val logger = KotlinLogging.logger {}
...
+    private val applicationRepository: ApplicationRepository,
     private val alluStatusRepository: AlluStatusRepository,
     private val cableReportService: CableReportService,
     private val disclosureLogService: DisclosureLogService,
     private val applicationLoggingService: ApplicationLoggingService,
     private val geometriatDao: GeometriatDao,
+    private val permissionService: PermissionService,
+    private val hankeRepository: HankeRepository,
 ) {
+    @Transactional
```
pitkni, line 44: If I understand correctly, this method does only database reads. ->  @Transactional(readOnly = true) should provide some performance benefits.

services/hanke-service/src/main/resources/db/changelog/changesets/024-connect-application-to-hanke.yml
```diff
@@ -0,0 +1,16 @@
...
+            tableName: applications
+            columns:
+              - column:
+                  name: hanke_id
+                  remarks: "Add relating hanke to application"
+                  type: bigint
+                  constraints:
+                    foreignKeyName: fk_application_hanke
+                    references: hanke(id)
+                    nullable: false
```
pitkni, line 16: Empty line should be added to end of file.

services/hanke-service/src/main/resources/db/changelog/changesets/025-clear-existing-applications.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM applications;
```
pitkni, line 1: Empty line should be added to end of file.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
```diff
@@ -1,15 +1,10 @@
...
 import fi.hel.haitaton.hanke.allu.ApplicationStatus
-import javax.persistence.Column
-import javax.persistence.Entity
-import javax.persistence.EnumType
-import javax.persistence.Enumerated
-import javax.persistence.GeneratedValue
-import javax.persistence.GenerationType
-import javax.persistence.Id
-import javax.persistence.Table
+import javax.persistence.*
```
corvidian, line 7: This should also be expanded.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogServiceTest.kt
```diff
@@ -1,11 +1,7 @@
 package fi.hel.haitaton.hanke.logging
 
 import fi.hel.haitaton.hanke.allu.CustomerType
-import fi.hel.haitaton.hanke.application.Contact
-import fi.hel.haitaton.hanke.application.Customer
-import fi.hel.haitaton.hanke.application.CustomerWithContacts
-import fi.hel.haitaton.hanke.application.PostalAddress
-import fi.hel.haitaton.hanke.application.StreetAddress
+import fi.hel.haitaton.hanke.application.*
```
corvidian, line 4: This should also be expanded.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
```diff
@@ -24,7 +19,16 @@ data class ApplicationEntity(
...
     fun toApplication() =
-        Application(id, alluid, alluStatus, applicationIdentifier, applicationType, applicationData)
+        Application(
+            id,
+            alluid,
+            alluStatus,
+            applicationIdentifier,
+            applicationType,
+            applicationData,
+            hanke.hankeTunnus ?: throw HankeNotFoundException(""),
```
corvidian, line 32: This should throw something else. HankeNotFoundException will show up as 404 Not Found to the user. This is clearly an internal error. The hanketunnus should be created when initially saving the hanke.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -28,19 +32,25 @@ private val logger = KotlinLogging.logger {}
...
     open fun getAllApplicationsForUser(userId: String): List<Application> {
-        return repo.getAllByUserId(userId).map { it.toApplication() }
+        val result = applicationRepository.getAllByUserId(userId).toMutableSet()
+        val hankkeet =
+            permissionService.getAllowedHankeIds(userId = userId, permission = PermissionCode.VIEW)
+        hankkeet.forEach { result.addAll(hankeRepository.getOne(it).hakemukset) }
+        return result.map { it.toApplication() }
```
corvidian, line 50: I think we can only return the applications with matching permissions. They should include the ones they have created, unless they are explicitly removed from the hanke.

Also, it's more efficient to get the hankkeet in a single query.

```suggestion
        val hankeIds =
            permissionService.getAllowedHankeIds(userId = userId, permission = PermissionCode.VIEW)
        return hankeRepository
            .findAllById(hankeIds)
            .flatMap { it.hakemukset }
            .map { it.toApplication() }
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -232,24 +272,42 @@ class ApplicationServiceITest : DatabaseTest() {
         verify { cableReportServiceAllu wasNot Called }
     }
 
+    @Test
+    fun `getAllApplicationsForUser returns applications for user hankkeet`() {
```
corvidian, line 276: I think this tests everything `getAllApplicationsForUser returns applications for the correct user` tests, so that one could be removed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -232,24 +272,42 @@ class ApplicationServiceITest : DatabaseTest() {
...
+        permissionService.setPermission(hanke2.id!!, username, Role.KATSELUOIKEUS)
+
+        val application1 = alluDataFactory.saveApplicationEntity(username = username, hanke = hanke)
+        val application2 =
+            alluDataFactory.saveApplicationEntity(username = "secondUser", hanke = hanke2)
+        alluDataFactory.saveApplicationEntity(username = "thirdUser", hanke = hanke3)
+
+        assertThat(applicationRepository.findAll()).hasSize(3)
+        val response = applicationService.getAllApplicationsForUser(username).map { it.id }
+        assertThat(response).containsExactlyInAnyOrder(application1.id, application2.id)
```
corvidian, line 291: While I'm not a puritan about AAA-pattern (Arrange, Act, Assert), I find it very helpful to have the Act part as a separate code block, i.e. surrounded by empty lines. Makes it easier to spot what's actually being tested.
```suggestion
        assertThat(applicationRepository.findAll()).hasSize(3)
        
        val response = applicationService.getAllApplicationsForUser(username)
                
        assertThat(response.map { it.id }).containsExactlyInAnyOrder(application1.id, application2.id)
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
corvidian: I'm not sure I understand your response correctly, but I think there should be explicit tests for what happens when the PermissionService returns false.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
corvidian: I think I understand this test now after the changes. It checks that the entity relations are set up in a way that the hanke can't be changed for an application even if we try.

I think this test is in the wrong place, since it doesn't test any ApplicationService functionality.

I'm not sure how I feel about the change in hanke being just silently ignored. Usually, when we do something stupid, we want to know about it. If someone writes code that tries to change the hanke of an existing application, it should fail loudly so the author knows they made an oopsie.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -1007,7 +1111,47 @@ class ApplicationServiceITest : DatabaseTest() {
...
+        every { cableReportServiceAllu.create(any()) }.returns(2)
+
+        val hanke = createHanke()
+        val application =
+            applicationService.create(
+                AlluDataFactory.createApplication(id = null, hankeTunnus = hanke.hankeTunnus!!),
+                username
+            )
+
+        assertThat(application.hankeTunnus).isEqualTo(application.hankeTunnus)
```
corvidian, line 1146: This is a tautology.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
nkala: I removed the test.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
nkala: Yes.

services/hanke-service/src/main/resources/db/changelog/changesets/024-connect-application-to-hanke.yml
nkala: Fixed file settings of the IDE.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
nkala: Done.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogServiceTest.kt
nkala: Done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
nkala: Done.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
nkala: It checked the cable report data so I didn't touch it.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
corvidian: It's still there?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
corvidian: Oh yes, that's right.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
nkala: This was fixed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
corvidian: I don't see the fix.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
nkala: There are more tests for that now.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
nkala: Should not be.

corvidian: Isn't this the same test slightly updated?
```
    @Test
    fun `Hanke of an application cannot be changed`() {
        every { cableReportServiceAllu.create(any()) }.returns(2)

        val hanke = createHanke()
        val application =
            applicationService.create(
                AlluDataFactory.createApplication(id = null, hankeTunnus = hanke.hankeTunnus!!),
                username
            )

        assertThat(application.hankeTunnus).isEqualTo(hanke.hankeTunnus)

        val applicationEntity = applicationRepository.getOne(application.id!!)
        applicationEntity.hanke = hankeRepository.save(HankeEntity(hankeTunnus = "HAI-1111"))
        applicationRepository.save(applicationEntity)

        assertThat(applicationRepository.getOne(application.id!!).hanke.hankeTunnus)
            .isEqualTo(hanke.hankeTunnus)
    }
```

PR #272, opened 2023-03-08
HAI-1406 Prevent deletion of hanke if it is being handled in allu

pitkni:
# Description

Hanke can only be deleted if it has not proceeded to being handled in Allu. This commit checks application statuses (and throws exception if they are) before proceeding to hanke deletion. 

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
@@ -1,8 +1,12 @@
 package fi.hel.haitaton.hanke
 
+import com.ninjasquad.springmockk.MockkBean
+import fi.hel.haitaton.hanke.allu.ApplicationStatus
+import fi.hel.haitaton.hanke.application.ApplicationService
 import fi.hel.haitaton.hanke.domain.Hanke
 import fi.hel.haitaton.hanke.domain.HankeYhteystieto
 import fi.hel.haitaton.hanke.domain.Hankealue
+import fi.hel.haitaton.hanke.factory.AlluDataFactory
 import fi.hel.haitaton.hanke.factory.DateFactory
 import fi.hel.haitaton.hanke.factory.HankeFactory
 import fi.hel.haitaton.hanke.factory.HankeFactory.withGeneratedArvioija
@@ -20,6 +24,11 @@ import fi.hel.haitaton.hanke.logging.Status
 import fi.hel.haitaton.hanke.logging.UserRole
 import fi.hel.haitaton.hanke.test.TestUtils
 import fi.hel.haitaton.hanke.test.TestUtils.nextYear
+import io.mockk.Called
+import io.mockk.clearAllMocks
+import io.mockk.confirmVerified
+import io.mockk.every
+import io.mockk.verify
 import java.time.format.DateTimeFormatter
 import java.time.temporal.ChronoUnit
 import javax.persistence.EntityManager
@@ -30,15 +39,20 @@ import org.assertj.core.api.Assertions.byLessThan
 import org.geojson.Feature
 import org.geojson.LngLatAlt
 import org.geojson.Polygon
+import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.EnumSource
 import org.skyscreamer.jsonassert.JSONAssert
 import org.skyscreamer.jsonassert.JSONCompareMode
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.data.repository.findByIdOrNull
 import org.springframework.jdbc.core.JdbcTemplate
 import org.springframework.security.test.context.support.WithMockUser
 import org.springframework.test.context.ActiveProfiles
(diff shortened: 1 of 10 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -178,11 +180,35 @@ open class HankeServiceImpl(
     }
 
     @Transactional
-    override fun deleteHanke(hanke: Hanke, userId: String) {
-        hankeRepository.deleteById(hanke.id!!)
-        hankeLoggingService.logDelete(hanke, userId)
+    override fun deleteHanke(hankeId: Int, userId: String): Boolean {
```
corvidian, line 183: Why change this to take hankeId instead of hanke, when the the hanke is already fetched from DB in the controller?

corvidian: Some extra handling is required to make sure that the hanke exists, but this could all be avoided by taking in the actual non-nullable Hanke to begin with.

corvidian: Maybe change the Hanke to a HankeEntity, if that makes checking the applications easier?

pitkni: In the end this handles HankeEntity and the controller had Hanke (which does not know anything about hakemukset).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -178,11 +180,35 @@ open class HankeServiceImpl(
...
+     */
+    private fun HankeEntity.anyHakemusProcessingInAllu(): Boolean =
+        hakemukset.any {
+            logger.info { "Hakemus ${it.id} has alluStatus ${it.alluStatus}" }
+            when (it.alluStatus) {
+                null,
+                ApplicationStatus.PENDING,
+                ApplicationStatus.PENDING_CLIENT -> !applicationService.isStillPending(it.alluid)
+                else -> true
+            }
```
corvidian, line 209: This could be moved to an overloaded isStillPending method that takes ApplicationEntity as a parameter. This way others can use the same logic.

The documentation could say that applications can't move back to a pending status after leaving it. To justify why the else-branch is valid.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: I feel a bit uncomfortable using the entity in the controller. A better way would be to include hakemukset in Hanke object. But that may produce a lot more changes in the code base.

pitkni: Ok, I decided to change my approach quite a bit. Logic is now as follows:

1. Controller fetches a pair of Hanke and List<Application> from the service.
2. HankeService delete operation has hanke and applications as parameters now.

Why use a pair? Because Hanke does not contain information of its applications and controller should not be able to handle database entities. This current implementation also minimises database operations so it is better in that sense too.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: Done.

pitkni: Done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -114,17 +114,17 @@ class HankeController(
...
         val userId = currentUserId()
         if (!permissionService.hasPermission(hankeId, userId, PermissionCode.DELETE)) {
             throw HankeNotFoundException(hankeTunnus)
         }
 
-        hankeService.deleteHanke(hanke, userId)
-
-        logger.info { "Deleted Hanke: ${hanke.toLogString()}" }
+        hankeService.deleteHanke(hankeId, userId).let { result ->
+            logger.info { "Deletion of Hanke: $hankeTunnus successful: $result" }
```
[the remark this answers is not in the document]
pitkni: I changed the whole approach based on review comments. This is now irrelevant.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -114,17 +114,17 @@ class HankeController(
...
 
-        val hanke = hankeService.loadHanke(hankeTunnus) ?: throw HankeNotFoundException(hankeTunnus)
-        val hankeId = hanke.id!!
-
-        val userId = currentUserId()
-        if (!permissionService.hasPermission(hankeId, userId, PermissionCode.DELETE)) {
-            throw HankeNotFoundException(hankeTunnus)
+        hankeService.getHankeHakemuksetPair(hankeTunnus).let {
+            val hanke = it.first
+            val hakemukset = it.second
```
corvidian, line 119: I find it easier to read if pairs are destructured on use:
```suggestion
        hankeService.getHankeHakemuksetPair(hankeTunnus).let { (hanke, hakemukset) =>
```

pitkni: Fair point, changed as suggested.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -226,6 +251,18 @@ class HankeServiceITests : DatabaseTest() {
         assertFeaturePropertiesIsReset(result, mutableMapOf("hankeTunnus" to hanke.hankeTunnus))
     }
 
+    @Test
+    fun `getHankeHakemuksetPair maps hanke and hakemukset to a pair correctly`() {
```
corvidian, line 255: There could be tests for the cases where the hanke has no applications and where the hanke doesn't exist.

pitkni: Added two tests to verify the two mentioned scenarios.

corvidian (approved):

PR #275, opened 2023-03-10
HAI-1403 Validation that application area are inside hankealue

corvidian:
# Description

Add validation to check that all application areas are inside some hankealue. The validation is run when creating or updating an application. It's also run when sending an application to check that the hankealue hasn't changed after creating the application.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Checklist:

 or other location: 

# Other relevant info
Old tests are failing, they need to be fixed. There needs to be a better way to create a hanke with an alue and an application that's inside that alue.


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -120,6 +128,15 @@ open class ApplicationService(
             "Invalid geometry received when updating application for user $userId, id=${application.id}, alluid=${application.alluid}, reason = ${validationError.reason}, location = ${validationError.location}"
         }
 
+        val hankeId = application.hanke.id!!
+        application.applicationData.areas?.let { areas ->
```
pitkni, line 132: This is checking for existing areas and NOT the new ones. This is an update function and it should verify newApplicationData.

markohaarni (approved):
Works as expected.

PR #281, opened 2023-03-17
HAI-1349: Add perustaja data to hanke

pitkni:
# Description

Add perustaja name and sahkoposti to hanke. For now, perustaja is optional data. It will be mandatory after task HAI-1350.

If perustaja is provided, sahkoposti is mandatory and name is optional (we do not receive it from Helsinki profile yet(?)).

Perustaja cannot be changed via endpoints after it has been provided.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing
Scenario 1: Create hanke without perustaja, should work.
Scenario 2: Create hanke with perustaja, should work, and it cannot be modified after it has been set.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Perustaja.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Perustaja.kt
@@ -2,6 +2,6 @@ package fi.hel.haitaton.hanke.domain
 
 import fi.hel.haitaton.hanke.PerustajaEntity
 
-data class Perustaja(val nimi: String?, val sahkoposti: String)
+data class Perustaja(val nimi: String?, val email: String)
 
-fun Perustaja.toEntity(): PerustajaEntity = PerustajaEntity(nimi, sahkoposti)
+fun Perustaja.toEntity(): PerustajaEntity = PerustajaEntity(nimi, email)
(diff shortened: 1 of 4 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Perustaja.kt
```diff
@@ -0,0 +1,7 @@
+package fi.hel.haitaton.hanke.domain
+
+import fi.hel.haitaton.hanke.PerustajaEntity
+
+data class Perustaja(val nimi: String?, val sahkoposti: String)
```
markohaarni, line 5: Should sahkoposti be named email instead, so it would be consistent, as email is used elsewhere, for example in HankeYhteystieto?

pitkni: This implementation is following the specification mentioned in: https://helsinkisolutionoffice.atlassian.net/browse/HAI-1349

But it is good to be consistent so I will change this as suggested.

pitkni: Done.

markohaarni (approved):

PR #283, opened 2023-03-28
HAI-1571 Add possibility to auto generate hanke on hakemus

pitkni:
# Description

User must be able to create an application (hakemus) without existing hanke. However, an application must always have a related hanke. This commit adds a feature to auto-generate a hanke for application if the endpoint POST /hakemukset is called with parameter luo-hakemus=true.

When the user decides to update any information of the autogenerated hanke, it must contain the required information / fulfil same requirements as with a normal hanke.  

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -907,6 +913,21 @@ class HankeServiceITests : DatabaseTest() {
...
+    fun `generateHankeWithApplication generates hanke based on application`() {
+        val inputApplication = AlluDataFactory.createApplication()
+        every { applicationService.create(any(), USER_NAME) } returnsArgument 0
+
+        val result = hankeService.generateHankeWithApplication(inputApplication, USER_NAME)
+
+        with(result) {
+            val application = applications.first()
+            assertThat(hanke.hankeTunnus).isEqualTo(application.hankeTunnus)
+            assertThat(hanke.nimi).isEqualTo(application.applicationData.name)
```
corvidian, line 926: So this checks that the names in the created application and the generated hanke are the same. Should this also check that it's the name from the input application?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -269,6 +276,15 @@ class ApplicationController(
         }
     }
 
+    fun checkApplicationPermission(hankeTunnus: String?, userId: String) {
+        hankeTunnus.let { tunnus ->
+            hankeService.getHankeId(tunnus).let { id ->
+                if (id == null || !permissionService.hasPermission(id, userId, EDIT_APPLICATIONS))
+                    throw HankeNotFoundException(hankeTunnus)
```
corvidian, line 283: Since this throws a HankeNotFoundException, this method is only applicable to creating an application. This should be reflected in the name, so it doesn't get re-used inappropriately.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -98,22 +108,25 @@ class ApplicationController(
...
         val userId = currentUserId()
-        val hankeTunnus = application.hankeTunnus
-        val hankeId = hankeService.getHankeId(hankeTunnus)
 
-        if (
-            hankeId == null ||
-                !permissionService.hasPermission(hankeId, userId, PermissionCode.EDIT_APPLICATIONS)
-        ) {
-            throw HankeNotFoundException(hankeTunnus)
+        if (generateHanke) {
```
corvidian, line 118: This function two completely different paths depending on the value of the `luo-hanke` parameter. I think it would make more sense to have these as different methods with different URLs. Using a query parameter to separate the two flows feels hacky.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -537,6 +555,7 @@ open class HankeServiceImpl(
         hanke.loppuPvm?.let { entity.loppuPvm = hanke.loppuPvm?.toLocalDate() }
 
         hanke.perustaja?.let { entity.perustaja = it.toEntity() }
+        hanke.generated?.let { entity.generated = it }
```
corvidian, line 558: Doesn't this allow the user to set the generated value when calling the update endpoint for a hanke?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -62,18 +63,18 @@ open class HankeServiceImpl(
     private val applicationService: ApplicationService,
 ) : HankeService {
 
-    override fun getHankeId(hankeTunnus: String): Int? =
-        hankeRepository.findByHankeTunnus(hankeTunnus)?.id
+    override fun getHankeId(hankeTunnus: String?): Int? =
```
corvidian, line 66: Calling this with a null parameter makes no sense. I think the caller should handle the situations where hankeTunnus might be null.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -148,6 +149,22 @@ open class HankeServiceImpl(
...
+     */
+    @Transactional
+    override fun generateHankeWithApplication(
+        application: Application,
+        userId: String
+    ): HankeWithApplications {
+        val hanke = generateHankeFrom(application)
+        val tunnus = hanke.hankeTunnus ?: throw HankeArgumentException("Hanke must have tunnus")
+        val createdApplication =
+            applicationService.create(application.withHanketunnus(tunnus), userId)
```
corvidian, line 163: Since Application is a data class, copy can be used instead of a custom method.
```suggestion
            applicationService.create(application.copy(hankeTunnus = tunnus), userId)
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -899,4 +918,39 @@ open class HankeServiceImpl(
...
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
+                HankeStatus.DRAFT,
+                null,
+                true,
```
corvidian, line 953: Adding the parameter names to this call makes it a lot easier to see e.g. which parameter is set to true. Idea can even do it for you, if you ask nicely.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -148,6 +149,22 @@ open class HankeServiceImpl(
...
 
+    /**
+     * Create application when no existing hanke. Autogenerates hanke and applies application to it.
+     */
+    @Transactional
+    override fun generateHankeWithApplication(
+        application: Application,
+        userId: String
+    ): HankeWithApplications {
+        val hanke = generateHankeFrom(application)
```
corvidian, line 160: Generating the hanke automatically from an application should only be allowed for johtoselvityshakemus, not any other kind of applications. This should be checked somewhere.

While we don't have other types of applications at the moment, it will be really easy to forget adding this check when we add other types of applications later, so the check should be added now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
corvidian: I think there should be a test that tries to set the generated value through the update API.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -98,22 +108,25 @@ class ApplicationController(
...
+                .first()
+                .also { disclosureLogService.saveDisclosureLogsForApplication(it, userId) }
         }
 
-        val createdApplication = service.create(application, userId)
-
-        disclosureLogService.saveDisclosureLogsForApplication(createdApplication, userId)
-        return createdApplication
+        checkApplicationPermission(application.hankeTunnus, userId)
+        return service.create(application, userId).also {
```
corvidian, line 127: How is using `also` here clearer than using a variable?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -269,6 +276,15 @@ class ApplicationController(
         }
     }
 
+    fun checkApplicationPermission(hankeTunnus: String?, userId: String) {
+        hankeTunnus.let { tunnus ->
```
corvidian, line 280: This `let` doesn't do anything here, just aliases `hankeTunnus` with `tunnus`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -269,6 +276,15 @@ class ApplicationController(
...
+            hankeService.getHankeId(tunnus).let { id ->
+                if (id == null || !permissionService.hasPermission(id, userId, EDIT_APPLICATIONS))
+                    throw HankeNotFoundException(hankeTunnus)
+            }
```
corvidian, line 284: Using control structures increases the cognitive load required to understand the code, so they shouldn't be used unnecessarily.

```suggestion
            val hankeId = hankeService.getHankeId(hankeTunnus)
            if (hankeId == null || !permissionService.hasPermission(hankeId, userId, EDIT_APPLICATIONS))
                throw HankeNotFoundException(hankeTunnus)
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -148,6 +149,22 @@ open class HankeServiceImpl(
...
     }
 
+    /**
+     * Create application when no existing hanke. Autogenerates hanke and applies application to it.
+     */
+    @Transactional
+    override fun generateHankeWithApplication(
+        application: Application,
+        userId: String
+    ): HankeWithApplications {
```
corvidian, line 159: The return type is HankeWithApplications, but only the applications are ever used, and only first one of those.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -58,18 +58,23 @@ open class ApplicationService(
...
-                ?: throw HankeNotFoundException(application.hankeTunnus)
-
-        application.applicationData.areas?.let { areas ->
-            checkApplicationAreasInsideHankealue(hanke.id!!, areas) { applicationArea ->
-                "Application geometry doesn't match any hankealue when creating a new application for user $userId, " +
-                    "hankeId = ${hanke.id}, application geometry = ${applicationArea.geometry.toJsonString()}"
+            hankeRepository.findByHankeTunnus(hankeTunnus)
+                ?: throw HankeNotFoundException(hankeTunnus)
+
+        if (!hanke.generated) {
```
corvidian, line 72: Is creating a new application allowed for an automatically generated hanke?

corvidian: `checkApplicationAreasInsideHankealue` is also called when updating an application and when sending the application to Allu. Both are relevant cases for a johtoselvityshakemus with a generated hanke, so those checks should be disabled for generated hanke as well.

There should be tests for all of these cases.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: Yes that is a problem. I will find some solution for this but in the long run we should have specific dtos at the controller layer to control what fields the endpoint consumer can affect / see. 

For example the id field used in request bodies to create stuff should probably not even be there.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
pitkni: Not much difference, I prefer returning the result (and _also_ do the disclosure stuff) straight away. I thought also would be a pretty good option for this.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: Yes in this case. But this is already used and can be used elsewhere also. The real issue here is that I think hanke should contain applications at the domain level. It does not and we end up with stuff like this.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
pitkni: Changed function name.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
pitkni: There are now to different endpoints to create and create with hanke generation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/Application.kt
```diff
@@ -14,5 +14,5 @@ data class Application(
     val applicationIdentifier: String?,
     val applicationType: ApplicationType,
     val applicationData: ApplicationData,
-    val hankeTunnus: String,
+    val hankeTunnus: String?,
```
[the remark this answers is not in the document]
pitkni: I added a new data class for this to handle cable reports without hanketunnus in this new feature. Application class is left as it was before this pr.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: This is now irrelevant as the implementation has changed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/Application.kt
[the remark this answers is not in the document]
corvidian: I like the new class, but seeing it now, it could be a highly specific DTO for creating the cable report application, like you said in the other thread. 

Specifically, these fields are set to null when creating an entity from the request, so they could be omitted.
```
    val alluid: Int?,
    val alluStatus: ApplicationStatus?,
    val applicationIdentifier: String?,
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: I added functionality to sanitize hanke create and updated requests. Generated is should always be false when creating or updating the hanke.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
pitkni: This is now irrelevant as we kept Application hanketunnus non nullable. Changed

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: This is now irrelevant as we kept Application hanketunnus non nullable. Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: This method only accepts CableReportWithoutHanke so any other type of application cannot be used. Will that do?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
pitkni: This implementation was reverted to (almost) what it was before this pr (separated endpoints). Is it ok now?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/Application.kt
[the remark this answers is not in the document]
pitkni: Modified as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: So, the service creates hanke and an application. I think it is good that it returns both, even though the controller does not need the hanke information in this case.

Not returning the created hanke creates functionality that is cumbersome to test / verify.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -92,13 +92,15 @@ class HankeController(
...
         if (hanke == null) {
             throw HankeArgumentException("No hanke given when creating hanke")
         }
-        val createdHanke = hankeService.createHanke(hanke)
+        val sanitizedHanke = hanke.copy(id = null, generated = false)
+
+        val userId = currentUserId()
+        logger.info { "Creating Hanke for user $userId: ${hanke.toLogString()} " }
+
+        val createdHanke = hankeService.createHanke(sanitizedHanke)
```
corvidian, line 103: I think deciding what to do with the generated value is business logic so it should be done in the service. There are other fields that are reset when creating a hanke, so this could be added there.
https://github.com/City-of-Helsinki/haitaton-backend/blob/59091c289426cdc6e135f240a131b02b998c3920/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt#L124-L132

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -170,12 +172,18 @@ class HankeController(
...
+    private fun Hanke.ensureUpdatable(updatedHanke: Hanke, hankeTunnusFromPath: String): Hanke {
+        if (hankeTunnusFromPath != updatedHanke.hankeTunnus) {
             throw HankeArgumentException("Hanketunnus not given or doesn't match the hanke data")
         }
-        if (perustaja != null && perustaja != newHanke.perustaja) {
+        if (perustaja != null && perustaja != updatedHanke.perustaja) {
             throw HankeArgumentException("Updating perustaja not allowed.")
         }
+
+        return updatedHanke.apply { generated = false }
```
corvidian, line 187: Same thing here, I think resetting the generated flag should be a part of the business logic. Maybe add it here?

https://github.com/City-of-Helsinki/haitaton-backend/blob/59091c289426cdc6e135f240a131b02b998c3920/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt#L172-L177

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
corvidian: Yes, maybe it's better to return both for testability.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
corvidian: Yes, there's no longer any way to end up here with something other than a cable report application.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -101,21 +105,42 @@ class ApplicationController(
...
-        }
+        checkPermissionToCreate(hankeTunnus, userId)
 
         val createdApplication = service.create(application, userId)
 
         disclosureLogService.saveDisclosureLogsForApplication(createdApplication, userId)
         return createdApplication
     }
 
+    @PostMapping("/luo-hanke")
```
corvidian, line 117: Would this be clearer as "/johtoselvityshakemukset" without "/hakemukset" in front?

Somehow "/hakemukset/luo-hanke" feels cumbersome, and it breaks common conventions about RESTful URLs.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
pitkni: createHanke method is used both for generated and non-generated hanke. How would we know what value to set for generated? In my opinion sanitising  user input is not necessarily business logic. I see that as a safety measure so that nothing funky ends up in the service. I guess there could be a specific mapper to handle transformation from user input -> actual data to be handled.

corvidian: Yes, I forgot the method is called from both flows. Once again the endpoint-specific DTOs would save us from this headache. I guess this is fine as-is.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: > `checkApplicationAreasInsideHankealue` is also called when updating an application and when sending the application to Allu. Both are relevant cases for a johtoselvityshakemus with a generated hanke, so those checks should be disabled for generated hanke as well.
> 
> There should be tests for all of these cases.

Added the same condition for updating and sending. Includes tests to verify the skip.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
corvidian: Yes, that's fine.

pitkni: Fixed.

corvidian (approved):

PR #292, opened 2023-04-13
HAI-1274 liitetiedostot

nkala:
# Description

Added the ability to add attachments to hankkeet.

### Jira Issue: 

HAI-1274

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

Currently you need a http-client that has the ability to send multipart files.

# Checklist:

 or other location: 

# Other relevant info
The state of the attachment is set to OK. This has to be PENDING when the virus scanner is attached.


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,113 @@
...
+
+        if (!permissionService.hasPermission(hankeId, currentUserId, PermissionCode.VIEW)) {
+            throw HankeNotFoundException(hankeTunnus)
+        }
+
+        val hankeAttachments = attachmentService.getHankeAttachments(hankeTunnus)
+        return hankeAttachments
+    }
+
+    @PostMapping("/hanke/{hankeTunnus}")
```
corvidian, line 34: I think these would be better as `/hanke/{hankeTunnus}/liitteet` since attachments are a part of hanke, not the other way around.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,113 @@
...
+        val info = attachmentService.get(uuid)
+
+        if (
+            info.hankeId == null ||
+                !permissionService.hasPermission(info.hankeId!!, currentUserId, PermissionCode.VIEW)
+        ) {
+            throw HankeNotFoundException("")
+        }
+
+        return attachmentService.getData(uuid)
```
corvidian, line 78: Should the `Content-Disposition` header when sending the data? To tell what the name of this file is? Or will the frontend handle all of that?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+) : AttachmentService {
+    @Transactional
+    override fun add(hankeTunnus: String, liite: MultipartFile): HankeAttachment {
+        if (!supportedFiletypes.contains(liite.contentType)) {
+            throw AttachmentUploadError("Filetype not supported: " + liite.contentType)
+        }
+
+        if (liite.size > 1024 * 1024 * FILESIZE_MAXIMUM_MEGABYTES) {
+            throw AttachmentUploadError(
+                "File size should not exceed ${FILESIZE_MAXIMUM_MEGABYTES}Mb"
```
corvidian, line 36: b is bit, B is byte.
```suggestion
                "File size should not exceed ${FILESIZE_MAXIMUM_MEGABYTES}MB"
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentControllerITests.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    @Test
+    fun `Uploading without a session should fail`() {
+        val file = MockMultipartFile("liite", "text.txt", "text/plain", "ABC".toByteArray())
+        send("HAI-123", file).andExpect(status().`is`(401))
+    }
+
+    @Test
+    @WithMockUser(username)
+    fun `Uploading without hankeTunnus should fail`() {
```
corvidian, line 65: Isn't this with an unknown hankeTunnus rather than without one?

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
```diff
@@ -0,0 +1,56 @@
...
+              - column:
+                  name: name
+                  type: varchar(256)
+                  remarks: "Name for the attachment file"
+                  constraints:
+                    nullable: false
+              - column:
+                  name: username
+                  type: varchar(256)
+                  remarks: "Username of the creator"
```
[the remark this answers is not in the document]
corvidian: Hanke uses
```
 createdbyuserid      | character varying(40)       |           |          |
 createdat            | timestamp without time zone |           |          |
```
for this and created, maybe use the same convention?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/HankeAttachmentEntity.kt
```diff
@@ -0,0 +1,79 @@
...
+        result = 31 * result + tila.hashCode()
+        return result
+    }
+}
+
+enum class AttachmentTila {
+    PENDING,
+    FAILED,
+    OK
+}
```
corvidian, line 79: This is the status of the virus check, correct? This should be reflected in the naming.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
```diff
@@ -0,0 +1,56 @@
...
+                    foreignKeyName: fk_attachment_hanke
+                    references: hanke(id)
+              - column:
+                  name: data
+                  type: BYTEA
+                  remarks: "Binary data of the attachment"
+                  constraints:
+                    nullable: false
+              - column:
+                  name: tila
```
corvidian, line 52: Why is this one name in Finnish? Status is used elsewhere in the database.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
```diff
@@ -0,0 +1,56 @@
...
+                    nullable: false
+              - column:
+                  name: created
+                  type: timestamp
+                  remarks: "Creation date of the liite"
+                  defaultValueComputed: CURRENT_TIMESTAMP
+                  constraints:
+                    nullable: false
+              - column:
+                  name: hanke_id
```
corvidian, line 38: Is the plan to use the same table for application attachments as well?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+
+private const val FILESIZE_MAXIMUM_MEGABYTES = 10
+
+class AttachmentServiceImpl(
+    val hankeRepository: HankeRepository,
+    val attachmentRepository: AttachmentRepository,
+) : AttachmentService {
+    @Transactional
+    override fun add(hankeTunnus: String, liite: MultipartFile): HankeAttachment {
+        if (!supportedFiletypes.contains(liite.contentType)) {
```
corvidian, line 30: Should we save the content type? For sending the data with the same type to the client later.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,113 @@
...
+                !permissionService.hasPermission(info.hankeId!!, currentUserId, PermissionCode.VIEW)
+        ) {
+            throw HankeNotFoundException("")
+        }
+
+        return info
+    }
+
+    @GetMapping("/{uuid}/data")
+    fun getAttachmentData(@PathVariable uuid: UUID): ByteArray {
```
corvidian, line 67: All of the new endpoints need OpenAPI documentation, i.e. the `@Operation` and `@ApiResponses` annotations.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/HankeAttachment.kt
```diff
@@ -0,0 +1,13 @@
...
+import java.time.LocalDateTime
+import java.util.*
+
+data class HankeAttachment(
+    val id: UUID?,
+    val name: String?,
+    val user: String?,
+    var created: LocalDateTime?,
+    var tila: AttachmentTila?,
+    var hankeId: Int?,
```
corvidian, line 12: Why are these three vars?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,113 @@
...
+    fun getAttachments(@PathVariable hankeTunnus: String): List<HankeAttachment> {
+        val currentUserId = currentUserId()
+        val hankeId =
+            hankeService.getHankeId(hankeTunnus) ?: throw HankeNotFoundException(hankeTunnus)
+
+        if (!permissionService.hasPermission(hankeId, currentUserId, PermissionCode.VIEW)) {
+            throw HankeNotFoundException(hankeTunnus)
+        }
+
+        val hankeAttachments = attachmentService.getHankeAttachments(hankeTunnus)
```
pitkni, line 30: This can be simplified to: return attachmentService.getHankeAttachments(hankeTunnus)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,113 @@
...
+    ): HankeAttachment {
+        val currentUserId = currentUserId()
+        val hankeId =
+            hankeService.getHankeId(hankeTunnus) ?: throw HankeNotFoundException(hankeTunnus)
+
+        if (!permissionService.hasPermission(hankeId, currentUserId, PermissionCode.EDIT)) {
+            throw HankeNotFoundException(hankeTunnus)
+        }
+
+        val newAttachment = attachmentService.add(hankeTunnus, attachment)
```
pitkni, line 47: This can be simplified to: return attachmentService.add(hankeTunnus, attachment)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentService.kt
```diff
@@ -0,0 +1,22 @@
+package fi.hel.haitaton.hanke.liitteet
+
+import java.util.*
+import org.springframework.stereotype.Service
+import org.springframework.web.multipart.MultipartFile
+
+@Service
+interface AttachmentService {
```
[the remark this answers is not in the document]
pitkni: My own preference is that if you have an interface SomeService with one implementation SomeServiceImpl, then you should not have an interface.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+                ?: throw AttachmentUploadError("Hanke not found")
+
+        val attachment =
+            HankeAttachmentEntity(
+                name = liite.name,
+                data = liite.bytes,
+                created = getCurrentTimeUTCAsLocalTime(),
+                username = currentUserId(),
+                hanke = hanke,
+                tila = AttachmentTila.OK // FIXME this should be set to PENDING before launch
```
pitkni, line 51: What does this comment mean? Why is it not pending now?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+import fi.hel.haitaton.hanke.getCurrentTimeUTCAsLocalTime
+import java.util.*
+import javax.persistence.EntityNotFoundException
+import org.springframework.orm.jpa.JpaObjectRetrievalFailureException
+import org.springframework.transaction.annotation.Transactional
+import org.springframework.web.multipart.MultipartFile
+
+private val supportedFiletypes =
+    setOf(
+        "application/pdf",
```
pitkni, line 14: Would it be better to use values from https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/MediaType.html

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,96 @@
...
+            )
+        }
+
+        val hanke =
+            hankeRepository.findByHankeTunnus(hankeTunnus)
+                ?: throw AttachmentUploadError("Hanke not found")
+
+        val attachment =
+            HankeAttachmentEntity(
+                name = liite.name,
```
pitkni, line 46: There should probably be some sort of validation / check for the file name as it is a user input which could be malicious.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
pitkni, line 30: _The Content-Type for uploaded files is provided by the user, and as such cannot be trusted, as it is trivial to spoof. Although it should not be relied upon for security, it provides a quick check to prevent users from unintentionally uploading files with the incorrect type._

https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html

pitkni: https://www.baeldung.com/java-file-mime-type

But then again, I don't know if this is over engineering things. If we can trust the incoming payload, I guess we can just use the contentType.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentRepository.kt
```diff
@@ -0,0 +1,12 @@
+package fi.hel.haitaton.hanke.liitteet
+
+import fi.hel.haitaton.hanke.HankeEntity
+import java.util.*
```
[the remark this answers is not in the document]
nkala: Was enabled but had an exception for java.util. Funny.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
nkala: The hanke related endpoints have been moved.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentControllerITests.kt
nkala: Changed the name to better reflect the test.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
```diff
@@ -0,0 +1,56 @@
...
+            columns:
+              - column:
+                  name: id
+                  type: UUID
+                  remarks: "Unique identifier for an attachment"
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: name
```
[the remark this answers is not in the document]
nkala: It is the only name the attachment has currently. Users can't give names to attachments while uploading. Also the original filename has been sanitized at this point. Can change it but feels like a 50/50.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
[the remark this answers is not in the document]
nkala: Done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/HankeAttachmentEntity.kt
nkala: Changed it.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
```diff
@@ -0,0 +1,56 @@
...
+              - column:
+                  name: hanke_id
+                  type: bigint
+                  remarks: "Hanke that owns the attachment"
+                  constraints:
+                    nullable: false
+                    foreignKeyName: fk_attachment_hanke
+                    references: hanke(id)
+              - column:
+                  name: data
```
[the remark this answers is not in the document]
nkala: Changed it to content.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
nkala: I got that part in the next PR.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
nkala: Naming as been updated.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
nkala: It can't be trusted.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
nkala: These have been added.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentService.kt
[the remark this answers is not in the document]
nkala: Decide and refactor later.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/HankeAttachment.kt
```diff
@@ -0,0 +1,13 @@
+package fi.hel.haitaton.hanke.liitteet
+
+import java.time.LocalDateTime
+import java.util.*
+
+data class HankeAttachment(
```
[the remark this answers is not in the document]
nkala: Renamed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/HankeAttachment.kt
nkala: Changed hankeId to hankeTunnus. Changed the names too.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
nkala: Done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
nkala: Done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
nkala: So that it is testable now before the scanning is attached.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
nkala: Done. Did this by changing the filename.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
nkala: The payload content checking should be done separately. I added issue HAI-1612 to handle it.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
corvidian: I don't think there was any need to move the controller methods, changing the mappings would've been enough.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,131 @@
...
+        ) {
+            throw HankeNotFoundException("")
+        }
+
+        val metadata = attachmentService.get(uuid)
+        val mimeType = URLConnection.guessContentTypeFromName(metadata.name)
+
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.parseMediaType(mimeType)
+        response.setHeader(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=" + metadata.name)
```
corvidian, line 85: Why use the HttpServletResponse, when you have the headers there? 
```suggestion
        headers.add(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=" + metadata.name)
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentControllerITests.kt
corvidian: Was this test removed? I can't find it anymore.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
corvidian: We'll revisit this in HAI-1612.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/HankeAttachment.kt
corvidian: Yes, but why aren't they vals like the first three?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,131 @@
...
+    @Autowired private val permissionService: PermissionService,
+) {
+
+    @GetMapping("/{uuid}")
+    @Operation(
+        summary = "Get attachment metadata by UUID",
+        description =
+            """
+                Return information about an attachment. Available even when scan is pending.
+            """
```
corvidian, line 40: Probably no reason to use a raw string here. It's useful if the string is several lines long, but here it isn't.
```suggestion
        description = "Return information about an attachment. Available even when scan is pending."
```
Same applies for all the new endpoints.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,131 @@
...
+    fun getAttachment(@PathVariable uuid: UUID): AttachmentMetadata {
+        val currentUserId = currentUserId()
+        val info = attachmentService.get(uuid)
+        val hankeId = hankeService.getHankeId(info.hankeTunnus)
+
+        if (
+            hankeId == null ||
+                !permissionService.hasPermission(hankeId, currentUserId, PermissionCode.VIEW)
+        ) {
+            throw HankeNotFoundException("")
```
corvidian, line 51: Shouldn't this throw `AttachmentNotFoundException` for the methods in this file, since there's no hanke in the URL?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        return attachmentService.removeAttachment(uuid)
+    }
+
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    @ExceptionHandler(HankeNotFoundException::class)
+    fun handleArgumentExceptions(ex: HankeNotFoundException): HankeError {
```
corvidian, line 116: `HankeNotFoundException` is already handled in `ControllerExceptionHandler`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
```diff
@@ -0,0 +1,131 @@
...
+    @ExceptionHandler(AttachmentNotFoundException::class)
+    fun handleArgumentExceptions(ex: AttachmentNotFoundException): HankeError {
+        return HankeError.HAI3002
+    }
+
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    @ExceptionHandler(AttachmentUploadError::class)
+    fun handleArgumentExceptions(ex: AttachmentUploadError): HankeError {
+        return HankeError.HAI3001
+    }
```
corvidian, line 130: Should we log the exceptions?

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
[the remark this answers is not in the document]
corvidian: I still think "filename" would be clearer. It tells everything at a glance. You don't need to know whether the user can add another name to the attachment or not.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
corvidian: What about the `@ApiResponses`?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
pitkni: I brought the endpoints back to attachmentcontroller. Fyi: I renamed the package from liitteet to attachment.

services/hanke-service/src/main/resources/db/changelog/changesets/030-create-liitetaulut.yml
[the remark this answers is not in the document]
pitkni: Fixed to file_name. Used underscores on other fields too to make it more readable.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
pitkni: These are now added.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
pitkni: Fixed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
pitkni: These should be fixed now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
pitkni: This is removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentsController.kt
pitkni: Logging added.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentController.kt
```diff
@@ -0,0 +1,234 @@
...
+
+@RestController
+@RequestMapping
+class AttachmentController(
+    @Autowired private val attachmentService: AttachmentService,
+    @Autowired private val hankeService: HankeService,
+    @Autowired private val permissionService: PermissionService,
+) {
+
+    @GetMapping("liite/{id}")
```
corvidian, line 43: Why change these to singluar? Isn't plural the usual convention?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentController.kt
```diff
@@ -0,0 +1,234 @@
...
+    fun removeAttachment(@PathVariable id: UUID) {
+        val info = attachmentService.getMetadata(id)
+        val hankeId = hankeService.getHankeId(info.hankeTunnus)
+
+        checkAttachmentPermission(hankeId, currentUserId(), EDIT)
+
+        return attachmentService.removeAttachment(id)
+    }
+
+    @GetMapping("/hanke/{hankeTunnus}/liite")
```
corvidian, line 146: The path is `/hankkeet/` in HankeController. Paths here should follow the same convention.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,110 @@
...
+    val hankeRepository: HankeRepository,
+    val hankeAttachmentRepository: HankeAttachmentRepository,
+) : AttachmentService {
+
+    fun getExtension(filename: String): String {
+        return filename.split(".").last().trim()
+    }
+
+    fun sanitize(filename: String): String {
+        return Regex("[^0-9a-zA-ZäåöÄÖÅ.\\s]").replace(filename.trim(), "")
```
corvidian, line 32: This will mangle many completely valid filenames.
- Castréninkatu will become Castrninkatu
- mannerheimintie-8-liikennesuunnitelma.pdf will be squashed to mannerheimintie8liikennesuunnitelma.pdf
- valtakirja_Patrik_Floréen.pdf will be valtakirjapatrikfloren.pdf

This will, however, keep all line breaks and tab symbols.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/liitteet/AttachmentServiceImpl.kt
```diff
@@ -25,35 +24,57 @@ class AttachmentServiceImpl(
     val hankeRepository: HankeRepository,
     val attachmentRepository: AttachmentRepository,
 ) : AttachmentService {
+
+    fun getExtension(filename: String): String {
```
corvidian: This can be replaced at the call site with `FilenameUtils.getExtension(filename)` from Apache Commons.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentServiceImpl.kt
```diff
@@ -0,0 +1,110 @@
...
+    )
+
+private const val FILESIZE_MAXIMUM_MEGABYTES = 10
+
+class AttachmentServiceImpl(
+    val hankeRepository: HankeRepository,
+    val hankeAttachmentRepository: HankeAttachmentRepository,
+) : AttachmentService {
+
+    fun getExtension(filename: String): String {
```
corvidian, line 27: This can be replaced at the call site with FilenameUtils.getExtension(filename) from Apache Commons.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentController.kt
pitkni: I guess singular is just my personal preference. I don't think neither is wrong. But plural seems to be the convention here, so I will change this.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentServiceImpl.kt
pitkni: Basically what we want is this:

https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html

Filename Sanitization
Filenames can endanger the system in multiple ways, either by using non acceptable characters, or by using special and restricted filenames. For Windows, refer to the following [MSDN guide](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming-conventions). For a wider overview on different filesystems and how they treat files, refer to [Wikipedia's Filename page](https://en.wikipedia.org/wiki/Filename).

In order to avoid the above mentioned threat, creating a random string as a file-name, such as generating a UUID/GUID, is essential. If the filename is required by the business needs, proper input validation should be done for client-side (e.g. active content that results in XSS and CSRF attacks) and back-end side (e.g. special files overwrite or creation) attack vectors. Filename length limits should be taken into consideration based on the system storing the files, as each system has its own filename length limit. If user filenames are required, consider implementing the following:

Implement a maximum length
Restrict characters to an allowed subset specifically, such as alphanumeric characters, hyphen, spaces, and periods
If this is not possible, block-list dangerous characters that could endanger the framework and system that is storing and using the files.

pitkni: Using a regex is not a good idea and the only guaranteed secure solution I can think of is that we generate the file name.

pitkni: Any opinions on this?

pitkni: If we want to verify and keep the original file name then we can do something like this: https://www.baeldung.com/java-validate-filename

pitkni:
```
/**
 * Checks file name length and creates a path object from file name. Invalid name throws
 * AttachmentUploadException.
 */
fun validateFileName(fileName: String) {
    if (fileName.length > 128) {
        throw AttachmentUploadException("File name too long. Maximum is 128.")
    }
    try {
        Paths.get(fileName)
    } catch (e: InvalidPathException) {
        throw AttachmentUploadException("File name is not valid.")
    }
}
```

How about something like this?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentController.kt
```diff
@@ -0,0 +1,234 @@
...
+        checkAttachmentPermission(hankeId, currentUserId(), VIEW)
+
+        val metadata = attachmentService.getMetadata(id)
+        val mimeType = URLConnection.guessContentTypeFromName(metadata.fileName)
+
+        val responseHeaders =
+            HttpHeaders().apply {
+                contentType = MediaType.parseMediaType(mimeType)
+                add(CONTENT_DISPOSITION, "attachment; filename=${metadata.fileName}")
+            }
```
corvidian, line 114: What's wrong with just plain old
```suggestion
        val headers = HttpHeaders()
        headers.contentType = MediaType.parseMediaType(mimeType)
        headers.add(CONTENT_DISPOSITION, "attachment; filename=${metadata.fileName}")
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentController.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentController.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentServiceImpl.kt
pitkni: I decided to change approach quite a bit. At least I was not able to get any errors in the approach I suggested.

There is now a separate attachmentvalidator which takes care of the initial checks before virus scan.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentServiceImpl.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentController.kt
pitkni: Fixed.

corvidian (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentServiceImpl.kt
corvidian: 👍 for separate validators.

With the current checks, filenames shouldn't cause problems for our users.

I'm not quite sure about throwing a general exception every time there's something slightly wrong with the filename, but I guess this is something we'll get feedback on if it's troublesome.

PR #294, opened 2023-04-14
HAI-1533 Add virus scanning client for received files

pitkni:
# Description

Add a webclient to make http calls for file virus scanning. 

Associated changes in azure devops:
- https://dev.azure.com/City-of-Helsinki/haitaton/_git/haitaton-hanke-service-pipelines/pullrequest/4475
- Added CLAMAV_BASE_URL = clamav-rest-api.clamav.svc.cluster.local:3000 for development, testing, staging and production in: https://dev.azure.com/City-of-Helsinki/haitaton/_library?itemType=VariableGroups

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
At this point, there is no direct way to test this (other than unit tests). 

Functionality can however be tested by:
1. Running the whole stack in docker compose
2. shut down the backend
3. run backend via ./gradlew bootRun with the following code inserted to FileScanService:
```
    @Scheduled(initialDelay = 1000, fixedDelay = 10000)
    fun foo() {
        scanFiles(setOf(Pair("testfile.pdf", ByteArray(1000))))
    }
```
Then you can check the logs that everything works. Example in logs:
```
{"@timestamp":"2023-04-14T13:54:32.927+03:00","@version":"1","message":"Scanning 1 files.","logger_name":"fi.hel.haitaton.hanke.security.FileScanService","thread_name":"scheduling-1","level":"INFO","level_value":20000}
{"@timestamp":"2023-04-14T13:54:32.965+03:00","@version":"1","message":"Files scanned successfully.","logger_name":"fi.hel.haitaton.hanke.security.FileScanService","thread_name":"scheduling-1","level":"INFO","level_value":20000}
```

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Configuration.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Configuration.kt
@@ -54,8 +54,6 @@ class Configuration {
     @Value("\${haitaton.allu.password}") lateinit var alluPassword: String
     @Value("\${haitaton.allu.insecure}") var alluTrustInsecure: Boolean = false
 
-    @Value("\${haitaton.clamav.baseUrl}") lateinit var clamAvUrl: String
-
     @Bean
     fun cableReportService(webClientBuilder: WebClient.Builder): CableReportService {
         val webClient =
@@ -67,12 +65,6 @@ class Configuration {
             AlluProperties(baseUrl = alluBaseUrl, username = alluUsername, password = alluPassword)
         return CableReportServiceAllu(webClient, alluProps)
     }
-
-    @Bean
-    fun fileScanClient(webClientBuilder: WebClient.Builder): WebClient =
-        webClientBuilder.baseUrl(clamAvUrl).build().also {
-            logger.info { "Initialized file scan client with base-url: $clamAvUrl" }
-        }
 
     private fun createInsecureTrustingWebClient(
         webClientBuilder: WebClient.Builder
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Configuration.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Configuration.kt
@@ -33,6 +33,7 @@ import fi.hel.haitaton.hanke.tormaystarkastelu.TormaystarkasteluTormaysService
 import fi.hel.haitaton.hanke.tormaystarkastelu.TormaystarkasteluTormaysServicePG
 import io.netty.handler.ssl.SslContextBuilder
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory
+import mu.KotlinLogging
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
@@ -42,6 +43,8 @@ import org.springframework.jdbc.core.JdbcOperations
 import org.springframework.web.reactive.function.client.WebClient
 import reactor.netty.http.client.HttpClient
 
+private val logger = KotlinLogging.logger {}
+
 @Configuration
 @Profile("default")
 class Configuration {
@@ -50,6 +53,8 @@ class Configuration {
     @Value("\${haitaton.allu.username}") lateinit var alluUsername: String
     @Value("\${haitaton.allu.password}") lateinit var alluPassword: String
     @Value("\${haitaton.allu.insecure}") var alluTrustInsecure: Boolean = false
+
+    @Value("\${haitaton.clamav.baseUrl}") lateinit var clamAvUrl: String
 
     @Bean
     fun cableReportService(webClientBuilder: WebClient.Builder): CableReportService {
@@ -62,6 +67,12 @@ class Configuration {
             AlluProperties(baseUrl = alluBaseUrl, username = alluUsername, password = alluPassword)
         return CableReportServiceAllu(webClient, alluProps)
     }
+
+    @Bean
+    fun fileScanClient(webClientBuilder: WebClient.Builder): WebClient =
+        webClientBuilder.baseUrl(clamAvUrl).build().also {
+            logger.info { "Initialized file scan client with base-url: $clamAvUrl" }
+        }
 
     private fun createInsecureTrustingWebClient(
         webClientBuilder: WebClient.Builder
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonProperty
 import fi.hel.haitaton.hanke.toJsonString
 import java.time.Duration.ofSeconds
 import mu.KotlinLogging
-import org.apache.commons.lang3.BooleanUtils.isNotFalse
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.core.io.ByteArrayResource
 import org.springframework.http.MediaType.APPLICATION_JSON
@@ -93,6 +92,6 @@ data class FileResult(
 
 class FileScanException(message: String) : RuntimeException(message)
 
-fun List<FileResult>.hasInfected(): Boolean = any { isNotFalse(it.isInfected) }
+fun List<FileResult>.hasInfected(): Boolean = any { it.isInfected != false }
 
-fun List<FileResult>.filterInfected(): List<FileResult> = filter { isNotFalse(it.isInfected) }
+fun List<FileResult>.filterInfected(): List<FileResult> = filter { it.isInfected != false }
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
@@ -24,13 +24,10 @@ class FileScanService(
     private val fileScanClient: WebClient,
 ) {
 
-    fun scanFiles(files: Set<Pair<String, ByteArray>>): FileScanResult {
+    fun scanFiles(files: Set<Pair<String, ByteArray>>): FileScanResponse {
         logger.info { "Scanning ${files.size} files." }
-        val response = getResults(files).also { it.validateStatus() }
-
-        val virusDetected = response.data.result.any { isNotFalse(it.isInfected) }
-
-        return FileScanResult(virusDetected, response).also { it.logStatus() }
+        val result = getResults(files).also { response -> response.validateStatus() }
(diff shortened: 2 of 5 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
```diff
@@ -0,0 +1,89 @@
...
+            .orElseThrow()
+    }
+
+    private fun FileScanResponse.validateStatus() {
+        if (!success) {
+            throw FileScanException("Scan failed, result: ${this.toJsonString()}")
+        }
+    }
+
+    private fun FileScanResult.logStatus() {
```
corvidian, line 68: Since embedding side-effecting code in a data-container class is generally considered a bad practice, I think emulating that practice with extension methods is bad practice as well. This might just as well be
```suggestion
    private fun logStatus(FileScanResult result) {
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
```diff
@@ -0,0 +1,89 @@
...
+        if (virusDetected) {
+            logger.error { "Infected file detected, scan result: ${this.toJsonString()}" }
+        } else {
+            logger.info { "Files scanned successfully." }
+        }
+    }
+}
+
+data class FileScanResult(val virusDetected: Boolean, val scanResponse: FileScanResponse)
```
corvidian, line 77: Would there be a need for this class if we had something like
```
fun FileScanResult.hasInfected() = data.result.any { isNotFalse(it.isInfected) }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        return FileScanResult(virusDetected, response).also { it.logStatus() }
+    }
+
+    private fun getResults(files: Set<Pair<String, ByteArray>>): FileScanResponse {
+        val data =
+            MultipartBodyBuilder()
+                .apply {
+                    files.forEach { (name, bytes) ->
+                        part(FORM_KEY, ByteArrayResource(bytes), APPLICATION_PDF).filename(name)
```
corvidian, line 41: The attachments are not all PDFs.

pitkni: Alright, would it be ok to define them as a parameter in scanFiles function?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
```diff
@@ -0,0 +1,89 @@
...
+private val logger = KotlinLogging.logger {}
+
+private const val FORM_KEY = "FILES"
+
+@Service
+class FileScanService(
+    private val fileScanClient: WebClient,
+) {
+
+    fun scanFiles(files: Set<Pair<String, ByteArray>>): FileScanResult {
```
corvidian, line 27: Forming a set with the binary data of the files might be unnecessarily costly. Using a list might be better here.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Configuration.kt
```diff
@@ -63,6 +68,12 @@ class Configuration {
         return CableReportServiceAllu(webClient, alluProps)
     }
 
+    @Bean
+    fun fileScanClient(webClientBuilder: WebClient.Builder): WebClient =
```
corvidian, line 72: Won't this interfere with other clients using WebClient?

pitkni: Unless you have a better approach in mind?

corvidian: >According to my understanding, this will be injected to fields named fileScanClient.

I'm under the impression that the type is the important part here. So the value would be injected into all autowired fields with the type WebClient. FileScanService is currently the only place where a WebClient is autowired since the Allu client builds its own instance.

Anyway, it might be better to construct the WebClient where it's needed.
```
class FileScanService(
    webClientBuilder: WebClient.Builder,
    @Value("\${haitaton.clamav.baseUrl}") clamAvUrl: String,
) {
    private val fileScanClient: WebClient =
        webClientBuilder.baseUrl(clamAvUrl).build().also {
            logger.info { "Initialized file scan client with base-url: $clamAvUrl" }
        }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
corvidian: Aren't we now at the point where we need to measure the scan speed?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni: Removed FileScanResult class and modified FileScanResponse to:

```
data class FileScanResponse(val success: Boolean, val data: FileScanData) {
    fun hasInfected(): Boolean = data.result.any { isNotFalse(it.isInfected) }
}
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
```diff
@@ -0,0 +1,86 @@
...
+private val logger = KotlinLogging.logger {}
+
+private const val FORM_KEY = "FILES"
+
+@Service
+class FileScanService(
+    private val fileScanClient: WebClient,
+) {
+
+    fun scanFiles(files: Set<Pair<String, ByteArray>>): FileScanResponse {
```
corvidian, line 27: FileScanResponse.success is always true when this function returns, right?  And if `hasInfected` is a function, it could be an extension function on FileScanData.

Then again FileScanData only has one field, result. So it can be unwrapped right away.

So if we have
```
fun List<FileResult>.hasInfected(): Boolean = any { isNotFalse(it.isInfected) }
```
defined somewhere, I think we can return just `List<FileResult>` from this class.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    private fun FileScanResponse.validateStatus() {
+        if (!success) {
+            throw FileScanException("Scan failed, result: ${this.toJsonString()}")
+        }
+    }
+
+    private fun logStatus(result: FileScanResponse) {
+        if (result.hasInfected()) {
+            logger.error { "Infected file detected, scan result: ${this.toJsonString()}" }
```
corvidian, line 67: Is this really an error? This could be a warning since we detect and handle the exceptional flow appropriately.

Dumping the whole result as JSON might not be the most useful information for someone reading the logs.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni: https://helsinkisolutionoffice.atlassian.net/browse/HAI-1535 it's here.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni: Fixed, list now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni: Approach changed here. Public function returns List<FileResult> and hasInfected is now an extension function (Theres also another extension function to filter infected files).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Configuration.kt
pitkni: Fixed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni: Fixed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni:
```
    private fun logStatus(results: List<FileResult>) {
        if (results.hasInfected()) {
            results.filterInfected().forEach {
                logger.warn { "Infected file detected, scan result: ${it.toJsonString()}" }
            }
        } else {
            logger.info { "Files scanned successfully." }
        }
    }
    
```
Changed approach to filter infected file results and log them separately. It is still in json format, but I think it is readable enough.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
```diff
@@ -0,0 +1,98 @@
...
+    val name: String,
+    @JsonProperty("is_infected") val isInfected: Boolean?,
+    val viruses: List<String>
+)
+
+class FileScanException(message: String) : RuntimeException(message)
+
+fun List<FileResult>.hasInfected(): Boolean = any { isNotFalse(it.isInfected) }
+
+fun List<FileResult>.filterInfected(): List<FileResult> = filter { isNotFalse(it.isInfected) }
```
corvidian, line 98: You can use simple equality here:
```suggestion
fun List<FileResult>.hasInfected(): Boolean = any { it.isInfected != false }

fun List<FileResult>.filterInfected(): List<FileResult> = filter { it.isInfected != false }
```

corvidian (approved):
Free to merge after this minor fix.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/FileScanService.kt
pitkni: Done.

PR #295, opened 2023-04-17
HAI-1525 Capability to send emails

corvidian:
# Description

Add the client and configuration for sending emails.

Sending email in the cloud environments is still open, so disable all email sending by default. Enable it for Docker Compose, bootRun and specific test environments.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Uncomment the temporary function in EmailSenderService.
2. Start the service with `./gradlew bootRun` or with docker compose.
3. Go to http://localhost:3003/ to see the sent email.

# Checklist:

 or other location: Readme


services/hanke-service/src/test/resources/application-emailtest.properties
```diff
@@ -0,0 +1,8 @@
+haitaton.email.enabled=true
```
pitkni, line 1: Would it be possible to rename this file to application-test.properties? That way it could be used as a properties file for spring boot tests in general.

docker-compose.yml
```diff
@@ -66,6 +80,10 @@ services:
       ALLU_USERNAME: ${ALLU_USERNAME}
       ALLU_PASSWORD: ${ALLU_PASSWORD}
       ALLU_INSECURE: ${ALLU_INSECURE}
+      ALLU_UPDATE_INITIAL_DELAY: ${ALLU_UPDATE_INITIAL_DELAY}
```
pitkni, line 83: WARN[0000] The "ALLU_UPDATE_INITIAL_DELAY" variable is not set. Defaulting to a blank string. 

I got that warning while running docker compose. Is that value supposed to be defined somewhere?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/email/EmailSenderService.kt
```diff
@@ -0,0 +1,69 @@
...
+        val helper = MimeMessageHelper(mimeMessage, true, "utf-8")
+        helper.setText(textBody, htmlBody)
+
+        helper.setTo(to)
+        helper.setSubject(subject)
+        helper.setFrom(from)
+        mailSender.send(mimeMessage)
+    }
+
+    fun sendJohtoselvitysCompleteEmail(
```
pitkni, line 55: Could be a matter of personal preference but I would declare public functions before private ones. This method calls the private method so it would be easier to follow what is going on if the public function was declared first.

services/hanke-service/README.md
```diff
@@ -167,6 +167,11 @@ cd haitaton-backend
 .\scripts\HEL-GIS-data-import.cmd
 ```
 
+# Emails
+When running locally, when sending emails from the system the emails are sent to smtp4dev, which is
```
pitkni, line 171: When running locally, when sending emails from the system the emails are sent to smtp4dev, -> When running locally, the system the emails are sent to smtp4dev,

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Extensions.kt
```diff
@@ -9,6 +10,13 @@ fun Any?.toJsonString(): String = OBJECT_MAPPER.writeValueAsString(this)
 fun Any?.toJsonPrettyString(): String =
     OBJECT_MAPPER.writerWithDefaultPrettyPrinter().writeValueAsString(this)
 
+fun String.getResource() =
+    // The class here is arbitrary, could be any class. Using ClassLoader might be cleaner, but it
```
pitkni, line 14: I would remove this comment, feel free to ignore.

corvidian: I think this might be useful so that no one wonders what's special about the JsonNode class.

services/hanke-service/src/test/resources/application-emailtest.properties
corvidian: There's nothing special about the name `application-test.properties`. It's read when the `test` profile is in use. The integration tests in Haitaton are using the `default` profile.

I could rename the `emailtest` profile as `test`, but this is not really suitable for general use, especially since it contains `spring.liquibase.enabled=false`.

docker-compose.yml
corvidian: Thank you for noticing. I had `ALLU_UPDATE_INITIAL_DELAY=60000000` in `.env`, so I didn't notice this.

I added an empty default value for this in docker-compose.yml to rid of the warning. I also added `ALLU_UPDATE_INITIAL_DELAY=60000` to .env.local as a hint to how to change the value.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/email/EmailSenderService.kt
corvidian: Yes, I agree. The public API of a class should come first so it's easier to see. I reordered the functions.

services/hanke-service/src/test/resources/application-emailtest.properties
pitkni: I tried to say that a more general purpose test profile properties file could be useful. But fair enough, It would not be useful in this case. This can be left as is.

pitkni (approved):

PR #302, opened 2023-04-20
HAI-1610 Change name to first and last name in Contact

pitkni:
# Description

Application Contact.name is divided into firstName and lastName.

Note. Allu still expects name information to be a single name attribute. Hence, firstName and lastName needs to be mapped into a full name. Full name will be null if first and last names are null.

Note. Previously saved application data jsonb fields will contain the name information as a full name. These cases will have their firstName and lastName null. It should not cause errors but it might be a good idea to delete these rows from the db.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing
This update requires changes in the front-end too. When both are implemented, just create a cable report application and verify that contact data is handled correctly.

# Checklist:

 or other location: 

# Other relevant info


markohaarni (approved):
Works as expected.

corvidian (approved):

PR #307, opened 2023-04-27
HAI-1558 Upgrade to Spring Boot 2.6

corvidian:
# Description

The upgrade brought with it a new version of AssertJ, which added an overloaded version of `allSatisfy`, which Kotlin can't separate from the previous one. After looking in to how to circumvent the issue, I found it simplest to replace those with asserts with assertk.

Also upgrade the two dependencies, which were giving security warnings in IDEA: PostgreSQL JDBC Driver and Testcontainers. Not that they would be the most critical, but you have to start somewhere.

Testcontainers had a change in 1.16.0 where they
> Only publish exposed ports (#4122)

This caused issues for us, since it seems like there has been a misundestanding on what exposed ports in Testcontainers mean. We were exposing a non-standard port in the Postgres container, probably to not collide with the database running on the host machine. But the exposed ports are from the perspective of the container. All ports are mapped to random ports on the host. So we were telling Testcontainers that the container has an open port 5433 we would like to share to the world, even though Postgres was running in port 5432. The reason it had worked previously, was because Testcontainers exposed all open ports on the container anyway.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/build.gradle.kts
+++ b/services/hanke-service/build.gradle.kts
@@ -5,7 +5,6 @@ import org.springframework.boot.gradle.tasks.run.BootRun
 group = "fi.hel.haitaton"
 version = "0.0.1-SNAPSHOT"
 java.sourceCompatibility = JavaVersion.VERSION_11
-val postgreSQLVersion = "42.2.18"
 val springDocVersion = "1.6.12"
 val geoJsonJacksonVersion = "1.14"
 val mockkVersion = "1.13.5"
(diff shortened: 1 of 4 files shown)


services/hanke-service/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ import org.springframework.boot.gradle.tasks.run.BootRun
 group = "fi.hel.haitaton"
 version = "0.0.1-SNAPSHOT"
 java.sourceCompatibility = JavaVersion.VERSION_11
-val postgreSQLVersion = "42.2.18"
+val postgreSQLVersion = "42.6.0"
```
pitkni, line 8: I think it is possible to let spring decide the compatible postgresql version. I.e. this can be removed, if there is no strict need for us to use a specific version.

corvidian: Good catch. Removed the version from the dependency.

I checked the other dependencies and noticed that there was an unnecessary declaration:
```
testImplementation("org.junit.jupiter:junit-jupiter:5.8.1")
```
`spring-boot-starter-test´ already imports that one.

Also `mockwebserver` has a managed version in Spring Boot, but in 2.6 it's a lot older than the one we use. 2.7 updates that dependency, so I'll remove the version in that PR.

pitkni (approved):

PR #310, opened 2023-05-02
HAI-1559 Miscellaneous dependency upgrades

corvidian:
# Description

I checked with a Gradle plugin, which dependencies can be upgraded. These are only ones that we can upgrade at this time. Some upgrades need a newer Spring, Kotlin, or Java. Some were still in beta, so I didn't use those yet.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


pitkni:
It could be stated in the pr what dependencies were impossible to update.

pitkni (approved):

PR #313, opened 2023-05-02
HAI-1562 Fix auth for users with ad groups

corvidian:
# Description

The response from the userinfo query was read as a String -> String map, which broke down when the response contains `ad_groups`, which is an array.

Fix this by only reading the `sub` field from the response, since it's the only one we care about.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Other relevant info
Alternative to #312


corvidian:
Closed in favor of #312

PR #315, opened 2023-05-05
HAI-1535 Connect file scanning to attachment uploads

pitkni:
# Description

If file does not pass validation / scan, an exception is thrown and attachment will not be saved.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

# Checklist:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/FileScanService.kt
```diff
@@ -30,14 +32,26 @@ class FileScanService(
             logger.info { "Initialized file scan client with base-url: $clamAvUrl" }
         }
 
-    fun scanFiles(files: List<Pair<String, ByteArray>>): List<FileResult> {
+    fun validate(file: MultipartFile): MultipartFile {
+        val scanInput = validScanInput(file)
```
corvidian, line 36: I don't think the business rules we have for what kind of attachments we allow should be a prerequisite for doing a virus scan. It feels like we're coupling two things that shouldn't have much to do with each other.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentValidator.kt
```diff
@@ -68,7 +73,7 @@ object FileNameValidator {
             "LPT9"
         )
 
-    fun validateFileName(input: String?): String {
+    fun validFileName(input: String?): String {
```
corvidian, line 76: This returns the same file name that was the input, it's in no way more valid than the input, except it can't be null. I prefer the original name for the function.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentValidator.kt
```diff
@@ -16,16 +16,21 @@ private val supportedFiletypes =
     )
 
 object AttachmentValidator {
-    fun validate(input: MultipartFile): MultipartFile {
-        val fileName = FileNameValidator.validateFileName(input.originalFilename)
+    fun validScanInput(attachment: MultipartFile): FileScanInput {
+        val fileName = FileNameValidator.validFileName(attachment.originalFilename)
+        val contentType = attachment.contentType.orEmpty()
```
corvidian, line 21: I'm having a hard time following what happens with null values in `attachment.contentType`. Here we change it to `""` if it's null. Then we give it to `contentTypeMatchesExtension` which can explicitly handle null values just fine. Then we log with the value just missing: `File liite.pdf extension does not match content type `.

For FileScanInput and further processing the value is taken from `attachment.contentType` again.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/FileScanService.kt
```diff
@@ -80,6 +94,26 @@ class FileScanService(
     }
 }
 
+data class FileScanInput(val name: String, val type: String, val bytes: ByteArray) {
+    override fun equals(other: Any?): Boolean {
```
corvidian, line 98: I think using the default instance identity for the byte array would be preferable to just ignoring it in equals and hashCode.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/FileScanService.kt
pitkni: My idea was to avoid duplication across application / hanke implementations. But yes, it might be better to keep these separate. I'll see what I can do.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentValidator.kt
pitkni: I somewhat feel that using !! here and there is a bit of an anti-pattern. But ok, multipartfile comes from Java and there's not much we can do about the nullability.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/FileScanService.kt
pitkni: Separated these two.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentValidator.kt
pitkni: Changed as suggested (almost).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentValidator.kt
pitkni: I cleaned the contentype handling. Currently we don't actually need it anywhere apart from the file name validation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/FileScanService.kt
pitkni: Fixed as suggested.

corvidian: What I was trying to say was the default data class implementation for equals and hashcode is fine here. IDEA warns about them here, because they use reference equals for arrays. So internally, they check something like:
```
        if (bytes === other.bytes) return false
```
I think that's ok here since we're not going to be comparing `FileScanInput`s against others that might have the exact same name and content unless they are the same instance. Currently neither `equals()` nor `hashCode()` is used anywhere.

The warning can be suppressed with:
```
data class FileScanInput(val name: String, @Suppress("ArrayInDataClass") val bytes: ByteArray)
```

This is not used anyways, so FFTI.

corvidian (approved):

PR #316, opened 2023-05-08
HAI-1545: Send application attachments to allu part 1/3

pitkni:
# Description

First part includes the scenario:
- When application is sent to Allu for the first time, all application attachments will be sent to Allu also.

Note: next parts will include:
- Sending of single attachments that have been added after initial sending to Allu.
- Restrictions on when attachments can be added.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. Add attachments to application.
2. Send application to Allu
3. Check logs and Allu that attachments have been sent

# Checklist:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
```diff
@@ -22,34 +23,38 @@ private val logger = KotlinLogging.logger {}
...
+                    }
+                    // Allu has gone back and forth on whether it returns the login token surrounded
+                    // with quotes or not. To be safe, remove them if they are found.
+                    .map { it.trim('"') }
+                    .block()
+
+            return token ?: throw AlluLoginException("Login token null")
         } catch (e: Throwable) {
-            throw AlluLoginException(e)
+            throw AlluLoginException(e.message)
```
corvidian, line 52: We want the complete exception here, not just the message. Otherwise, we'll lose valuable information on what caused the error.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
```diff
@@ -22,34 +23,38 @@ private val logger = KotlinLogging.logger {}
...
+                    .bodyToMono(String::class.java)
+                    .doOnError(WebClientResponseException::class.java) {
+                        logError("Error logging in to Allu", it)
+                    }
+                    // Allu has gone back and forth on whether it returns the login token surrounded
+                    // with quotes or not. To be safe, remove them if they are found.
+                    .map { it.trim('"') }
+                    .block()
+
+            return token ?: throw AlluLoginException("Login token null")
```
corvidian, line 50: I'm not sure if there's any way we can get a null in `token`. Using `!!` would be fine if there isn't.

The `token` val is not really needed here, `?:` or `!!` can be added directly after the `.block()`.

corvidian: I think AlluApiException would make more sense here than AlluLoginException.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
```diff
@@ -281,6 +273,33 @@ class CableReportServiceAllu(
...
 
+    private fun postAttachment(alluId: Int, token: String, attachment: Attachment) {
+        logger.info { "Sending attachment for application $alluId." }
+
+        val builder = MultipartBodyBuilder()
+        builder
+            .part("metadata", attachment.metadata, MediaType.APPLICATION_JSON)
+            .filename("metadata")
+        builder
+            .part("file", ByteArrayResource(attachment.file), MediaType.APPLICATION_PDF)
```
corvidian, line 284: We'll be sending other attachments besides PDFs now. The media type should be read from `attachment.metadata.mimeType`.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAlluITests.kt
```diff
@@ -68,6 +70,45 @@ class CableReportServiceAlluITests {
...
+        val file = "test file content".toByteArray()
+        val attachment = Attachment(metadata, file)
+        val mockResponse = MockResponse().setResponseCode(200)
+        (1..3).forEach { _ -> mockWebServer.enqueue(mockResponse) }
+        val attachments = listOf(attachment, attachment, attachment)
+
+        service.addAttachments(alluId, attachments)
+
+        assertThat(mockWebServer.takeRequest()).isValidLoginRequest()
+        assertThat(attachments).each {
```
corvidian, line 105: We're not asserting anything from attachments.
```suggestion
        attachments.foreach {
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
pitkni: Changed to AlluApiException.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
pitkni: Changed as suggested. Currently the junit test does not verify the content type. Actually there wasn't any junit tests for the whole method I think. This can be added but testing multipart content seems to be overly complicated.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAlluITests.kt
pitkni: Changed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
corvidian: Seems like it could be checked with:
```
        val boundary = request.headers["Content-Type"]?.split(";boundary=")!![1]
        val reader = MultipartReader(request.body, boundary)

        val types = buildList {
            while (true) {
                val part = reader.nextPart() ?: break
                add(part.headers["Content-Type"])
            }
        }
        
        assertThat(types).containsExactly("application/json", "image/png")
```
Where everything but the assert could be in a private function.

pitkni: Added content type check.

PR #320, opened 2023-05-11
HAI-1545 Send added attachment if application PENDING

pitkni:
# Description

Part 2/3 of sending attachments to Allu. After initial sending, if another attachment is added, it should be sent to Allu if application status is PENDING.

Note: the prevention of adding / deleting attachments after handling has started in Allu will be implemented in the next pr.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. Create application, send to allu. Add attachment. Should be sent.
2. Create application, add attachments, should not send.

# Checklist:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
@@ -15,7 +15,6 @@ import fi.hel.haitaton.hanke.allu.ApplicationStatus.PENDING
 import fi.hel.haitaton.hanke.allu.CableReportService
 import fi.hel.haitaton.hanke.application.ApplicationEntity
 import fi.hel.haitaton.hanke.application.ApplicationNotFoundException
-import fi.hel.haitaton.hanke.application.ApplicationService
 import fi.hel.haitaton.hanke.attachment.FILE_NAME_PDF
 import fi.hel.haitaton.hanke.attachment.HANKE_TUNNUS
 import fi.hel.haitaton.hanke.attachment.USERNAME
@@ -66,7 +65,6 @@ import org.testcontainers.junit.jupiter.Testcontainers
 @WithMockUser(USERNAME)
 @TestPropertySource(locations = ["classpath:application-test.properties"])
 class ApplicationAttachmentServiceITest : DatabaseTest() {
-    @MockkBean private lateinit var applicationService: ApplicationService
     @MockkBean private lateinit var cableReportService: CableReportService
     @Autowired private lateinit var applicationAttachmentService: ApplicationAttachmentService
     @Autowired private lateinit var applicationAttachmentRepository: ApplicationAttachmentRepository
@@ -85,7 +83,7 @@ class ApplicationAttachmentServiceITest : DatabaseTest() {
     @AfterEach
     fun tearDown() {
         checkUnnecessaryStub()
-        confirmVerified(applicationService, cableReportService)
+        confirmVerified(cableReportService)
         mockWebServer.shutdown()
     }
 
@@ -189,7 +187,7 @@ class ApplicationAttachmentServiceITest : DatabaseTest() {
         assertThat(result.applicationId).isEqualTo(application.id)
         assertThat(result.attachmentType).isEqualTo(typeInput)
         assertThat(result.scanStatus).isEqualTo(OK)
-        verify { applicationService wasNot Called }
+        verify { cableReportService wasNot Called }
     }
 
     @Test
@@ -204,7 +202,8 @@ class ApplicationAttachmentServiceITest : DatabaseTest() {
                     it.alluStatus = PENDING
                 }
                 .toApplication()
-        every { applicationService.isStillPending(application) } returns true
+        every { cableReportService.getApplicationInformation(alluId) } returns
+            AlluDataFactory.createAlluApplicationResponse(alluId, PENDING)
 
         applicationAttachmentService.addAttachment(
             applicationId = application.id!!,
@@ -213,7 +212,7 @@ class ApplicationAttachmentServiceITest : DatabaseTest() {
         )
 
         verifyOrder {
-            applicationService.isStillPending(application)
+            cableReportService.getApplicationInformation(alluId)
             cableReportService.addAttachment(alluId, any())
         }
     }
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
@@ -7,11 +7,14 @@ import assertk.assertions.hasSize
 import assertk.assertions.isEmpty
 import assertk.assertions.isEqualTo
 import assertk.assertions.isNotNull
+import com.ninjasquad.springmockk.MockkBean
 import fi.hel.haitaton.hanke.DatabaseTest
 import fi.hel.haitaton.hanke.HankeEntity
 import fi.hel.haitaton.hanke.HankeRepository
+import fi.hel.haitaton.hanke.allu.ApplicationStatus.PENDING
 import fi.hel.haitaton.hanke.application.ApplicationEntity
 import fi.hel.haitaton.hanke.application.ApplicationNotFoundException
+import fi.hel.haitaton.hanke.application.ApplicationService
 import fi.hel.haitaton.hanke.attachment.FILE_NAME_PDF
 import fi.hel.haitaton.hanke.attachment.HANKE_TUNNUS
 import fi.hel.haitaton.hanke.attachment.USERNAME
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -102,4 +111,19 @@ class ApplicationAttachmentService(
...
+    /**
+     * Attachment should be sent if application is in Allu but has not yet proceeded to handling
+     * (status PENDING). If PENDING, application is in Allu and alluId should exist.
+     */
+    private fun sendIfPending(application: Application, attachment: ApplicationAttachmentEntity) =
+        with(application) {
+            logger.info {
+                "Check application if should send attachment, alluId: '$alluid', alluStatus: '$alluStatus'"
+            }
+            if (alluStatus == PENDING) {
```
corvidian, line 124: `PENDING_CLIENT` applications should be updated as well.

Having separate logic for whether an application should be updated in Allu is not great. `ApplicationService.isStillPending` would be a good place, but it updates the status from Allu, which is probably not necessary here.

corvidian: Actually, checking the status from Allu wouldn't be terrible here. I was thinking of a different scenario, but this is a separate endpoint call. I think Allu accepts attachments also when the application is in HANDLING, so the upload will succeed whether the application is pending or not. I think we should check the current status from Allu here, not rely on what the status was during the last background update.

corvidian: Maybe:
```suggestion
            if (alluId != null && isStillPending(application)) {
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -72,7 +77,11 @@ class ApplicationAttachmentService(
                 application = application,
             )
 
-        return attachmentRepository.save(applicationAttachment).toMetadata().also {
+        val savedAttachment = attachmentRepository.save(applicationAttachment)
```
corvidian, line 80: Is refusing to save an attachment after the application is no longer pending coming in a later PR or should be added in this one?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -72,7 +77,11 @@ class ApplicationAttachmentService(
...
+
+        sendIfPending(application.toApplication(), savedAttachment)
```
corvidian, line 82: The API documentation should be updated to say we're uploading the attachment to Allu in some cases.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
pitkni: So if HANDLING is acceptable here, we can't use isStillPending call.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
```diff
@@ -174,6 +180,26 @@ class ApplicationAttachmentServiceITest : DatabaseTest() {
         assertThat(result.scanStatus).isEqualTo(OK)
     }
 
+    @Test
+    fun `addAttachment when application pending should send also`() {
```
corvidian, line 184: The negative case should be checked as well. I.e. add verify to the previous test.
```
verify { applicationService wasNot Called }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -216,6 +216,10 @@ open class ApplicationService(
         return applicationRepository.save(application).toApplication()
     }
 
+    open fun sendAttachment(alluId: Int, attachment: Attachment) {
```
corvidian, line 219: Couldn't ApplicationAttachmentService call the Allu client itself? Routing attachment updates through ApplicationService feels like mixing their responsibilities.

Maybe instead, sendAllAttachments could be moved to ApplicationAttachmentService, so that all Allu operations on attachments are done there.

pitkni: I think using applicationservice in the middle is ok because:
1. These are attachments of an application.
2. We already have applicationservice injected. Using cablereportservice (how about kaivuuilmoitus etc.) here will bring another dependency to attachments.

So I was thinking that talking to application service is the simplest solution.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
corvidian: HANDLING shouldn't be allowed here, I meant Allu accepts attachments regardless of the status.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Sorry, I was not clear enough previously. I meant that if we are going to use isStillPending from applicationservice (suggested in another comment), then we already have applicationservice at use. To me it is confusing to use cablereportservice(send) and applicationservice (isstillpending) to communicate with allu.

**But** if we are going to refactor cablereportservice, we can proceed as you suggested.

pitkni: I did some fixes. I chose to keep sendAllAttachments in applicationservice as a private functions because the use case there is to send attachments along with the application for the first time.

If this method was moved to applicationattachmentservice, then we would have a circular dependency of applicationservice calling attachmentservice and vice versa.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
pitkni: Added.

pitkni: Added.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni:
```
open class ApplicationService(
    private val applicationRepository: ApplicationRepository,
    private val alluStatusRepository: AlluStatusRepository,
    private val cableReportService: CableReportService,
    private val applicationLoggingService: ApplicationLoggingService,
    private val hankeKayttajaService: HankeKayttajaService,
    private val emailSenderService: EmailSenderService,
    private val geometriatDao: GeometriatDao,
    private val permissionService: PermissionService,
    private val hankeRepository: HankeRepository,
) {
```

This also is starting to look like we should be removing dependencies and not add them.

corvidian (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Attachment handling moved to attachment service. I believe this is more like what you suggested?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Configuration.kt
```diff
@@ -168,26 +166,6 @@ class Configuration {
             tormaystarkasteluDao
         )
 
-    @Bean
-    fun hankeAttachmentsService(
```
pitkni, line 172: These are removed because they have Service annotation and no need to create the beans like this.

corvidian: If the `@Bean`s in `IntegrationTestConfiguration` are used preferably over the `@Service` bean, we could drop most of the stuff in this file.

PR #321, opened 2023-05-12
HAI-1545 Add constraints to adding / deleting attachments

pitkni:
# Description

Attachment can be deleted if the application has not proceeded to handling in Allu.

Note: Attachments cannot be deleted from Allu, no such api endpoint exists there.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. You can add / delete attachments if application has not proceeded to handling in allu.
2. You cannot add / delete attachments if application is handling in allu.

# Checklist:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -122,27 +137,28 @@ class ApplicationAttachmentService(
...
-     * Attachment should be sent if application is in Allu (alluId present) but not yet in handling.
-     */
-    private fun sendIfPending(application: Application, attachment: ApplicationAttachmentEntity) =
-        with(application) {
-            logger.info { "Check application if should send attachment, alluId: '$alluid'" }
-            if (alluid != null && sendable(alluid, alluStatus)) {
-                cableReportService.addAttachment(alluid, attachment.toAlluAttachment())
-            }
-        }
+    private fun gonePastHandling(application: ApplicationEntity): Boolean = !isPending(application)
```
corvidian, line 140: I think `!isPending` is both more accurate and easier to understand than `gonePastHandling`.

Using different names for the same things between ApplicationService and ApplicationAttachmentService is unnecessarily confusing.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentController.kt
```diff
@@ -182,4 +211,10 @@ class ApplicationAttachmentController(
     @Hidden
     fun applicationNotFound(ex: ApplicationNotFoundException): HankeError =
         HankeError.HAI2001.also { logger.warn(ex) { ex.message } }
+
+    @ExceptionHandler(ApplicationAlreadyProcessingException::class)
+    @ResponseStatus(HttpStatus.CONFLICT)
+    @Hidden
+    fun applicationAlreadyProcessing(ex: ApplicationAlreadyProcessingException): HankeError =
```
corvidian, line 218: This and the not found handler could also be moved to ControllerExceptionHandler, so they don't have to be duplicated here and in ApplicationController.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
pitkni: Changed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentController.kt
pitkni: Changed as suggested.

corvidian (approved):

PR #322, opened 2023-05-15
HAI-1612 Add verification of uploaded attachment content type

pitkni:
# Description

Verify content type during upload using apache tika, valid value is saved to db and can be reused without re-checking.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing
1. Add attachment normally, should succeed.
2. Add attachment with header Content-Type that is not the actual content type of the file, should be detected and should fail.

# Checklist:

 or other location:


corvidian (approved):
👍

PR #330, opened 2023-05-22
HAI-1304 Add authentication to GDPR API

corvidian:
# Description

Authenticate the API as instructed in https://profile-api.dev.hel.ninja/docs/gdpr-api/

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
See the README changes for instructions on testing.

# Checklist:

 or other location:


.env.local
```diff
@@ -4,3 +4,4 @@ AUTH_DB_VOLUME=~/haitaton-auth-data
 NGINX_CONF=nginx.conf
 FRONT_DOCKERFILE=Dockerfile-local
 ALLU_UPDATE_INITIAL_DELAY=60000
```
pitkni, line 6: Would it be better to use HAITATON_GDPR_ENABLED = false? DISABLE = false is a double negative, and sometimes they might be confusing.

docker-compose.yml
```diff
@@ -77,6 +77,7 @@ services:
       HAITATON_OAUTH2_CLIENT_ID: https://api.hel.fi/auth/haitatonapidev
       HAITATON_OAUTH2_USER_INFO_URI: https://tunnistamo.test.hel.ninja/openid/userinfo
       #      HAITATON_CORS_ALLOWED_ORIGINS: http://localhost:8000
+      HAITATON_GDPR_DISABLED: ${HAITATON_GDPR_DISABLED:-true}
```
pitkni, line 80: Is this '-' on purpose?

corvidian: Yes. https://stackoverflow.com/a/70772707

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/gdpr/GdprController.kt
```diff
@@ -157,9 +165,33 @@ class GdprController(
...
+            throw AuthenticationException("JWT sub was $sub, but user id in URL was $userId")
+        }
+        if (scopes == null || !scopes.contains(requiredScope)) {
+            throw AuthenticationException(
+                "JWT scopes were $scopes, which didn't include $requiredScope"
+            )
+        }
+    }
+
     class NotImplementedError(endpointName: String) :
```
pitkni, line 182: Could this be moved to for example ControllerExceptionHandler? This seems like a good fit for more general usage also.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/OAuth2ResourceServerSecurityConfiguration.kt
```diff
@@ -24,11 +36,68 @@ class OAuth2ResourceServerSecurityConfiguration(
...
     }
+
+    /**
+     * Higher priority filter chain for authenticating GDPR API requests.
+     *
+     * When GDPR API is enabled, i.e. haitaton.gdpr.disabled is false, all requests to the GDPR API
+     * endpoints are authenticated by parsing and validating the JWT in the request.
+     */
+    @Bean
+    @Order(2)
```
pitkni, line 52: How does this work, should there be a bean with order 1 somewhere?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/gdpr/GdprController.kt
corvidian: For general use, we probably want to return a HankeError code. Here we don't, since we're implementing an externally described API.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/OAuth2ResourceServerSecurityConfiguration.kt
corvidian: The beans with higher priorities (lower number) are considered first. If there's no order, the bean is regarded as the lowest priority and gets an Order of `Integer.MAX_VALUE`.

While developing, I had a test bean with `@Order(1)`, but I didn't need it after all.

It doesn't matter what the value is, as long as it's smaller than `Integer.MAX_VALUE`, but it might be better to change this to `@Order(1)`, so a reader doesn't look for it elsewhere.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/gdpr/GdprController.kt
corvidian: Come to think of it, there's no real reason to check the disabled property in the controller anymore since SecurityFilter denies all access to the endpoints if disabled is true.

corvidian: Removed the disabled check and the NotImplemented stuff.

pitkni (approved):

PR #336, opened 2023-05-26
HAI-1688 Add real hankeTunnus to allu cable report identificationNumber

pitkni:
# Description

Remove identificationNumber from applicationdata used in application endpoints. Front-end does not need the value.

Value (hankeTunnus) is set when applicationdata is mapped to alludata.

Additional: similar thing is done with clientApplicationKind. The same value as workDescription field is used here as requested by Allu people (to sort incoming applications by description). 

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Send application to Allu and check hanketunnus.

Haitaton:
image

Allu:
image

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
@@ -561,7 +561,6 @@ class ApplicationServiceITest : DatabaseTest() {
         every { cableReportServiceAllu.getApplicationInformation(21) } returns
             AlluDataFactory.createAlluApplicationResponse(21)
         val expectedAlluRequest = newApplicationData.toAlluData(HANKE_TUNNUS)
-        assertEquals(application.hanke.hankeTunnus, expectedAlluRequest.identificationNumber)
         justRun { cableReportServiceAllu.update(21, expectedAlluRequest) }
         justRun { cableReportServiceAllu.addAttachment(21, any()) }
 
@@ -577,7 +576,7 @@ class ApplicationServiceITest : DatabaseTest() {
 
         verifyOrder {
             cableReportServiceAllu.getApplicationInformation(21)
-            cableReportServiceAllu.update(21, newApplicationData.toAlluData(HANKE_TUNNUS))
+            cableReportServiceAllu.update(21, expectedAlluRequest)
             cableReportServiceAllu.addAttachment(21, any())
         }
     }
@@ -925,7 +924,6 @@ class ApplicationServiceITest : DatabaseTest() {
         val applicationData = application.applicationData as CableReportApplicationData
         val pendingApplicationData = applicationData.copy(pendingOnClient = false)
         val expectedAlluRequest = pendingApplicationData.toAlluData(HANKE_TUNNUS)
-        assertEquals(application.hanke.hankeTunnus, expectedAlluRequest.identificationNumber)
         every { cableReportServiceAllu.create(expectedAlluRequest) } returns 26
         justRun { cableReportServiceAllu.addAttachment(26, any()) }
         justRun { cableReportServiceAllu.addAttachments(26, any()) }
@@ -947,7 +945,7 @@ class ApplicationServiceITest : DatabaseTest() {
         )
         assertEquals(ApplicationStatus.PENDING, savedApplication.alluStatus)
         verifyOrder {
-            cableReportServiceAllu.create(pendingApplicationData.toAlluData(HANKE_TUNNUS))
+            cableReportServiceAllu.create(expectedAlluRequest)
             cableReportServiceAllu.addAttachment(26, any())
             cableReportServiceAllu.addAttachments(26, any())
             cableReportServiceAllu.getApplicationInformation(26)
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
@@ -27,15 +27,11 @@ import fi.hel.haitaton.hanke.allu.AlluCableReportApplicationData
 import fi.hel.haitaton.hanke.allu.AlluException
 import fi.hel.haitaton.hanke.allu.AlluStatusRepository
 import fi.hel.haitaton.hanke.allu.ApplicationStatus
-import fi.hel.haitaton.hanke.allu.Attachment
 import fi.hel.haitaton.hanke.allu.CableReportService
 import fi.hel.haitaton.hanke.asJsonResource
 import fi.hel.haitaton.hanke.asUtc
-import fi.hel.haitaton.hanke.attachment.common.ApplicationAttachmentType.MUU
-import fi.hel.haitaton.hanke.attachment.testFile
 import fi.hel.haitaton.hanke.domain.Hanke
 import fi.hel.haitaton.hanke.factory.AlluDataFactory
-import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.createAttachmentMetadata
 import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.teppoEmail
 import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.withCustomer
 import fi.hel.haitaton.hanke.factory.ApplicationHistoryFactory
@@ -90,6 +86,7 @@ import org.springframework.test.context.jdbc.Sql
 import org.testcontainers.junit.jupiter.Testcontainers
 
 private const val USERNAME = "test7358"
+private const val HANKE_TUNNUS = "HAI23-5"
 
 private val dataWithoutAreas = AlluDataFactory.createCableReportApplicationData(areas = listOf())
 
@@ -563,7 +560,9 @@ class ApplicationServiceITest : DatabaseTest() {
             )
         every { cableReportServiceAllu.getApplicationInformation(21) } returns
             AlluDataFactory.createAlluApplicationResponse(21)
-        justRun { cableReportServiceAllu.update(21, newApplicationData.toAlluData()) }
+        val expectedAlluRequest = newApplicationData.toAlluData(HANKE_TUNNUS)
+        assertEquals(application.hanke.hankeTunnus, expectedAlluRequest.identificationNumber)
+        justRun { cableReportServiceAllu.update(21, expectedAlluRequest) }
         justRun { cableReportServiceAllu.addAttachment(21, any()) }
 
         val response =
@@ -578,7 +577,7 @@ class ApplicationServiceITest : DatabaseTest() {
 
         verifyOrder {
             cableReportServiceAllu.getApplicationInformation(21)
-            cableReportServiceAllu.update(21, newApplicationData.toAlluData())
+            cableReportServiceAllu.update(21, newApplicationData.toAlluData(HANKE_TUNNUS))
             cableReportServiceAllu.addAttachment(21, any())
         }
     }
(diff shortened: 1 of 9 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
```diff
@@ -2,57 +2,64 @@ package fi.hel.haitaton.hanke.application
...
-                ?: throw AlluDataException("applicationData.rockExcavation", AlluDataError.NULL)
-        return applicationData.workDescription +
-            if (rockExcavation) "\nLouhitaan" else "\nEi louhita"
-    }
+    private const val AREAS = "areas"
+    private const val CONTRACTOR = "contractorWithContacts"
+    private const val CUSTOMER = "customerWithContacts"
+    private const val DEVELOPER = "propertyDeveloperWithContacts"
+    private const val EXCAVATION = "rockExcavation"
+    private const val INVOICING = "invoicingCustomer"
```
corvidian, line 21: I don't see much point in defining a constant for a value that's used only once. Especially here, where the value should be identical to the field name. I think using a constant that has a different name from the field in question adds an unneeded layer of obfuscation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -427,51 +427,69 @@ open class ApplicationService(
...
             updateApplicationInAllu(application)
         }
     }
 
-    private fun updateApplicationInAllu(application: ApplicationEntity): Int {
-        val alluId =
-            application.alluid ?: throw ApplicationArgumentException("AlluId null in update.")
-        logger.info { "Uploading updated application with alluId $alluId" }
+    private fun updateApplicationInAllu(application: Application): Int =
+        with(application) {
```
corvidian, line 441: I don't think every function should start with `with() {`. Overall, I'm not the biggest fan of this pattern. I think it makes it less clear which symbols are local variables and which are fields of the object. All for saving a few characters when referencing the object?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -427,51 +427,69 @@ open class ApplicationService(
         }
     }
 
-    private fun sendApplicationToAllu(application: ApplicationEntity): Int {
+    private fun sendToAllu(entity: ApplicationEntity): Int {
```
corvidian, line 430: What was wrong with saying what was being sent in the method name?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -427,51 +427,69 @@ open class ApplicationService(
...
+        val application = entity.toApplication()
+        val attachments = entity.attachments.map { it.toAlluAttachment() }
```
corvidian, line 432: The mapped attachments are only needed in the create-branch, so they should be mapped in that branch.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -427,51 +427,69 @@ open class ApplicationService(
         }
     }
 
-    private fun sendApplicationToAllu(application: ApplicationEntity): Int {
+    private fun sendToAllu(entity: ApplicationEntity): Int {
+        val application = entity.toApplication()
```
corvidian, line 431: Why map the entity to the DTO/domain object? This seems to make everything downstream just a little bit more complicated.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: I think it is pretty clear from the function signature what will be sent. But I will change it back to what it was.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Not having to reference the object name on every access is exactly what 'with' is meant for. I'll go through this and see if this can / should be changed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
pitkni: I like using constants to lessen the likelihood of typos when repeating the same string values. Currently they are not repeating very much so I guess they can be removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
```diff
@@ -74,4 +81,16 @@ object ApplicationDataMapper {
             }
         }
     }
+
+    private fun path(vararg field: String) = field.joinToString(separator = ".", prefix = BASE)
+
+    private fun ZonedDateTime?.orThrow(path: String) = this ?: throw AlluDataException(path, NULL)
```
corvidian, line 87: 👍 Nice addition.

Though if this was generic, it could also be used for `rockExcavation` in `workDescription()`.
```suggestion
    private fun <T> T?.orThrow(path: String) = this ?: throw AlluDataException(path, NULL)
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Furthermore, shorter is usually better in Kotlin. But I agree that is up to a certain extent.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
```diff
@@ -74,4 +81,16 @@ object ApplicationDataMapper {
...
+
+    private fun workDescription(cableReport: CableReportApplicationData): String =
+        with(cableReport) {
+            val excavation = rockExcavation ?: throw AlluDataException(path(EXCAVATION), NULL)
+            return workDescription + excavation.description()
+        }
+
+    private fun Boolean.description(): String = if (this) "\nLouhitaan" else "\nEi louhita"
```
corvidian, line 95: I think this makes more sense as a regular function. Assigning a `description` to `Boolean` doesn't offer any hints on what it returns.

```suggestion
    private fun excavationText(excavation: Boolean) =
        if (excavation) "\nLouhitaan" else "\nEi louhita"
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -923,7 +924,9 @@ class ApplicationServiceITest : DatabaseTest() {
                 }
         val applicationData = application.applicationData as CableReportApplicationData
         val pendingApplicationData = applicationData.copy(pendingOnClient = false)
-        every { cableReportServiceAllu.create(pendingApplicationData.toAlluData()) } returns 26
+        val expectedAlluRequest = pendingApplicationData.toAlluData(HANKE_TUNNUS)
+        assertEquals(application.hanke.hankeTunnus, expectedAlluRequest.identificationNumber)
```
corvidian, line 928: Please test this separately, not inside the arrange phase of some other test.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
```diff
@@ -2,57 +2,64 @@ package fi.hel.haitaton.hanke.application
...
+            val description = workDescription(cableReport = this)
+            AlluCableReportApplicationData(
+                name = name,
+                customerWithContacts = customerWithContacts.toAlluData(path(CUSTOMER)),
+                geometry = getGeometry(applicationData = this),
+                startTime = startTime.orThrow(path(TIME_START)),
+                endTime = endTime.orThrow(path(TIME_END)),
+                pendingOnClient = pendingOnClient,
+                identificationNumber = hankeTunnus,
+                clientApplicationKind = description, // intentional
```
corvidian, line 40: This change should probably show up somewhere in tests as well.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
pitkni: Fixed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
pitkni: These are now removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
pitkni: Fixed by using generics.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
corvidian: I think this is a valid use for with():
```
    fun toAlluData(
        hankeTunnus: String,
        applicationData: CableReportApplicationData
    ): AlluCableReportApplicationData =
        with(applicationData) {
```
That method is clearly focused on the application data. It doesn't call any class dependencies or use any external values. And the references to it are always in the same position, so it's clear even without IDE colouring.

Shorter is only better if it's also clearer. Clarity, not terseness, is what defines good code.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Fixed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Moved mapping to a later stage.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: toAlluData is now tested in ApplicationDataMapperTest.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationDataMapper.kt
pitkni: toAlluData is now tested in ApplicationDataMapperTest.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
corvidian: Seems like both `updateApplicationInAllu` and `createApplicationInAllu` work just as well without mapping the entity to a domain object.

The only change is that getting hankeTunnus needs to be done through hanke: `application.hanke.hankeTunnus!!`. This could be simplified by adding a getter to ApplicationEntity for getting the hankeTunnus:
```kotlin
    fun hankeTunnus(): String = hanke.hankeTunnus!!
```
Sadly, Kotlin's lazy values don't seem to be working with JPA entities.

pitkni: Changed as suggested.

corvidian (approved):

PR #339, opened 2023-05-31
HAI-1679 Cancel application if initial attachment upload fails

corvidian:
# Description

When sending an application to Allu for the first time, we need to send all attachments at the same time. But if one of the uploads fails, the upload fails from Haitaton perspective, so the application is not marked as sent.

From Allu perspective, however, the application was uploaded successfully, and the officer looking at it doesn't know there are attachments missing.

When the user resends the application from Haitaton, a duplicate is created in Allu.

The proper way to fix this would be:
1. Mark the uploaded application as sent to Allu, even if the attachments uploads fail. Set the status to PENDING_CLIENT in both Allu and Haitaton.
2. Mark the attachments that were sent successfully as sent.
3. Tell the user that the sending failed and ask them to retry.
4. When the user tries to resend the application, notice that the application is already in Allu, and update it instead of creating a new one.
5. Send only the applications that haven't been sent before.

This would be a big change, however. Among other things, we don't have a way to mark applications as sent at the moment.

So instead, cancel the application in Allu if uploading the attachments fail. The application will remain in Allu in the cancelled state for all eternity. Failing the attachment uploads should be rare enough that this is not a big problem.

Also, add a comment to Allu on behalf of the Haitaton system to say why the application was cancelled. While adding comments, also add a comment when a user cancels the application in Haitaton while it's pending in Allu.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
1. Add an attachment of over 20MB to an application draft.
2. Send it to Allu. The attachment upload should fail.
3. Check in Allu that Haitaton has cancelled the application and added a comment about why it was cancelled.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportService.kt
@@ -35,4 +35,6 @@ interface CableReportService {
     fun getApplicationInformation(applicationId: Int): AlluApplicationResponse
 
     fun cancel(applicationId: Int)
+
+    fun sendSystemComment(applicationId: Int, msg: String): Int
 }
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportService.kt
@@ -5,36 +5,36 @@ import java.time.ZonedDateTime
 interface CableReportService {
 
     fun getApplicationStatusHistories(
-        applicationIds: List<Int>,
+        alluApplicationIds: List<Int>,
         eventsAfter: ZonedDateTime
     ): List<ApplicationHistory>
 
     fun create(cableReport: AlluCableReportApplicationData): Int
 
-    fun update(applicationId: Int, cableReport: AlluCableReportApplicationData)
+    fun update(alluApplicationId: Int, cableReport: AlluCableReportApplicationData)
 
-    fun addAttachment(applicationId: Int, attachment: Attachment)
+    fun addAttachment(alluApplicationId: Int, attachment: Attachment)
 
-    fun addAttachments(alluId: Int, attachments: List<Attachment>)
+    fun addAttachments(alluApplicationId: Int, attachments: List<Attachment>)
 
-    fun getInformationRequests(applicationId: Int): List<InformationRequest>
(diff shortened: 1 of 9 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
```diff
@@ -296,6 +298,30 @@ class CableReportServiceAllu(
             .orElseThrow()
     }
 
+    /** Send a comment to the application with Haitaton system as the sender. */
```
pitkni, line 301: The comment could explain the returned Int.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportService.kt
```diff
@@ -35,4 +35,6 @@ interface CableReportService {
     fun getApplicationInformation(applicationId: Int): AlluApplicationResponse
 
     fun cancel(applicationId: Int)
+
+    fun sendSystemComment(applicationId: Int, msg: String): Int
```
pitkni, line 39: Might not be in the scope of this pr but I'm often confused with the applicationId. Is it applicationId in haitaton or applicationId in Allu (alluId)? It seems that we are using applicationId to describe both. This should be addressed somehow, but maybe not now.

corvidian: The emergent convention seems to be that in the Allu client `applicationId` references the id in Allu and the Haitaton database id everywhere else. This kind of makes sense, since Allu API uses the term `applicationId` in some JSON objects to refer to their own id.

It would probably make more sense in Haitaton to rename it to `alluId` in the Allu client API.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportServiceAllu.kt
corvidian: Added an explanation of the return value.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/CableReportService.kt
corvidian: Allu has other IDs like IDs for comments and decision attachments, so `alluId` didn't seem like the correct name, at least for all occasions. I used `alluApplicationId`.

pitkni (approved):

PR #341, opened 2023-06-05
HAI-1704 Add validation to prevent start time be after end time.

pitkni:
# Description

Add validation to prevent start time be after end time. Application POST and PUT endpoints should not accept application startTime to be after endTime.

Additional: I refactored the whole application data validation. It is more than likely, that more validation will be implemented in the future, and I think this commit will make that work a bit easier.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Set start time to be after end time. Endpoint should return 400.

# Checklist:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
@@ -29,10 +29,10 @@ class ApplicationValidator : ConstraintValidator<ValidApplication, Application> 
 
     private fun isValidCableReport(cableReportData: CableReportApplicationData): Boolean =
         listOf(
-                validStartAndEnd(cableReportData.startTime, cableReportData.endTime),
-                validCustomersWithContacts(cableReportData.customersWithContacts())
+                { validStartAndEnd(cableReportData.startTime, cableReportData.endTime) },
+                { validCustomersWithContacts(cableReportData.customersWithContacts()) }
             )
-            .all { it }
+            .all { it() }
 
     /**
      * Checks if the start is before or equal to the end.
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
@@ -2,33 +2,87 @@ package fi.hel.haitaton.hanke.validation
 
 import fi.hel.haitaton.hanke.application.Application
 import fi.hel.haitaton.hanke.application.CableReportApplicationData
+import fi.hel.haitaton.hanke.application.Customer
+import fi.hel.haitaton.hanke.application.CustomerWithContacts
 import fi.hel.haitaton.hanke.isValidBusinessId
+import java.time.ZonedDateTime
 import javax.validation.ConstraintValidator
 import javax.validation.ConstraintValidatorContext
+import mu.KotlinLogging
+
+private val logger = KotlinLogging.logger {}
 
 class ApplicationValidator : ConstraintValidator<ValidApplication, Application> {
-    override fun isValid(application: Application?, context: ConstraintValidatorContext?): Boolean {
+    override fun isValid(application: Application, context: ConstraintValidatorContext?): Boolean {
 
-        val applicationData =
-            application?.applicationData
-                ?: throw InvalidApplicationDataException("Application data missing.")
-
-        return when (applicationData) {
-            is CableReportApplicationData ->
-                listOfNotNull(
-                        applicationData.customerWithContacts.customer,
-                        applicationData.contractorWithContacts.customer,
-                        applicationData.representativeWithContacts?.customer,
-                        applicationData.propertyDeveloperWithContacts?.customer,
-                    )
-                    .all {
-                        when {
-                            it.registryKey == null -> true
-                            it.registryKey.isValidBusinessId() -> true
-                            else -> throw InvalidApplicationDataException("Y-tunnus invalid.")
-                        }
-                    }
+        return when (application.applicationData) {
+            is CableReportApplicationData -> validOrThrow(application.applicationData)
         }
+    }
+
+    private fun validOrThrow(cableReportData: CableReportApplicationData): Boolean {
+        if (!isValidCableReport(cableReportData)) {
+            throw InvalidApplicationDataException("Received cable report contains invalid data.")
+        }
+        return true
+    }
+
+    private fun isValidCableReport(cableReportData: CableReportApplicationData): Boolean =
+        with(cableReportData) {
+            listOf(
+                    validStartAndEnd(startTime, endTime),
+                    validCustomerWithContactsNotNull(customerWithContacts),
+                    validCustomerWithContactsNullable(representativeWithContacts),
+                    validCustomerWithContactsNullable(contractorWithContacts),
+                    validCustomerWithContactsNullable(propertyDeveloperWithContacts)
+                )
+                .all { it }
+        }
+
+    /**
+     * Checks if the start is before or equal to the end.
+     *
+     * @param start the start time (nullable)
+     * @param end the end time (nullable)
+     * @return true if start is before or equal to end, false if either is null or if start is after
+     * end.
+     */
+    private fun validStartAndEnd(start: ZonedDateTime?, end: ZonedDateTime?): Boolean {
+        logger.info { "Application startDate: '$start', endDate: '$end'." }
+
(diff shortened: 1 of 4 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
```diff
@@ -2,33 +2,87 @@ package fi.hel.haitaton.hanke.validation
...
+    private fun validStartAndEnd(start: ZonedDateTime?, end: ZonedDateTime?): Boolean {
+        logger.info { "Application startDate: '$start', endDate: '$end'." }
+
+        if (start == null && end == null) {
+            return true
+        }
+
+        if (start == null || end == null) {
+            return false
+        }
```
corvidian, line 59: Why?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
```diff
@@ -2,33 +2,87 @@ package fi.hel.haitaton.hanke.validation
...
+            throw InvalidApplicationDataException("Received cable report contains invalid data.")
+        }
+        return true
+    }
+
+    private fun isValidCableReport(cableReportData: CableReportApplicationData): Boolean =
+        with(cableReportData) {
+            listOf(
+                    validStartAndEnd(startTime, endTime),
+                    validCustomerWithContactsNotNull(customerWithContacts),
```
corvidian, line 34: `contractorWithContacts` is also not nullable.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
```diff
@@ -2,33 +2,87 @@ package fi.hel.haitaton.hanke.validation
...
+            return true
+        }
+
+        return validRegistryKey(customer.registryKey)
+    }
+
+    private fun validRegistryKey(key: String?) = key?.isValidBusinessId() ?: true
+
+    /** Currently verifies customer only. */
+    private fun validCustomerWithContactsNotNull(contacts: CustomerWithContacts): Boolean =
```
corvidian, line 76: Is this version really necessary? The nullable version does the exact same checks when the value is not null.

corvidian, line 76: Is this version really necessary? The nullable version does the exact same validation when the value is not null.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
```diff
@@ -2,33 +2,87 @@ package fi.hel.haitaton.hanke.validation
...
+        validCustomer(contacts.customer)
+
+    /** Currently verifies customer only. */
+    private fun validCustomerWithContactsNullable(contacts: CustomerWithContacts?): Boolean {
+        if (contacts == null) {
+            return true
         }
+
+        return validCustomer(contacts.customer)
```
corvidian, line 85: I think using `?.let` and `?:` is a pretty idiomatic way to write a 'ternary' for null checks in Kotlin.
```suggestion
        contacts?.let { validCustomer(it.customer) } ?: true
```

Of course, if there were a lot of validation code inside the let, this would be more cumbersome than the if-then-return.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
corvidian: The problem with this is that the user can enter a date for one of the fields and then use the previous- or save and suspend buttons to save the incomplete page. Validating for the missing other value will cause the save to fail. And it's not obvious to the user why this happens.

For proper full validation, we should have one set of validation rules for saving drafts that allow for mandatory values to be missing and one set for sending the application to be processed, which would also check that all mandatory values are entered.

pitkni:
```
    private fun validStartAndEnd(start: ZonedDateTime?, end: ZonedDateTime?): Boolean {
        logger.info { "Application startDate: '$start', endDate: '$end'." }

        if (start == null && end == null) {
            return true
        }

        if (start == null || end == null) {
            return false
        }

        return !start.isAfter(end)
    }
```
Both must be null or both must be not null. Do we want to allow only one of them to contain a value?

pitkni: So, what are you suggesting here? Removal of 

```
        if (start == null || end == null) {
            return false
        }
```
or something else?

corvidian: One check with
```kotlin
        if (start == null || end == null) {
            return true
        }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
pitkni: Nice catch, thanks!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
```diff
@@ -2,33 +2,87 @@ package fi.hel.haitaton.hanke.validation
...
+
+        if (start == null || end == null) {
+            return false
+        }
+
+        return !start.isAfter(end)
+    }
+
+    /** Currently checks registryKey only. */
+    private fun validCustomer(customer: Customer?): Boolean {
```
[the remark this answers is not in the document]
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
pitkni: I think this can be ignored now as the implementation changed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
pitkni: Yes, it is redundant. I changed the handling of customerswithcontacts.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
pitkni: Done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
pitkni: I changed that to   

`validCustomersWithContacts(cableReportData.customersWithContacts())`

which is a list of not null.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
```diff
@@ -2,34 +2,67 @@ package fi.hel.haitaton.hanke.validation
...
         }
+        return true
     }
+
+    private fun isValidCableReport(cableReportData: CableReportApplicationData): Boolean =
+        listOf(
+                validStartAndEnd(cableReportData.startTime, cableReportData.endTime),
+                validCustomersWithContacts(cableReportData.customersWithContacts())
+            )
+            .all { it }
```
corvidian, line 35: It would make sense to me to have a list of lambdas, so we don't have to do every validation if the first one fails.
```suggestion
        listOf(
                { validStartAndEnd(cableReportData.startTime, cableReportData.endTime) },
                { validCustomersWithContacts(cableReportData.customersWithContacts()) },
            )
            .all { it() }
```

pitkni: Continuing the validation has some advantages too as it tells what issues (instead of an issue) are present (logs).

But it would not change the outcome, failed validation is a failed validation.

Fixed as suggested.

corvidian (approved):

PR #342, opened 2023-06-06
HAI-1723 Re-enable delete hanke endpoint

corvidian:
# Description

Delete hanke button is active in the UI in the first release, so the backend endpoint should be enabled as well.

I thought I needed to introduce a second feature flag, so I did some refactoring to the features. I introduced a centralized class for checking if an feature is enabled. In the end, it made a lot more sense to remove the feature-check from `DELETE /hanke/:id`, but I kept the refactoring since I think it will make handling feature flags easier in the future.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
The feature flag should still prevent creating and updating a hanke if you add
```
      HAITATON_FEATURE_HANKE_EDITING: false
```
to docker-compose.yml.

The delete endpoint should not be affected by this change.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Features.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Features.kt
@@ -1,0 +1,28 @@
+package fi.hel.haitaton.hanke.configuration
+
+import fi.hel.haitaton.hanke.EndpointDisabledException
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.ConstructorBinding
+
+@ConfigurationProperties(prefix = "haitaton")
+@ConstructorBinding
+data class Features(val features: Map<Feature, Boolean>) {
+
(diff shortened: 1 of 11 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/Features.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    /** Disabled by default, if not in application.properties. */
+    private fun isEnabled(feature: Feature): Boolean = features.getOrDefault(feature, false)
+
+    /**
+     * Throws an exception if the feature is not enabled.
+     *
+     * Disabled by default, if not in application.properties.
+     */
+    fun check(feature: Feature) {
```
pitkni, line 19: Check may be a bit vague. Could this be renamed for example checkIfEnabled etc? On the other hand javadoc seems to explain the purpose pretty well.

pitkni:
Good improvements.

pitkni (approved):

PR #344, opened 2023-06-07
HAI-1706 Change attachment deletion logic to check if application is sent

pitkni:
# Description

Previously application attachment deletion was prevented if application was not pending in Allu anymore.

However, as Allu does not allow deletion of attachments, we cannot allow deletion of attachments that are in Allu.

If application is in Allu (it has alluId), it means that the attachments are there also. Hence, deletion is not allowed.

Note: Front end is done on a separate ticket to not even try deleting attachments in these situations.

Additional: Did some modifications / refactoring to ControllerExceptionHandler to improve consistency.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Send application to Allu. Try deleting attachment, should not succeed (returns HTTP 409) and attachment is not deleted.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeError.kt
```diff
@@ -29,7 +29,7 @@ enum class HankeError(val errorMessage: String) {
     HAI1032("Invalid Hankealue data"),
     HAI2001("Application not found"),
     HAI2002("Incompatible application data type"),
-    HAI2003("Application is already processing in Allu and can no longer be updated."),
+    HAI2003("Application situation in Allu prevents operation."),
```
corvidian, line 32: I think this is too vague. It might be better to have a different exception and error code for the new situation.

pitkni: This is now separated to different exception and HAI2003 remains untouched.

PR #351, opened 2023-06-12
HAI-1748 Add more validation for application data

corvidian:
# Description

Validations added:
- There is at most one orderer
- Text fields don't have just blanks
- Country has a country code (UI uses constant FI here)

On top of the earlier:
- Registry key (business-id) is valid
- Start date is not after the end date

I ended up rewriting the validation using the same validators used in validating if a hanke is public or not. I refactored that quite a bit, though.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
This is easiest to verify with the UI before https://github.com/City-of-Helsinki/haitaton-ui/pull/323 is merged.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/Validators.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/Validators.kt
@@ -1,8 +1,6 @@
 package fi.hel.haitaton.hanke.validation
 
-private typealias ValidationError = String
-
-private val ValidationSuccess: ValidationError? = null
+import java.time.ZonedDateTime
 
 /**
  * Validator helper methods for on-the-fly validation (not at the REST API-boundary or the database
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/Validators.kt
```diff
@@ -11,67 +9,104 @@ private val ValidationSuccess: ValidationError? = null
...
+    /**
+     * Fails if the string is defined, non-empty and has nothing but whitespace characters.
+     * - null -> Ok
+     * - "" -> Ok
+     * - " " -> Fail
+     * - "*" -> Ok
+     * - "\t\n \t\n" -> Fail
+     * - " f " -> Ok
+     */
+    fun notJustWhitespace(value: String?, path: String): ValidationResult =
```
pitkni, line 46: How about notBlank?

pitkni (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/Validators.kt
corvidian: There's already a `notBlank()`, which calls `String.isBlank()`. That one considers both null and empty strings as blanks.

This was not easy to name, because it's different than the traditional definition of a blank string. I couldn't find a perfect name, so I went with one that's hopefully clear after reading the associated doc, but unclear enough to maybe get the reader to check the doc.

PR #352, opened 2023-06-12
HAI-1760 Add amount limit for uploaded attachments 

pitkni:
# Description

Add amount limit for uploaded attachments per application / hanke

Attachment amount limit is set to 20.

Additional: Fixed an issue with y-tunnus validation. Adding a non numeric value to y-tunnus would have crashed the validation. Tried to improve readability also.

## Type of change

- [x] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Adding more than 20 attachments should not be possible for hanke or application.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -20,6 +20,9 @@ fun getCurrentTimeUTCAsLocalTime(): LocalDateTime {
 
 fun currentUserId(): String = SecurityContextHolder.getContext().authentication.name
 
+/** Function to validate a given condition. Argument should throw if condition is not met. */
+fun checkCondition(condition: () -> Unit) = condition()
```
corvidian, line 24: How is `checkCondition { condition() }` different from `condition()`?

corvidian: The signature looks like it would do something like turn a boolean into an exception, so this might be distracting.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentEntity.kt
```diff
@@ -184,13 +185,21 @@ interface HankeAttachmentRepository : JpaRepository<HankeAttachmentEntity, UUID>
...
 @Repository
 interface ApplicationAttachmentRepository : JpaRepository<ApplicationAttachmentEntity, UUID> {
     @Modifying
     @Query("DELETE FROM ApplicationAttachmentEntity WHERE id = :id")
     fun deleteAttachment(id: UUID)
+
+    @Query(
+        "SELECT count(a.id) FROM ApplicationAttachmentEntity a WHERE a.application.id = :applicationId"
+    )
+    fun countByApplication(@Param("applicationId") applicationId: Long): Int
```
corvidian, line 202: Can't this be done with a derived query?
```suggestion
    fun countByApplicationId(applicationId: Long): Long
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -157,6 +173,14 @@ class ApplicationAttachmentService(
     private fun sendAttachment(alluId: Int, attachment: ApplicationAttachmentEntity) {
         cableReportService.addAttachment(alluId, attachment.toAlluAttachment())
     }
+
+    private fun attachmentAmountExceeded(applicationId: Long): Boolean {
```
corvidian, line 177: The attachment isn't exceeded until there are at least 21 attachments, so the name is slightly inaccurate.

```suggestion
    private fun attachmentAmountReached(applicationId: Long): Boolean {
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -58,14 +61,12 @@ class ApplicationAttachmentService(
...
-            throw ApplicationAlreadyProcessingException(application.id, application.alluid)
-        }
-
-        validateAttachment(attachment)
+        val application =
+            findApplication(applicationId).also { application ->
+                checkCondition { pending(application) }
+                checkCondition { roomForAttachment(applicationId) }
+                checkCondition { validFile(attachment) }
+            }
```
corvidian, line 69: I like the idea of collecting the different condition checks into a single block with simple method calls for the checks. Though I think `checkCondition` is superfluous since it does nothing. Also, the methods should be named so it's clear what they're used for.

```suggestion
            findApplication(applicationId).also { application ->
                ensurePending(application)
                ensureRoomForAttachment(applicationId)
                validateFile(attachment)
            }
```

Names like `pending()` are vague in what they're used for. Does it return a pending version of the application? Does it return a boolean on whether the application is considered pending? Or does it do something else, like here?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -33,8 +36,15 @@ fun currentUserId(): String = SecurityContextHolder.getContext().authentication.
 fun BusinessId.isValidBusinessId(): Boolean {
     logger.info { "Verifying businessId: $this" }
 
-    if (length != 9 || this[7] != '-') {
-        return false.also { logger.warn { "Invalid format." } }
+    if (length != 9 || substring(0, 7).any { !it.isDigit() } || this[7] != '-') {
```
corvidian, line 39: This check is getting complex enough that a regex would be a simplification. The same regex could be used to split the id to the sequence number and check digit parts.

```kotlin
    val businessIdRegex = "^(\\d{7})-(\\d)\$".toRegex() // Can be at class level

    val (sequenceNumber, checkDigit) =
        businessIdRegex.find(this)?.destructured
            ?: return false.also { logger.warn { "Invalid format." } }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
pitkni: I originally decided to keep it to improve readability of this code block. But as it is somewhat confusing and not bringing any added value, I'll remove it.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -43,13 +53,18 @@ fun BusinessId.isValidBusinessId(): Boolean {
...
-            val result = 11 - remainder
+            val calculatedCheck = if (remainder == 0) 0 else 11 - remainder
 
-            return when {
-                remainder == 1 -> false.also { logger.warn { "Remainder not valid." } }
-                (remainder == 0 && 0 == check) || result == check -> true
-                else -> false.also { logger.warn { "Invalid businessId." } }
+            if (remainder == 1) {
+                logger.warn { "Remainder not valid." }
+                return false
```
corvidian, line 60: I think it would be clearer if calculating the check digit would be clearly separated from checking the check digit.

Like,
```kotlin
    val calculatedCheck =
        sequenceNumber
            .map { it.digitToInt() }
            .zip(listOf(7, 9, 10, 5, 8, 4, 2))
            .sumOf { (num, multiplier) -> num * multiplier }
            .mod(11)
            .let { remainder ->
                when (remainder) {
                    1 -> {
                        logger.warn { "Remainder not valid." }
                        return false
                    }
                    0 -> 0
                    else -> 11 - remainder
                }
            }
```
After this, checking that the given and calculated check digits are the same can be done in separate, clear step:
```kotlin
    return if (calculatedCheck == checkDigit.toInt()) {
        true
    } else {
        logger.warn { "Check digit doesn't match." }
        false
    }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
pitkni: Removed checkCondition.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentEntity.kt
pitkni: countByApplicationId uses count(*) and using count(id) may have some performance benefits as it does not read the whole row. But this just may be a case of over engineering so I'll use countByApplicationId(applicationId: Long) for the sake of simplicity.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
pitkni: checkCondition is now removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentEntity.kt
pitkni: Changed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
pitkni: Added this with slight modifications.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentEntity.kt
corvidian: `count(*)` doesn't read the whole row, it only needs to find out how many there are. `count(id)` counts how many rows there are where id is not null, so it's generally slower.

Of course, with a limit of 20 attachments per application, this is all very theoretical. Unless there's a measurable benefit for doing otherwise, the choice should always be in favour of simplicity and clarity.

corvidian (approved):

PR #353, opened 2023-06-13
HAI-1758 Drop scan status from attachments

corvidian:
# Description

Drop the unused scan status column from both attachment tables. Remove it from the entities and other places that use it.

The frontend seems to except the backend to return the scan status, even though it doesn't use it for anything. Add a constant `"scanStatus": "OK"` to each attachment object returned from the backend.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentMetadata.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentMetadata.kt
@@ -1,5 +1,6 @@
 package fi.hel.haitaton.hanke.attachment.common
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import java.time.OffsetDateTime
 import java.util.UUID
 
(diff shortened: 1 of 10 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentMetadata.kt
```diff
@@ -1,23 +1,25 @@
 package fi.hel.haitaton.hanke.attachment.common
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import java.time.OffsetDateTime
 import java.util.UUID
 
+const val OK_STATUS = "OK"
```
pitkni, line 7: This can be removed.

corvidian: Removed

pitkni (approved):

PR #354, opened 2023-06-13
HAI-1758 Split attachment content to separate tables

corvidian:
# Description

Improve the memory use around attachments by separating the binary
content of the attachment files into separate tables and entities. Make
sure to only load the attachment content only when needed, and
separately for each attachment.

Also, remove oneToMany mapping of application and attachments. Add on
delete cascade to allow deletion of the owning entity of attachments.

Limit the concurrency on sending all attachments to Allu. Send at most 4
attachments cconcurrently, hopefully meaning only those 4 need their
content to be loaded into memory.

When testing locally, Allu didn't accept files fast enough when the
concurrency was set at four. Or maybe my connection couldn't transfer
the files fast enough.

There are possibly any number of environment-specific things that affect
how many files can be and should be uploaded concurrently. Make the
concurrency configurable by an environment variable, so we can tweak
this as necessary.

Set the default at three which seemed like a safe value when testing
locally.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Check that adding, sending and removing attachments works like before.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentContentEntity.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentContentEntity.kt
@@ -21,7 +21,6 @@ abstract class AttachmentContentEntity(
     /** Attachment data, i.e. the file itself. */
     @Lob
     @Type(type = "org.hibernate.type.BinaryType")
-    @Basic(fetch = FetchType.LAZY)
     @NotNull
     var content: ByteArray,
 )
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentContentEntity.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentContentEntity.kt
@@ -1,10 +1,8 @@
 package fi.hel.haitaton.hanke.attachment.common
 
 import java.util.UUID
-import javax.persistence.Basic
 import javax.persistence.Column
 import javax.persistence.Entity
-import javax.persistence.FetchType
 import javax.persistence.Id
 import javax.persistence.Lob
 import javax.persistence.MappedSuperclass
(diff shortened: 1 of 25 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentContentEntity.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.stereotype.Repository
+
+@MappedSuperclass
+abstract class AttachmentContentEntity(
+    @Id @Column(name = "attachment_id") var attachmentId: UUID,
+
+    /** Attachment data, i.e. the file itself. */
+    @Lob
+    @Type(type = "org.hibernate.type.BinaryType")
+    @Basic(fetch = FetchType.LAZY)
```
pitkni, line 24: I think Basic(fetch = FetchType.LAZY) was not working in the first place. It could possibly be removed.

See: https://vladmihalcea.com/the-best-way-to-lazy-load-entity-attributes-using-jpa-and-hibernate/

Basic(fetch = FetchType.LAZY)

This configuration alone is not sufficient because Hibernate requires bytecode instrumentation to intercept the attribute access request and issue the secondary select statement on demand.</p>

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentEntity.kt
```diff
@@ -165,34 +133,27 @@ class ApplicationAttachmentEntity(
 
         other as ApplicationAttachmentEntity
 
-        if (attachmentType != other.attachmentType) return false
-        if (application != other.application) return false
-
-        return true
+        return id == other.id
```
pitkni, line 136: Equals and hashcode functions in this class can possibly be improved. I'm not sure if using the superclass for comparing is a good idea. I know that this pr is not to blame. Maybe something like:

```
    override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (other !is ApplicationAttachmentEntity) return false

        if (id != other.id) return false
        if (attachmentType != other.attachmentType) return false
        if (applicationId != other.applicationId) return false

        return true
    }

    override fun hashCode(): Int {
        var result = id.hashCode()
        result = 31 * result + attachmentType.hashCode()
        result = 31 * result + applicationId.hashCode()
        return result
    }
```

Fields other than id can possible be removed, but I'm not sure.

corvidian: I'm pretty convinced we could get away with not defining these at all. We don't have any use for them ourselves, as long as we stick to using Lists instead of Sets. But I'm not sure how JPA would like that.

pitkni (approved):

PR #361, opened 2023-06-16
HAI-1792 Add validation to hakemus hanke generation endpoint

pitkni:
# Description

Add missing validation to hakemushankegeneration endpoint. We have a pretty decent and thorough validation in place for Application instances so I did not want to copypaste the same validation for this special case of hanke generation. Instead, I added an interface HaitatonApplication which is implemented by Application and CableReportWithoutHanke. This way we can use the validator to hanke HaitatonApplications and validate both types of applications.

Additional: replaced toNewApplication extension function as a normal class function. There was no need to use extension function.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
POST http://localhost:3001/api/hakemukset/johtoselvitys
```
{
    "id": 148,
    "alluid": null,
    "alluStatus": null,
    "applicationIdentifier": null,
    "applicationType": "CABLE_REPORT",
    "applicationData": {
        "applicationType": "CABLE_REPORT",
        "name": "Merikadun vj",
        "customerWithContacts": {
            "customer": {
                "type": null,
                "name": "",
                "country": "FI",
                "email": "",
                "phone": "",
                "registryKey": null,
                "ovt": null,
                "invoicingOperator": null,
                "sapCustomerNumber": null
            },
            "contacts": [
                {
                    "firstName": "Suvi",
                    "lastName": "Kupari",
                    "email": "foo@bar.com",
                    "phone": "05000000",
                    "orderer": true
                },
                {
                    "firstName": "Suvi",
                    "lastName": "Kupari",
                    "email": "foo@bar.com",
                    "phone": "",
                    "orderer": true
                }
            ]
        },
        "areas": [],
        "startTime": null,
        "endTime": null,
        "pendingOnClient": true,
        "workDescription": "Vesijohdon saneeraus",
        "contractorWithContacts": {
            "customer": {
                "type": null,
                "name": "",
                "country": "FI",
                "email": "",
                "phone": "",
                "registryKey": null,
                "ovt": null,
                "invoicingOperator": null,
                "sapCustomerNumber": null
            },
            "contacts": [
                {
                    "firstName": "",
                    "lastName": "",
                    "email": "",
                    "phone": "",
                    "orderer": false
                }
            ]
        },
        "rockExcavation": false,
        "postalAddress": {
            "streetAddress": {
                "streetName": "Merikatu 3"
            },
            "postalCode": "",
            "city": ""
        },
        "representativeWithContacts": null,
        "invoicingCustomer": null,
        "customerReference": null,
        "area": null,
        "propertyDeveloperWithContacts": null,
        "constructionWork": false,
        "maintenanceWork": true,
        "emergencyWork": false,
        "propertyConnectivity": true
    },
    "hankeTunnus": "HAI23-116"
}
```
Application has two orderers, and thus should fail validation.

# Checklist:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
@@ -10,7 +10,7 @@ import fi.hel.haitaton.hanke.andReturnBody
 import fi.hel.haitaton.hanke.domain.HankeWithApplications
 import fi.hel.haitaton.hanke.factory.AlluDataFactory
 import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.withContacts
-import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.withCustomer
+import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.withCustomerContacts
 import fi.hel.haitaton.hanke.factory.HankeFactory
 import fi.hel.haitaton.hanke.getResourceAsBytes
 import fi.hel.haitaton.hanke.permissions.PermissionCode.EDIT_APPLICATIONS
@@ -252,13 +252,9 @@ class ApplicationControllerITest(@Autowired override val mockMvc: MockMvc) : Con
     fun `create with hanke generation calls validation fails with invalid data and returns 400`() {
         val applicationInput =
             AlluDataFactory.createApplication()
-                .withCustomer(
-                    customer =
-                        AlluDataFactory.createCompanyCustomer()
-                            .withContacts(
-                                AlluDataFactory.createContact(orderer = true),
-                                AlluDataFactory.createContact(orderer = true)
-                            )
+                .withCustomerContacts(
+                    AlluDataFactory.createContact(orderer = true),
+                    AlluDataFactory.createContact(orderer = true)
                 )
                 .toCableReportWithoutHanke()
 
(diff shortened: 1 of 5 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
```diff
@@ -246,6 +247,26 @@ class ApplicationControllerITest(@Autowired override val mockMvc: MockMvc) : Con
...
+            AlluDataFactory.createApplication()
+                .withCustomer(
+                    customer =
+                        AlluDataFactory.createCompanyCustomer()
+                            .withContacts(
+                                AlluDataFactory.createContact(orderer = true),
+                                AlluDataFactory.createContact(orderer = true)
+                            )
+                )
+                .toCableReportWithoutHanke()
```
corvidian, line 263: There's an extension for changing the contacs of the customer.
```suggestion
            AlluDataFactory.createApplication()
                .withCustomerContacts(
                    AlluDataFactory.createContact(orderer = true),
                    AlluDataFactory.createContact(orderer = true)
                )
                .toCableReportWithoutHanke()
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/Application.kt
```diff
@@ -7,29 +7,35 @@ enum class ApplicationType {
     CABLE_REPORT,
 }
 
+/** Interface to enable Application and CableReportWithoutHanke handling equivalently. */
+interface HaitatonApplication {
```
corvidian, line 11: The alternative would be to have two separate validators that share the code validating the application data. That one would be conceptually simpler, but would of course involve more code.

I think the two approaches are six of one, half a dozen of the other. Except maybe for the naming. `HaitatonApplication` sounds like it's a special case of `Application` when it's the other way round. It's hard to find a good name for the interface, since it should be `Application`, but that one is already taken and for a good reason. The best name I can come up with is something like `BaseApplication`. The other approach would of course circumvent this naming problem.

pitkni: I guess baseapplication is ok. 

I ended up using an interface after initially having a separate validator with shared validation functions. Mainly due to tests, the implementation had something like 700 lines of code changes, most of it duplication. 

If there is not much of a difference which approach is better, I'd prefer this one.

pitkni: Renamed interface to BaseApplication.

corvidian (approved):

PR #371, opened 2023-06-26
HAI-1762 Add code coverage report

pitkni:
# Description

Add configuration for code coverage report using jacoco library. Coverage report results are supposed to be viewed in SonarCloud.

Note: this task only includes the creation of code coverage report. It is uploaded to Sonar Cloud via Azure DevOps pipeline.

Note: The currently used sonar cloud bot does not work with gradle builds as the scan is performed before the coverage report finishes. So that can only be used for static code analysis for code smells etc.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing

`./gradlew test integrationTest jacocoTestReport`

image

Check the created index.html for created coverage report combining test and integrationTest.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/build.gradle.kts
```diff
@@ -64,6 +64,8 @@ plugins {
 	kotlin("plugin.jpa") version "1.6.21"
 	idea
 	id("com.github.ben-manes.versions") version "0.42.0"
+	id("jacoco")
+	id("org.sonarqube") version "4.2.1.3168"
```
corvidian, line 68: Do we need the Sonarqube plugin, since we're not implementing the SonarCloud upload?

services/hanke-service/build.gradle.kts
```diff
@@ -120,6 +122,8 @@ dependencies {
 	annotationProcessor("org.springframework.boot:spring-boot-configuration-processor")
 }
 
+jacoco { toolVersion = "0.8.7" }
```
corvidian, line 125: The latest release is 0.8.10 (https://github.com/jacoco/jacoco/releases)

corvidian:
There should be something in the README about this.

services/hanke-service/build.gradle.kts
```diff
@@ -152,6 +156,13 @@ tasks {
 			exceptionFormat = TestExceptionFormat.FULL
 		}
 	}
+
+	jacocoTestReport {
+		executionData.setFrom(
+            fileTree(buildDir).include("/jacoco/test.exec", "/jacoco/integrationTest.exec")
+        )
+        dependsOn("integrationTest")
```
corvidian, line 164: I'm not sure if this task should depend on the integration test task. This will cause the integration tests to be run, even if you just want to create the report from the test run you just finished.

At least, if it depends on this task, it should also depend on the test task.

pitkni: It is possible to run them separately. The fact still is that there is a dependency from report to the tests. I.e. when you run jacocoTestReport, the tests must have been run before to have data for coverage.

How about this: 		`shouldRunAfter("test", "integrationTest")` ?

And then we should have some instructions in readme.

pitkni: shouldRunAfter
> For each supplied task, this action adds a task 'ordering', and does not specify a 'dependency' between the tasks. As such, it is still possible to execute 'taskY' without first executing the 'taskX' in the example.

This will not specify an actual dependency, but it will ensure at least correct ordering if multiple tasks are executed.

pitkni:
``` 
./gradlew test integrationTest jacocoTestReport                                                                                                                
> Task :services:hanke-service:integrationTest FAILED

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':services:hanke-service:integrationTest' (type 'Test').
  - Gradle detected a problem with the following location: '/haitaton-backend/services/hanke-service/build/jacoco/integrationTest.exec'.
    
    Reason: Task ':services:hanke-service:jacocoTestReport' uses this output of task ':services:hanke-service:integrationTest' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

I'm getting this error if I try to remove the declared dependency.

services/hanke-service/build.gradle.kts
pitkni: Based on the azure devops pipeline, it is not needed. Removed.

services/hanke-service/build.gradle.kts
pitkni: Updated.

services/hanke-service/build.gradle.kts
corvidian: Looks like I didn't try to run them at the same time, except with the check task I made. Seems like it's trying to run the report at the same time as it's running the integration tests. Adding the `shouldRunAfter` didn't help with this, since it allows the tasks to be run in parallel. The stronger `mustRunAfter` did.

pitkni: It is now using mustRunAfter

pitkni:
> There should be something in the README about this.

services/hanke-service/build.gradle.kts
```diff
@@ -156,6 +159,16 @@ tasks {
...
 	}
+
+	jacocoTestReport {
+		mustRunAfter("test", "integrationTest")
+		reports {
+			xml.required.set(true)
+		}
+		executionData.setFrom(
+            fileTree(buildDir).include("/jacoco/test.exec", "/jacoco/integrationTest.exec")
+        )
```
corvidian, line 170: For some reason, this file uses tabs for indentation. These lines use spaces instead.
```suggestion
			fileTree(buildDir).include("/jacoco/test.exec", "/jacoco/integrationTest.exec")
		)
```

Or we could re-indent the whole file with spaces. Or we could make Spotless and ktfmt do the formatting for us by adding this under spotless.kotlin: `target("**/*.kt", "**/*.kts")`.

services/hanke-service/README.md
```diff
@@ -28,11 +28,24 @@ Starting the application/services can be done afterwards with command line at ha
...
+Gradle Jacoco plugin is used to create a code coverage report.
+```
+$ ./gradlew :services:hanke-service:test :services:hanke-service:integrationTest :services:hanke-service:jacocoTestReport
+```
+Created report can be found at paths:
+
+| 	Format | Path	                                           |
+|---------|-------------------------------------------------|
+| 	 Html  | 	build/reports/jacoco/test/html/index.html      |
+| 	 Xml   | 	build/reports/jacoco/test/jacocoTestReport.xml |
```
corvidian, line 47: Why are there tabs here? Makes the content hard to read, depending on tab width settings.

pitkni: Tabs should be removed now.

services/hanke-service/build.gradle.kts
pitkni: Added target("**/*.kt", "**/*.kts")

Pretty much the whole file is now changed, but I guess it is a good thing.

corvidian: Yes, it changed the tabs to spaces, so it touched almost every line.

corvidian (approved):

PR #373, opened 2023-06-28
HAI Improve logging in adding attachments

corvidian:
# Description

Write to log at the beginning of when adding an attachment starts, and include the size of the attachment in that log. This will help us debug e.g. situations where the antivirus scan fails.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
@@ -63,6 +63,10 @@ class ApplicationAttachmentService(
         attachmentType: ApplicationAttachmentType,
         attachment: MultipartFile
     ): ApplicationAttachmentMetadata {
+        logger.info {
+            "Adding attachment to application, applicationId = $applicationId, " +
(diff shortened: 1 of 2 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -63,6 +63,10 @@ class ApplicationAttachmentService(
         attachmentType: ApplicationAttachmentType,
         attachment: MultipartFile
     ): ApplicationAttachmentMetadata {
+        logger.info {
+            "Adding attachment to application, applicationId = $applicationId, " +
+                "attachment name = ${attachment.originalFilename}, size = ${attachment.bytes.size}"
```
pitkni, line 68: Content type could be added also. Feel free to ignore.

pitkni (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
corvidian: That's a good idea, added it as well.

PR #378, opened 2023-07-05
HAI-1637 Remove organisaatiot

corvidian:
# Description

There used to be a manually maintained list of organisations that were selectable for hanke contacts. This idea has been dropped some time already, but some of the code has remained in the codebase. The database table was also never removed.

Remove the database table and all code related to the pre-stored organisaatiot.

This includes the organisaatioId field in contacts. This was used to link an yhteystieto to a specific organisaatio, even though it was never a foreign key. It's always null or zero, so retaining it serves no purpose.

The frontend has no references to the organisaatioId or the organisaatio API anymore, so this can be merged and deployed freely.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


services/hanke-service/src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -105,3 +105,5 @@ databaseChangeLog:
       file: db/changelog/changesets/037-add-delete-cascade-for-attachments.yml
   - include:
       file: db/changelog/changesets/038-add-attachment-content-tables.sql
+  - include:
```
pitkni, line 108: 040-drop-column-organisaatioid-from-yhteystiedot should be included here.

corvidian: Yes it should, thank you. I added it.

pitkni (approved):

PR #379, opened 2023-07-06
HAI-1638 Remove GeometriaController

corvidian:
# Description

The concept of hanke-geometria has been superceded by the concept of hankealue. One of the old geometria-endpoints was still up, but unused.

Remove the remaining geometria-endpoint, and the since it's the last method in the controller, the whole controller. Also remove the loadGeometriat method, which wasn't used anywhere else.

Refactor HankeControllerSecurityTests to test single calls separately to make it possible to see if any unnecessary stubbing of mocks is done.

Refactor HankeControllerITests
- Divide the tests for different endpoints to different inner classes.
- Move HankeControllerHankeEditingDisabledITests to a separate file.
- Move the unauthorized tests from HankeControllerSecurityTests to
  HankeControllerITests.
- Remove the positive tests from HankeControllerSecurityTests, since
  they were duplicates for tests in HankeControllerITests.
  - Delete HankeControllerSecurityTests, since it was now empty.
- Use verifySequence instead of many separate verify -blocks.
- Introduce a ResultMatcher for checking that a response is a specific
  HankeError, based on a private method in HankeControllerSecurityTests.
  - Use it in all tests in HankeControllerSecurityTests
  - This should be used more in tests. Just checking for a status might
    lead to false positives, since we don't know what caused the status.
- Disable a test for checking that a perustaja without email is not
  allowed. The tested functionality doesn't seem to be implemented. It
  passed earlier, because the test was broken.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
```diff
@@ -72,563 +72,597 @@ class HankeControllerITests(@Autowired override val mockMvc: MockMvc) : Controll
...
+            }
+        }
 
-    @Test
-    fun `create hanke with perustaja without sahkoposti returns 400`() {
-        val hakemus = HankeFactory.create().withPerustaja()
-        val content: ObjectNode = OBJECT_MAPPER.valueToTree(hakemus)
-        (content.get("perustaja") as ObjectNode).remove("sahkoposti")
+        @Test
+        @Disabled("Seems like this is not implemented yet.")
```
pitkni, line 376: At a quick glance, this may not be the only missing validation. If that is the case, we should create a ticket of the required fix. Otherwise it is likely that this test will remain disabled for quite some time.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
```diff
@@ -72,563 +72,597 @@ class HankeControllerITests(@Autowired override val mockMvc: MockMvc) : Controll
...
     }
 
-    @Test
-    fun `createHanke when perustaja is provided should return provided information`() {
-        val hanke = HankeFactory.create().withPerustaja()
-        every { hankeService.createHanke(any()) } returns hanke
-        justRun { permissionService.setPermission(any(), any(), Role.KAIKKI_OIKEUDET) }
+    @Nested
+    inner class CreateHanke {
+        private val url = BASE_URL
```
pitkni, line 295: Is this necessary?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
corvidian: There are tickets for perustaja:
- [HAI-1348](https://helsinkisolutionoffice.atlassian.net/browse/HAI-1348) for saving the perustaja info in FE.
- [HAI-1350](https://helsinkisolutionoffice.atlassian.net/browse/HAI-1350) for validating that the info is set in BE.

I don't what know other missing validations you're referring to.

corvidian: Ah yes, I see now. The reason for the error is that the email field is non-nullable. So we don't need validation to catch the missing value.

The difference between HAI0003 and HAI1002 is that for HAI0003 we can't even read the request to start checking its validity. It's incorrect here; the missing email should be a validation error. But I think that would require a separate DTO with nullable fields or a custom deserializer.

I think this PR has strayed too far from the original issue already, so fixing the validation won't be done here. And as long as nothing really uses the codes, it's not critical to have the correct code in every situation, as long as the HTTP response code is correct.

I fixed this to test the current implementation and re-enabled the test.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
corvidian: Strictly speaking, it's not necessary here, `BASE_URL` could be used as-is. Still, I added it to keep the pattern with the other nested classes.

The reason I added the `url` variables was that in some test returning 404, the URL was incorrect, giving the correct HTTP response, but because of wrong reasons. Having the same URL in all tests of the same endpoint makes sure this can't happen.

pitkni (approved):

PR #382, opened 2023-08-07
HAI-1646 Update README.md

pitkni:
# Description

Updated README.md and deleted services/hanke-service/README.md. Auth-service will soon be removed, so the README can be located at the root dir. 

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing

# Checklist:

 or other location: 

# Other relevant info


corvidian:
Is it possible to first remove the root's README.md, and then move the edited file to the root with `git mv`? It would be nice to keep the history of the file.

pitkni:
> Is it possible to first remove the root's README.md, and then move the edited file to the root with `git mv`? It would be nice to keep the history of the file.
> 
> Currently, there's no way to say what has been updated.

Is it better now? I don't know why I can't reply to your comment.

corvidian:
> Is it better now?
Looks better now.

> I don't know why I can't reply to your comment.
For some reason, GitHub handles main-level comments differently from comments pointing to source code.

README.md
```diff
@@ -1,30 +1,201 @@
...
-$ ./gradlew build
+
+#### Usage examples
+
+| Command                                                                                 | Explanation                                                                      |
+|-----------------------------------------------------------------------------------------|----------------------------------------------------------------------------------|
+| `docker-compose up -d`                                                                  | Run the entire stack, -d can be omitted                                          |
+| `docker-compose up db haitaton-hanke`                                                   | Run individual services                                                          |
+| `docker-compose stop haitaton-ui && docker-compose up -d --build --no-deps haitaton-ui` | Rebuild and run an individual service                                            |
+| `docker-compose down`                                                                   | Stop containers and removes containers networks volumes and images created by up |
```
corvidian, line 44: I find this really annoying to read in the editor. It's way too wide. Maybe some other solution than a table would be better?

pitkni: I agree that it is annoying to read in the editor, but in the actual usage it looks pretty good to me.
image


I can try bullet points, if the are ok?

corvidian: I mostly look at the README in the editor, or otherwise like in GitHub diffs, as raw markdown. So I think both the markdown and the rendered view are important to be readable.

Maybe nested bullet points?
```
- `docker-compose up -d`
  - Run the entire stack, -d can be omitted
- `docker-compose up db haitaton-hanke`
  - Run individual services
- `docker-compose stop haitaton-ui && docker-compose up -d --build --no-deps haitaton-ui`
  - Rebuild and run an individual service
- `docker-compose down`
  - Stop containers and removes containers networks volumes and images created by up
```

corvidian (approved):

PR #385, opened 2023-08-10
HAI Endpoint for unlinking applications from allu

corvidian:
# Description

Endpoint for unlinking applications from Allu on dev and test environments.


corvidian:
Opened as #387

PR #387, opened 2023-08-11
HAI-1842 Add endpoint for unlinking applications from Allu in dev and test

corvidian:
# Description

Allu removes all of its data in the test environment when it's updated. To mitigate problems this causes to Haitaton dev and test environments, add an endpoint that removes the linking between Haitaton applications and Allu. The links are non-functional after an Allu update anyway.

Make the endpoint callable directly from Swagger UI, so it calling it doesn't require a developer.

The endpoint is only visible and accessible from specific environments, namely dev and test environments. This is accomplished with environment variables.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Create an application and then call the endpoint to unlink it.

# Checklist:

 or other location: TODO

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/testdata/TestDataServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/testdata/TestDataServiceITest.kt
@@ -1,0 +1,77 @@
+package fi.hel.haitaton.hanke.testdata
+
+import assertk.assertThat
+import assertk.assertions.each
+import assertk.assertions.hasSize
+import assertk.assertions.isNull
+import assertk.assertions.isTrue
+import fi.hel.haitaton.hanke.DatabaseTest
+import fi.hel.haitaton.hanke.allu.ApplicationStatus
+import fi.hel.haitaton.hanke.application.ApplicationRepository
+import fi.hel.haitaton.hanke.factory.AlluDataFactory
+import fi.hel.haitaton.hanke.factory.HankeFactory
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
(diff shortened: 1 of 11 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/testdata/TestDataServiceITest.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    @Autowired private lateinit var testDataService: TestDataService
+    @Autowired private lateinit var applicationRepository: ApplicationRepository
+    @Autowired private lateinit var alluDataFactory: AlluDataFactory
+    @Autowired private lateinit var hankeFactory: HankeFactory
+
+    @Nested
+    inner class UnlinkApplicationsFromAllu {
+        @Test
+        fun `Without applications does nothing`() {
```
pitkni, line 40: Does this actually verify that it does not do anything?

corvidian: The important thing is that it doesn't throw an exception with an empty list. I renamed the test.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/AccessRules.kt
```diff
@@ -26,6 +26,11 @@ object AccessRules {
                 "/v3/api-docs/**",
             )
             .permitAll()
+            .mvcMatchers(HttpMethod.POST, "/testdata/unlink-applications")
```
pitkni, line 29: Would it be better to require an access token and log the user doing this operation? I don't have a clear opinion myself so I'm just pondering.

corvidian: Requiring an access token would in practice limit the use to developers.

pitkni (approved):

PR #389, opened 2023-08-14
HAI-1843 Add new Dockerfile for platta

pitkni:
# Description

Platta requires a new Dockerfile to build the application and create the runtime image in the same file.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/Dockerfile-platta
+++ b/services/hanke-service/Dockerfile-platta
@@ -1,0 +1,24 @@
+# Uses different base images for build and run. RedHat Ubi stronger security policies prevent gradle builds.
+
+# Stage 1: Build application
+FROM eclipse-temurin:17 as build
+
+# Set the working directory, copy gradle and source code
+WORKDIR /app
+COPY gradlew settings.gradle.kts ./
+COPY gradle/ ./gradle/
+COPY services/hanke-service/build.gradle.kts ./services/hanke-service/
+COPY services/hanke-service/src ./services/hanke-service/src
+COPY scripts/wait-for-it.sh .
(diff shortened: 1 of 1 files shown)


services/hanke-service/Dockerfile-platta
```diff
@@ -0,0 +1,24 @@
...
+# Stage 1: Build application
+FROM eclipse-temurin:17 as build
+
+# Set the working directory, copy gradle and source code
+WORKDIR /app
+COPY gradlew settings.gradle.kts ./
+COPY gradle/ ./gradle/
+COPY services/hanke-service/build.gradle.kts ./services/hanke-service/
+COPY services/hanke-service/src ./services/hanke-service/src
+COPY scripts/wait-for-it.sh .
```
corvidian, line 12: I don't think we need this script for anything? It's used in Docker compose to wait for e.g. database to start before starting the backend service, but the database and other services are running constantly on Platta.

pitkni: I had a hunch that it may not be needed, but it was included in the original Dockerfile. Removed that line.

corvidian: There's another row with this, which would probably fail now that the other was removed.
```
COPY --from=build /app/wait-for-it.sh .
```

pitkni: Thanks, missed that one.

corvidian (approved):

PR #390, opened 2023-08-18
HAI-1844: Create hankekayttaja instance for Hanke perustaja

pitkni:
# Description

Add functionality to include Hanke perustaja in hanke kayttaja listing. Hanke perustaja is now mandatory information in Hanke creation. The not-null requirement is enforced in the database also. Existing entries will have empty string for name and email.

A hanke generated from a cable report is a special case. In these cases perustaja is generated from the ordered person in application contacts.

Additional: permission for hanke creator is now set in the service layer as the permission entity is required for token creation.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

- Create johtoselvitys without and existing hanke. After first page, check:
  -  GET http://localhost:3001/api/hankkeet/{HANKETUNNUS}/kayttajat 
  - Should contain a perustaja record similar to:
```{
  "kayttajat": [
    {
      "id": "51e03361-d313-4093-a365-172f0a2817ac",
      "sahkoposti": "foo.bar@testmail.com",
      "nimi": "Matti Meikäläinen",
      "rooli": "KAIKKI_OIKEUDET",
      "tunnistautunut": true
    }
  ]
}
```
- Continue to sending the application. After application is sent, check:
  - GET http://localhost:3001/api/hankkeet/{HANKETUNNUS}/kayttajat 
  - Should contain all contacts but they should have KATSELUOIKEUS and tunnistautunut = false.
# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -151,12 +152,20 @@ open class HankeServiceImpl(
...
+            initAccessForCreatedHanke(it, userId)
             hankeLoggingService.logCreate(it, userId)
         }
     }
 
+    private fun initAccessForCreatedHanke(hanke: Hanke, userId: String) {
+        val hankeId = hanke.id!!
+        val permissionAll = permissionService.setPermission(hankeId, userId, Role.KAIKKI_OIKEUDET)
+        logger.info { "Saving token for Hanke perustaja." }
+        hankeKayttajaService.createToken(hankeId, hanke.perustaja.toUserContact(), permissionAll)
```
corvidian, line 165: Perustaja doesn't need a token, since they're getting KAIKKI_OIKEUDET permissions directly. A token is only needed for activating the user by creating the permission row later when we know which user they're supposed to be.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -80,7 +80,7 @@ data class Hanke(
     //
     @JsonView(ChangeLogView::class)
     @field:Schema(description = "Hanke founder contact information")
-    var perustaja: Perustaja? = null,
+    var perustaja: Perustaja,
```
corvidian, line 83: I think we can't merge this until https://helsinkisolutionoffice.atlassian.net/browse/HAI-1348 has been done in frontend. Otherwise, we'll break hanke creation from the UI, if I'm not mistaken?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: Fixed the implementation so that perustaja does not get a kayttajatunniste.

services/hanke-service/src/main/resources/db/changelog/changesets/043-change-hanke-perustaja-non-nullable.yaml
```diff
@@ -0,0 +1,21 @@
...
+            columnName: perustajanimi
+            defaultValue: ""
+        - addDefaultValue:
+            tableName: hanke
+            columnName: perustajaemail
+            defaultValue: ""
+        - addNotNullConstraint:
+            tableName: hanke
+            columnName: perustajanimi
+            defaultNullValue: ""
```
corvidian, line 17: I don't understand why we're changing a clear indicator for a missing value (null value) to an unclear one.

Also for the name, we know we can't get the name in every situation.

pitkni: Would it be better to remove the constraint on the db / entity level? I.e. keep the objects as they are and just validate the the request in the controller?

corvidian: If we want to add the constraints at least on the email, we could wipe the current hanke_kayttaja and kayttaja_tunniste. They're not really usable anyway.

Having a tight constraint on the name will cause problems when we use Helsinki AD authentication (like we do currently in the dev environment).

corvidian: Yes, I think we can't add the constraints to DB. We can't figure out the founder after the fact, so the unknown ones should be null. Or maybe we could do a complex query to find the contact with the orderer flag from every hanke since all hanke in production are cable report applications. 

Having a tight constraint on the name might cause problems when we use Helsinki AD authentication (like we do currently in the dev environment, and probably for local development).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
pitkni: Perustaja field is nullable again, but yes: this cannot be merged before the frontend is implemented. Creating a hanke without Perustaja will fail validation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -235,21 +236,22 @@ open class HankeServiceImpl(
...
             logger.info { "Hakemus ${it.id} has alluStatus ${it.alluStatus}" }
             !applicationService.isStillPending(it)
         }
 
+    private fun initAccessForCreatedHanke(hanke: Hanke, userId: String) {
+        val hankeId = hanke.id!!
+        val permissionAll = permissionService.setPermission(hankeId, userId, Role.KAIKKI_OIKEUDET)
+        val perustaja = hanke.perustaja
+        if (perustaja != null) {
+            hankeKayttajaService.addHankeFounder(hankeId, perustaja, permissionAll)
```
corvidian, line 250: So this still creates a user for every created hanke if the caller gives a perustaja?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
```diff
@@ -67,19 +63,21 @@ class HankeKayttajaService(
...
         logger.info { "Saved the new user token, id=${kayttajaTunnisteEntity.id}" }
 
         saveUser(
-            HankeKayttajaEntity(
-                hankeId = hankeId,
-                nimi = contact.name,
-                sahkoposti = contact.email,
-                permission = null,
-                kayttajaTunniste = kayttajaTunnisteEntity
+            hankeKayttajaRepository.save(
```
corvidian, line 73: `saveUser` has `hankeKayttajaRepository.save(hankeKayttajaEntity)` as it's first expression, so adding it here is quite redundant.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -114,8 +116,7 @@ class HankeServiceITests : DatabaseTest() {
 
     @Test
     fun `create Hanke with full data set succeeds and returns a new domain object with the correct values`() {
-        val hanke: Hanke =
-            getATestHanke().withYhteystiedot { it.id = null }.withPerustaja().withHankealue()
+        val hanke: Hanke = getATestHanke().withYhteystiedot { it.id = null }.withHankealue()
```
corvidian, line 119: There should be a test with the perustaja data. Readding it to the test data here causes a failure at
```
assertThat(hankeKayttajaRepository.findAll()).hasSize(4)
```

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeFactory.kt
```diff
@@ -99,6 +100,12 @@ class HankeFactory(
                 hankeStatus,
             )
 
+        /** Create minimal Entity with identifier fields and mandatory fields. */
+        fun createNewEntity(
```
corvidian, line 104: The name could tell the difference between this and `create()`.
```suggestion
        fun createMinimal(
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: Idea is that the perustaja is defined by the backend (before this step). It is not yet implemented as this pr will only focus on johtoselvitys/hanke generation. At least that is how I understood things based on the discussion last week. 

I can add a validator to make sure that the client does not set this field.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
pitkni: There is a save operation and a log operation. I'm not sure whether the function is redundant or not.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeFactory.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -241,6 +242,16 @@ open class HankeServiceImpl(
             !applicationService.isStillPending(it)
         }
 
+    private fun initAccessForCreatedHanke(hanke: Hanke, userId: String) {
+        val hankeId = hanke.id!!
+        val permissionAll = permissionService.setPermission(hankeId, userId, Role.KAIKKI_OIKEUDET)
+        val perustaja = hanke.perustaja
+        if (perustaja != null) {
```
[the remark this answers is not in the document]
pitkni: Done.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
pitkni: There is now two tests. One including perustaja and one with no perustaja and no contacts.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
corvidian: I don't think the saveUser is redundant, saving the entity here before passing it to saveUser is.

pitkni: Oops, yes now I see the issue.

pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
corvidian: Ah yes, I didn't notice that the controller wiped the perustaja. I don't know if this is something the controller should be doing, it feels very much like business logic.

We could really use a separate DTO for hanke creation since there are so many fields we need to ignore.

Do we need perustaja in Hanke? With the current understanding, we don't want to add it to any responses and we don't want it to be modifiable with PUT requests. We could add it directly to the entity in `createHanke` and then forget about it. This would require we add `Perustaja?` as a separate parameter to `createHanke`. I think this would also make the logic surrounding the perustaja more transparent. For now, we could make the perustaja null in the controller call. When adding the hankekayttaja for all hanke, we could add a method in front of createHanke that finds the perustaja from the contacts.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/PermissionServiceITest.kt
```diff
@@ -20,17 +21,19 @@ import org.springframework.security.test.context.support.WithMockUser
 import org.springframework.test.context.ActiveProfiles
 import org.testcontainers.junit.jupiter.Testcontainers
 
+private const val HANKE_TUNNUS = HankeFactory.defaultHankeTunnus
+private const val USERNAME = "user"
```
corvidian, line 25: This pattern is mostly used to notate the current user, i.e. the user that's used as `@WithMockUser`. Here, the user needs to be different, so this pattern might not be optimal.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/PermissionServiceITest.kt
```diff
@@ -118,56 +121,56 @@ class PermissionServiceITest : DatabaseTest() {
...
         )
 
-        assertFalse(permissionService.hasPermission(hankeId, username, PermissionCode.EDIT))
+        assertFalse(permissionService.hasPermission(hankeId, USERNAME, PermissionCode.EDIT))
     }
 
     @Test
     fun `setPermission creates a new permission`() {
-        val hankeId = saveSeveralHanke(1)[0].id!!
+        val hankeId = saveSeveralHanke(listOf(HANKE_TUNNUS))[0].id!!
```
corvidian, line 171: Feels like there's a lot of changes in this file. They could be avoided by adding a delete after the hanke creation in the failing tests.

```
permissionRepository.deleteAll()
```

pitkni: Fixed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/PermissionServiceITest.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: A separate dto for these requests would be a major improvement, I agree.

I removed Perustaja from the domain object as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -309,12 +305,15 @@ class HankeController(
...
 
-    private fun Hanke.validateUpdatable(updatedHanke: Hanke, hankeTunnusFromPath: String) {
-        if (hankeTunnusFromPath != updatedHanke.hankeTunnus) {
-            throw HankeArgumentException("Hanketunnus not given or doesn't match the hanke data")
-        }
-        if (perustaja != null && perustaja != updatedHanke.perustaja) {
-            throw HankeArgumentException("Updating perustaja not allowed.")
+    private fun validateUpdatable(existing: Hanke, updated: Hanke, hankeTunnusFromPath: String) {
+        val tunnusMatch =
+            listOf(existing.hankeTunnus, updated.hankeTunnus).all { it == hankeTunnusFromPath }
```
corvidian, line 310: Since the existing hanke is loaded with the hanketunnus from the path, they can't be different. So there's no reason to check that they're the same.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
```diff
@@ -48,22 +49,39 @@ class HankeKayttajaService(
         filterNewContacts(hankeId, contacts).forEach { contact -> createToken(hankeId, contact) }
     }
 
+    @Transactional
+    fun addHankeFounder(hankeId: Int, founder: Perustaja, permissionEntity: PermissionEntity) {
+        logger.info { "Saving token for Hanke perustaja." }
```
corvidian, line 54: This is adding a user, not a token.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -242,12 +244,16 @@ open class HankeServiceImpl(
...
+        perustaja: PerustajaEntity?,
+        userId: String
+    ) {
         val hankeId = hanke.id!!
         val permissionAll = permissionService.setPermission(hankeId, userId, Role.KAIKKI_OIKEUDET)
-
-        hanke.perustaja?.let { hankeKayttajaService.addHankeFounder(hankeId, it, permissionAll) }
-
+        perustaja?.let {
+            hankeKayttajaService.addHankeFounder(hankeId, it.toDomainObject(), permissionAll)
```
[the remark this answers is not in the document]
pitkni: I just usually prefer the actual saved objects in these scenarios, but I guess the original domain object is valid also. Changed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
pitkni: Fixed.

corvidian (approved):

PR #391, opened 2023-08-22
HAI-1527 API for changing permissions

corvidian:
# Description

Add API for changing the permissions of hanke kayttajat.

Audit logging will be added in a separate PR.

Add a security scheme to the OpenAPI configuration. This enables the use of the Swagger UI by copying over your bearer token. Reformat some OpenAPI descriptions, since they were shown as code blocks.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
You can use Swagger UI to send requests to test the API.

# Checklist:

 or other location: README for Swagger UI authentication

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
@@ -2,35 +2,47 @@ package fi.hel.haitaton.hanke.permissions
 
 import fi.hel.haitaton.hanke.HankeError
 import fi.hel.haitaton.hanke.HankeService
+import fi.hel.haitaton.hanke.configuration.Feature
+import fi.hel.haitaton.hanke.configuration.FeatureFlags
 import fi.hel.haitaton.hanke.currentUserId
 import fi.hel.haitaton.hanke.logging.DisclosureLogService
+import fi.hel.haitaton.hanke.validation.ValidHanke
+import io.swagger.v3.oas.annotations.Hidden
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.media.Content
 import io.swagger.v3.oas.annotations.media.Schema
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.responses.ApiResponses
+import io.swagger.v3.oas.annotations.security.SecurityRequirement
 import mu.KotlinLogging
+import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
+import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 
 private val logger = KotlinLogging.logger {}
 
 @RestController
 @RequestMapping("/hankkeet/{hankeTunnus}/kayttajat")
+@SecurityRequirement(name = "bearerAuth")
 class HankeKayttajaController(
     private val hankeService: HankeService,
     private val hankeKayttajaService: HankeKayttajaService,
     private val permissionService: PermissionService,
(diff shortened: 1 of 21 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/Permissions.kt
```diff
@@ -27,6 +27,8 @@ enum class PermissionCode(val code: Long) {
 interface PermissionRepository : JpaRepository<PermissionEntity, Int> {
     fun findOneByHankeIdAndUserId(hankeId: Int, userId: String): PermissionEntity?
 
+    fun findAllByHankeId(hankeId: Int): List<PermissionEntity>
```
pitkni, line 30: Do we need an index for hankeid fk?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -71,4 +80,124 @@ class HankeKayttajaController(
...
+                    content = [Content(schema = Schema(implementation = HankeError::class))]
+                ),
+            ]
+    )
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    fun updatePermissions(
+        @ValidHanke @RequestBody permissions: PermissionUpdate,
+        @PathVariable hankeTunnus: String
+    ) {
+        // featureFlags.ensureEnabled(Feature.PERMISSION_EDITING)
```
pitkni, line 138: Unused code should be deleted and can be retrieved from source control history if required.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -71,4 +80,124 @@ class HankeKayttajaController(
...
+        val hanke = hankeService.findHankeOrThrow(hankeTunnus)
+
+        permissionService.verifyHankeUserAuthorization(
+            userId = userId,
+            hanke = hanke,
+            permissionCode = PermissionCode.MODIFY_EDIT_PERMISSIONS
+        )
+
+        val deleteAdminPermission =
+            permissionService.hasPermission(
```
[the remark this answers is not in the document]
corvidian: Yes, but it felt like breaking separation of concerns. All other authorization checks are done in the controller layer, so I didn't want to push this to the service layer.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
corvidian: Oh yeah, I meant to ask about this, whether we need a feature flag for this. I'm guessing we still have some time until we're ready to release the user management feature.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/Permissions.kt
corvidian: It might be useful. Though, I think having an index for the `findOneByHankeIdAndUserId` is more important, since that query is done on every request.

I added indices for both of these.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
pitkni: Hmm, I guess we should follow the same procedure as with hanke. So let's use a feature flag for edit endpoints.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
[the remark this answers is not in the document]
pitkni: This authorisation check is actually done in the service layer, you just get the value to check here. But I do not have a better solution for this. We can keep it as is.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
corvidian: I added the feature flag.

In another commit, I went a bit further and used the feature flag to stop the creation of user tokens until we are ready with the user management feature. The tokens really have any meaning if we don't send the invitation emails.

pitkni (approved):

PR #396, opened 2023-08-28
HAI-1873 Add API for identifying a user

corvidian:
# Description

Add an endpoint that takes a user token in the request body and creates a permission for the user to the hanke the token is associated with.

The token is deleted from DB after use, since it can't be used again anyway.

The motivation is to activate the permissions that have been created for a contact person. When creating a contact person we don't know which user the contact person is, so we create a user token, that's emailed to the contact. The contact can then use this endpoint (through UI) to activate the given permission, i.e. change the token placeholder to an actual permission.

As an unreleated change, configure springdoc to use application/json as the default for all API descriptions. This way, we don't have to specify the media type for every single response, including all the error responses.

Also, configure springdoc to use a reference for the HankeError enum, so the definition isn't copied for each error response in the OpenAPI JSON. This cut down the size of the JSON file to about half.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
@@ -16,7 +16,6 @@ import io.swagger.v3.oas.annotations.responses.ApiResponses
 import io.swagger.v3.oas.annotations.security.SecurityRequirement
 import mu.KotlinLogging
 import org.springframework.http.HttpStatus
-import org.springframework.http.MediaType.APPLICATION_JSON_VALUE
 import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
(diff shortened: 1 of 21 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
```diff
@@ -91,6 +93,62 @@ class HankeKayttajaService(
         hankeKayttajaRepository
             .findByHankeIdAndSahkopostiIn(hankeId, contacts.map { it.email })
             .map { it.sahkoposti }
+
+    @Transactional
+    fun createPermissionFromToken(userId: String, tunniste: String) {
```
pitkni, line 98: Declaring public functions before private ones would improve readability.

pitkni:
This pr is for main. Change it to dev.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -39,7 +38,7 @@ class HankeKayttajaController(
     private val featureFlags: FeatureFlags,
 ) {
 
-    @GetMapping(produces = [APPLICATION_JSON_VALUE])
+    @GetMapping("/hankkeet/{hankeTunnus}/kayttajat", produces = [APPLICATION_JSON_VALUE])
```
pitkni, line 41: produces = [APPLICATION_JSON_VALUE] can this be removed now? Referring to application.properties change.

corvidian: I removed all explicit produces tags and added a check for all GET request tests that the response content type is correct.

pitkni (approved):

PR #398, opened 2023-09-01
HAI Allow for custom git pre-push hooks

corvidian:
# Description

I want to have a custom pre-push git hook, which isn't useful for other people. If I changed the existing one, it would get overwritten by the gradle script.

Replace the pre-push script with a script that runs all scripts from a folder. Add the maintained pre-push script as one of the files in that folder.

That is,
- `githooks/pre-push` is installed as `.git/hooks/pre-push`
- All files under `githooks/pre-push.d` are copied under `.git/hooks/pre-push.d
- A developer can add custom scripts to `.git/hooks/pre-push.d` and they won't be overwritten by the gradle build script
- `.git/hooks/pre-push` runs all scripts under `.git/hooks/pre-push.d` and fails if any of them fail

With this structure, developers can have custom hooks while still keeping the build-hook maintained in version control.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Checklist:

 or other location: README

### DIFF
--- a/githooks/pre-push
+++ b/githooks/pre-push
@@ -1,3 +1,16 @@
 #!/bin/sh
 
-./gradlew build integrationTest
+#!/bin/bash
+
+hooks="$(dirname "$0")/pre-push.d"
+
+for hook in ${hooks}/*; do
+    bash $hook
+    RESULT=$?
+    if [ $RESULT != 0 ]; then
+        echo "pre-push.d/$hook returned non-zero: $RESULT, abort push"
+        exit $RESULT
+    fi
+done
+
+exit 0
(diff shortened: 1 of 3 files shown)


githooks/pre-push
```diff
@@ -1,3 +1,16 @@
 #!/bin/sh
 
-./gradlew build integrationTest
+#!/bin/bash
```
pitkni, line 3: Is there a reason to have two shebangs here?

corvidian: Thank you, this was a copy-paste mistake. Removed the latter one.

pitkni (approved):

PR #400, opened 2023-09-05
HAI-1875 Add audit logging for new users and tokens

corvidian:
# Description

Add logging for creating permissions, hankekayttaja and hanketunniste.

Adding the audit reasonably required some refactoring:
- In PermissionService, make the old setPermission function be explicitly for creating new permisisions, so logging can be added.
- Move updating the kayttooikeustaso of a permission to PermissionService to centralize modifications and logging.
- In HankeKayttajaService, make the saveUser function to handle creating new users explicitly so that logging can be added.
- Rename kayttooikeustaso in PermissionEntity to kayttooikeustasoEntity and add a new getter for the actual kayttooikeustaso, so that `permissionEntity.kayttooikeustaso` is required instead of the old `permissionEntity.kayttooikeustaso.kayttooikeustaso`.
- Change references to permissionEntity.kayttooikeustasoEntity to use either the new getter for reads or PermissionService functions for modifications where ever possible.
- Move creating the permission for a founder inside addHankeFounder, since it seemed like a better packaging and made testing the audit logs easier.
- Move the static hasPermission from PermissionService's companion object to PermissionEntity and KayttooikeustasoEntity.
- Add userIds to several calls se that the logging can use it.
- Move Permission logging to PermissionLoggingService from HankeKayttajaLoggingService.
- In tests, use custom asserts to make audit log tests more readable and props to give better error messages when a test fails.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
- Create and send a johtoselvityshakemus.
- Check that there are creates for hankekayttaja and kayttajatunniste and one create for permission in the database.
 
# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
@@ -973,14 +973,6 @@ class HankeKayttajaServiceITest : DatabaseTest() {
         k.prop(HankeKayttajaEntity::kayttajaTunniste).isNotNull()
     }
 
-    private fun Assert<PermissionEntity>.isOfEqualDataTo(other: PermissionEntity) =
-        given { actual ->
-            assertThat(actual.id).isEqualTo(other.id)
-            assertThat(actual.hankeId).isEqualTo(other.hankeId)
-            assertThat(actual.userId).isEqualTo(other.userId)
-            assertThat(actual.kayttooikeustaso).isEqualTo(other.kayttooikeustaso)
-        }
-
     private val expectedNames =
         arrayOf(
             "yhteys-etu1 yhteys-suku1",
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
@@ -7,6 +7,7 @@ import assertk.assertThat
 import assertk.assertions.containsExactly
 import assertk.assertions.containsExactlyInAnyOrder
 import assertk.assertions.each
+import assertk.assertions.exactly
 import assertk.assertions.first
 import assertk.assertions.hasClass
 import assertk.assertions.hasSize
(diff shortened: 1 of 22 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
```diff
@@ -844,36 +956,29 @@ class HankeKayttajaServiceITest : DatabaseTest() {
...
-        k.transform { it.kayttajaTunniste }.isNotNull()
+        k.prop(HankeKayttajaEntity::id).isNotNull()
+        k.prop(HankeKayttajaEntity::nimi).isIn(*expectedNames)
+        k.prop(HankeKayttajaEntity::sahkoposti).isIn(*expectedEmails)
+        k.prop(HankeKayttajaEntity::hankeId).isEqualTo(hankeId)
+        k.prop(HankeKayttajaEntity::permission).isNull()
+        k.prop(HankeKayttajaEntity::kayttajaTunniste).isNotNull()
     }
 
     private fun Assert<PermissionEntity>.isOfEqualDataTo(other: PermissionEntity) =
```
pitkni, line 976: This seems to be an unused function now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/PermissionService.kt
```diff
@@ -16,43 +23,26 @@ class PermissionService(
             it.hankeId
         }
 
-    fun hasPermission(hankeId: Int, userId: String, permission: PermissionCode): Boolean {
-        val kayttooikeustaso =
-            permissionRepository.findOneByHankeIdAndUserId(hankeId, userId)?.kayttooikeustaso
-        return hasPermission(kayttooikeustaso, permission)
-    }
+    fun hasPermission(hankeId: Int, userId: String, permission: PermissionCode): Boolean =
```
pitkni, line 26: Do you think there is a way to make these work with hankeTunnus instead of hankeId? Currently the way the majority of our endpoints work is that we receive hankeTunnus in the controller, then find hankeId and then call this service. This is somewhat unrelated to this pr, but it is something that I've been wondering.

corvidian: An easy solution would be to overload this:
```
fun hasPermission(hankeTunnus: String, userId: String, permission: PermissionCode): Boolean =
    hasPermission(hankeService.getHankeId(hankeTunnus), userId, permission)
```
Or maybe we would need to use hankeRepository instead of the service to avoid a circular dependency.

In a broader view, there's not much point in having both hankeId and hankeTunnus, since hankeTunnus is created for every hanke at first save. We could refactor the database and code to use hankeTunnus as the primary key for the hanke and get rid of hankeId. This wouldn't be the simplest refactoring, but it might be simpler than it seems at first glance.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeFactory.kt
```diff
@@ -55,6 +57,12 @@ class HankeFactory(
     }
 
     fun save(hanke: Hanke) = hankeService.createHanke(hanke)
+    fun saveMinimal(hankeTunnus: String): HankeEntity =
```
pitkni, line 60: Nitpicking, but a line break would be nice.

pitkni: Furthermore, could this be:

    fun saveMinimal(hankeTunnus: String = hanketunnusService.newHanketunnus()): HankeEntity =
        hankeRepository.save(HankeEntity(hankeTunnus = hankeTunnus))

No need for overloading then.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
corvidian: Thank you, removed.

pitkni (approved):

PR #414, opened 2023-09-07
HAI-1526 Send hanke invitation emails

pitkni:
# Description

Include the email sending feature to creation of new hanke users.

Note: this commit does not include sending of application invitations. It will be implemented on another commit.

Additional: fix a bug in hanke controller where .copy function did not copy fields that are not in the hanke constructor.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Apart from unit tests, there is no easy way of verifying this. Verifying the outgoing email would require a hanke with contact persons. That depends on: https://helsinkisolutionoffice.atlassian.net/browse/HAI-1586 . Furthermore, having a hankekayttaja for hanke creator depends on this: https://helsinkisolutionoffice.atlassian.net/browse/HAI-1348

However, this feature is guarded by a feature flag so it should be safe to have some of this as unused code for now.

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
@@ -7,6 +7,7 @@ import fi.hel.haitaton.hanke.factory.AlluDataFactory
 import fi.hel.haitaton.hanke.factory.DateFactory
 import fi.hel.haitaton.hanke.factory.HankeFactory
 import fi.hel.haitaton.hanke.factory.HankeFactory.Companion.withGeneratedOmistaja
+import fi.hel.haitaton.hanke.factory.HankeFactory.Companion.withYhteystiedot
 import fi.hel.haitaton.hanke.geometria.Geometriat
 import fi.hel.haitaton.hanke.logging.DisclosureLogService
 import fi.hel.haitaton.hanke.permissions.PermissionCode
@@ -329,14 +330,19 @@ class HankeControllerITests(@Autowired override val mockMvc: MockMvc) : Controll
 
         @Test
         fun `Sanitize hanke input and return 200`() {
-            val hanke = HankeFactory.create().apply { generated = true }
-            every { hankeService.createHanke(hanke.copy(id = null, generated = false)) } returns
-                hanke.copy(generated = false)
+            val hanke = HankeFactory.create().withYhteystiedot().apply { generated = true }
+            val expectedServiceArgument =
+                hanke.apply {
+                    generated = false
+                    id = null
+                }
+            every { hankeService.createHanke(expectedServiceArgument) } returns
+                expectedServiceArgument
 
             post(url, hanke).andExpect(status().isOk)
 
             verifySequence {
-                hankeService.createHanke(any())
+                hankeService.createHanke(expectedServiceArgument)
                 disclosureLogService.saveDisclosureLogsForHanke(any(), any())
             }
         }
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
@@ -332,10 +332,7 @@ class HankeControllerITests(@Autowired override val mockMvc: MockMvc) : Controll
         fun `Sanitize hanke input and return 200`() {
             val hanke = HankeFactory.create().withYhteystiedot().apply { generated = true }
             val expectedServiceArgument =
-                hanke.apply {
-                    generated = false
-                    id = null
-                }
+                HankeFactory.create(id = null).withYhteystiedot().apply { generated = false }
             every { hankeService.createHanke(expectedServiceArgument) } returns
                 expectedServiceArgument
 
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
@@ -132,14 +132,14 @@ class HankeKayttajaServiceITest : DatabaseTest() {
     }
 
     @Nested
-    inner class GetKayttajaByCurrentUser {
+    inner class GetKayttajaByUserId {
 
         @Test
         fun `When user exists should return current hanke user`() {
             val hankeWithApplications = hankeFactory.saveGenerated(userId = USERNAME)
 
             val result: HankeKayttajaDto? =
-                hankeKayttajaService.getKayttajaByUserId(hankeId = hankeWithApplications.hanke.id!!)
+                hankeKayttajaService.getKayttajaByUserId(hankeWithApplications.hanke.id!!, USERNAME)
 
             assertThat(result).isNotNull()
             with(result!!) {
@@ -153,7 +153,7 @@ class HankeKayttajaServiceITest : DatabaseTest() {
 
         @Test
         fun `When no hanke should return null`() {
-            val result: HankeKayttajaDto? = hankeKayttajaService.getKayttajaByUserId(hankeId = 123)
+            val result: HankeKayttajaDto? = hankeKayttajaService.getKayttajaByUserId(123, USERNAME)
 
             assertThat(result).isNull()
         }
@@ -164,7 +164,7 @@ class HankeKayttajaServiceITest : DatabaseTest() {
             permissionRepository.deleteAll()
 
             val result: HankeKayttajaDto? =
-                hankeKayttajaService.getKayttajaByUserId(hankeId = hanke.id!!)
+                hankeKayttajaService.getKayttajaByUserId(hanke.id!!, USERNAME)
 
             assertThat(result).isNull()
         }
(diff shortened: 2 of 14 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
```diff
@@ -329,14 +330,19 @@ class HankeControllerITests(@Autowired override val mockMvc: MockMvc) : Controll
...
         fun `Sanitize hanke input and return 200`() {
-            val hanke = HankeFactory.create().apply { generated = true }
-            every { hankeService.createHanke(hanke.copy(id = null, generated = false)) } returns
-                hanke.copy(generated = false)
+            val hanke = HankeFactory.create().withYhteystiedot().apply { generated = true }
+            val expectedServiceArgument =
+                hanke.apply {
+                    generated = false
+                    id = null
+                }
```
corvidian, line 338: Apply changes the object, doesn't create a new one.
I.e. these asserts fail:
```kotlin
            val hanke = HankeFactory.create().withYhteystiedot().apply { generated = true }
            val expectedServiceArgument =
                hanke.apply {
                    generated = false
                    id = null
                }
            assertThat(hanke.generated).isTrue()
            assertThat(hanke.id).isNotNull()
```
So this test is now calling the API with a sanitized hanke. It's safer to create a new hanke:

```suggestion
            val expectedServiceArgument =
                HankeFactory.create(id = null).withYhteystiedot().apply { generated = false }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -182,7 +182,11 @@ When Hanke is created:
         if (hanke == null) {
             throw HankeArgumentException("No hanke given when creating hanke")
         }
-        val sanitizedHanke = hanke.copy(id = null, generated = false)
+        val sanitizedHanke =
+            hanke.apply {
+                id = null
+                generated = false
+            }
```
corvidian, line 189: sanitizedHanke is the same object as hanke, so the new value is redundant.
```suggestion
        hanke.id = null
        hanke.generated = false
```
Though I think this should be done in the service, not the controller. There's a logical place for it now that `createHankeInternal` was split from `createHanke`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/email/EmailSenderService.kt
```diff
@@ -127,8 +127,7 @@ class EmailSenderService(
         mailSender.send(mimeMessage)
     }
 
-    private fun defineInviterInfo(name: String?, email: String): String =
-        if (name.isNullOrBlank()) "Asioija $email" else "$name ($email)"
+    private fun inviterInfo(name: String, email: String): String = "$name ($email)"
```
corvidian, line 130: Name and email could be just split into two separate parameters now that there's no branching.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
```diff
@@ -59,7 +85,12 @@ class HankeKayttajaService(
                 .flatMap { it.alikontaktit }
                 .mapNotNull { userContactOrNull(it.fullName(), it.email) }
 
-        filterNewContacts(hankeId, contacts).forEach { contact -> createToken(hankeId, contact) }
+        filterNewContacts(hankeId, contacts)
+            .map { contact -> createHankeKayttaja(hankeId, contact) }
+            .also { kayttajaList ->
+                val inviter = getKayttajaByCurrentUser(hankeId)
+                sendHankeInvitationEmails(hanke, inviter, kayttajaList)
```
corvidian, line 92: Feels like sending the emails would be a logical part of `createHankeKayttaja`, since we want to send mail every time we create a new one.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
pitkni: Thanks, fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
pitkni: Fixed as suggested. Actually the correct solution here will be (BE: Hankkeen luomiseen erillinen DTO) https://helsinkisolutionoffice.atlassian.net/browse/HAI-1893

So I will leave this "sanitation" in the controller for now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
pitkni: I feel that sending the email should not be part of createHankeKayttaja (createTunnisteAndKayttaja after rebase from dev). I think it is the responsibility of saveNewTokensFromHanke to control the bigger picture.

On the other hand, I do agree that it is better to send the email every time a user is created. I changed this part accordingly. 

Additionally, createTunnisteAndKayttaja would require a Hanke as parameter. That would cause changes to saveNewTokensFromApplication. Also, the current implementation does not affect saveNewTokensFromApplication as I will implement that on another pr.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/email/EmailSenderService.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
```diff
@@ -21,12 +24,32 @@ class HankeKayttajaService(
...
     private val logService: HankeKayttajaLoggingService,
+    private val emailSenderService: EmailSenderService,
 ) {
 
     @Transactional(readOnly = true)
     fun getKayttajatByHankeId(hankeId: Int): List<HankeKayttajaDto> =
         hankeKayttajaRepository.findByHankeId(hankeId).map { it.toDto() }
 
+    @Transactional(readOnly = true)
+    fun getKayttajaByUserId(hankeId: Int, userId: String = currentUserId()): HankeKayttajaDto? {
```
corvidian, line 35: In production code, this is only called with a userId, but in tests, it's always called without one.

Maybe drop the default parameter? If at some point we need to, we can add a function that calls this one. It's not intuitive what happens when you ask something `ByUserId` without the user ID. So a separate function would be clearer on what it does.
```
fun getKayttajaForCurrentUser(hankeId: Int) = getKayttajaByUserId(hankeId, currentUserId())
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
```diff
@@ -104,6 +131,58 @@ class HankeKayttajaServiceITest : DatabaseTest() {
         }
     }
 
+    @Nested
+    inner class GetKayttajaByCurrentUser {
```
corvidian, line 135: The function was renamed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
corvidian: If we want to keep DTOs separate we shouldn't use them for internal data processing.

pitkni: Would it be ok to have:

```
    @Transactional(readOnly = true)
    fun getKayttajaByUserId(hankeId: Int, userId: String): HankeKayttajaDto? =
        getKayttajaEntityByUserId(hankeId, userId)?.toDto()
```

And then the current function would be private and uses the entity. And it is used for internal processing.

corvidian: No one needs the DTO at the moment. If someone would need it, they could just call `.toDto()`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
corvidian: Let's see how the next PR goes.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
```diff
@@ -173,11 +173,11 @@ class HankeKayttajaServiceITest : DatabaseTest() {
         fun `When no kayttaja should return null`() {
             val hankeWithApplications = hankeFactory.saveGenerated(userId = USERNAME)
             val hankeId = hankeWithApplications.hanke.id!!
-            val jou = hankeKayttajaService.getKayttajaByUserId(hankeId)!!
+            val jou = hankeKayttajaService.getKayttajaByUserId(hankeId, USERNAME)!!
```
corvidian, line 176: What's jou?

pitkni: It is my go-to placeholder for variables I don't know how to name. This is the nasty side-effect. Fixed.

corvidian (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
pitkni: Fixed.

PR #425, opened 2023-09-12
HAI Add OIDC client id to docker compose config

corvidian:
# Description

Add REACT_APP_OIDC_CLIENT_ID to docker compose config for Haitaton UI. This enables changing from Helsinki AD login to Suomi.fi login quickly without rebuilding the UI.

Just edit the value in docker-compose.yml and then restart the service without rebuilding:
```sh
docker-compose stop haitaton-ui && docker-compose up -d --no-deps haitaton-ui
```

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Checklist:

 or other location:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -107,6 +107,8 @@ services:
       dockerfile: ${FRONT_DOCKERFILE}
     image: haitaton-ui
     container_name: haitaton-ui
+    environment:
+      REACT_APP_OIDC_CLIENT_ID: haitaton-ui-dev
     ports:
       - 8000:8000
     volumes:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -107,6 +107,8 @@ services:
       dockerfile: ${FRONT_DOCKERFILE}
     image: haitaton-ui
     container_name: haitaton-ui
+    environment:
+      REACT_APP_OIDC_CLIENT_ID: haitaton-ui-dev
```
pitkni, line 111: Would it be better to have a commented line here to state the possible values?

corvidian: That's a good idea. I added some documentation here.

Also, the default was accidentally changed to Suomi.fi -authentication. I changed it back to Helsinki AD.

pitkni (approved):

PR #429, opened 2023-09-15
Hai 1514/refactor authorization

corvidian:
# Description

Refactor authorization to use separate component classes. Call the methods in those classes from `@PreAuthorize` annotations to separate authorization from business logic.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/Authorizer.kt
```diff
@@ -0,0 +1,37 @@
...
+        hankeTunnus: String,
+        permissionCode: PermissionCode
+    ): Boolean {
+        val hankeIds = hankeRepository!!.findOneByHankeTunnus(hankeTunnus)
+        authorize(hankeIds?.id, permissionCode) { HankeNotFoundException(hankeTunnus) }
+        return true
+    }
+
+    @Transactional(readOnly = false)
+    open fun authorizeHankeTunnus(hankeTunnus: String, permissionCode: String): Boolean =
```
corvidian, line 29: `permissionCode` is a String because the Spring Expression Language `@PreAuthorize` uses can only handle enums if given the full path, like:
```
    @PreAuthorize( 
        "@hankeAuthorizer.authorizeHankeTunnus(#hankeTunnus,T(fi.hel.haitaton.hanke.permissions.PermissionCode).EDIT)"
    )
```
I think this is quite unwieldy, so I thought I would use the names instead. These are all static values, so any mistakes should come up in tests.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeService.kt
```diff
@@ -10,11 +10,7 @@ interface HankeService {
 
     fun loadHanke(hankeTunnus: String): Hanke?
 
-    fun findHankeOrThrow(hankeTunnus: String): Hanke
-
-    fun getHankeId(hankeTunnus: String): Int?
-
-    fun getHankeIdOrThrow(hankeTunnus: String): Int
+    fun findIds(hankeTunnus: String): HankeIds?
```
pitkni, line 13: Could the returned object be renamed to something like HankeIdentifier? This made me confused as I thought it returning some sort of list of id by hankeTunnus.

corvidian: Good point, there were already variables named `hankeIds` that were lists of ids of Hanke.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/Authorizer.kt
```diff
@@ -0,0 +1,37 @@
...
+        authorizeHankeTunnus(hankeTunnus, PermissionCode.valueOf(permissionCode))
+}
+
+@Component
+class HankeAuthorizer(
```
pitkni, line 34: Is there a reason why only HankeAuthorizer is placed here? Would it be better to either have this in a separate file or move application and hankekayttajaauthorizers here too?

Looking at this it seemed (to me at least) that there is only one implementation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeController.kt
```diff
@@ -267,16 +256,13 @@ On update following will happen automatically:
                 )
             ]
     )
+    @PreAuthorize("@hankeAuthorizer.authorizeHankeTunnus(#hankeTunnus, 'DELETE')")
     fun deleteHanke(@PathVariable hankeTunnus: String) {
         logger.info { "Deleting hanke: $hankeTunnus" }
 
+        // TODO: Move getHankeWithApplications inside deleteHanke
```
pitkni, line 263: This is a good idea, but I recommend writing a jira task instead of todo comments in the code.

corvidian: I did the refactoring in ea7678ad7ddca7d7db1ea3c3c0011ffea9b3ed38.

I also removed HankeWithApplications completely since it wasn't really needed anywhere.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/Authorizer.kt
corvidian: Moved to hanke-package as a separate file.

Some kind of package-reconfiguration might be in order for Hanke-related classes. They are weirdly in the root package of the project, while other concepts are in their own packages.

pitkni: That seems to be because of fi.hel.haitaton.hanke being the root (would lead to hanke.hanke structure). Maybe Hanke stuff could be moved to domain package or the root should be just fi.hel.haitaton and then we could have a separate hanke package (I would prefer having haitaton as root). Maybe not as part of this pr.

pitkni (approved):

PR #434, opened 2023-09-21
HAI-1599 Remove generated hanke on application delete

pitkni:
# Description

Change application delete logic to remove generated hanke also if hanke does not contain any more applications.

Note: HankeService hanke delete operation still uses application delete which does not affect hanke. This is because hanke calling delete for application and vice versa is an unwanted situation.

Additional: 
- Added a hanke mapper to transform entities to domain objects. I think it is better to have a general purpose common mapper for this (instead of hiding the logic in private functions).
- Seems like a huge pr but 90 % of the "changes" are due to applying nested classes to application tests to have some structure.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Create johtoselvitys and delete it. The related hanke should be removed also.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
```diff
@@ -0,0 +1,64 @@
...
+import fi.hel.haitaton.hanke.domain.Hanke
+import fi.hel.haitaton.hanke.domain.HankeYhteystieto
+import fi.hel.haitaton.hanke.domain.Hankealue
+import fi.hel.haitaton.hanke.geometria.GeometriatService
+import org.springframework.stereotype.Component
+
+@Component
+class HankeMapper(private val geometriatService: GeometriatService) {
+
+    fun domainFromEntity(entity: HankeEntity): Hanke =
```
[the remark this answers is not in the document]
pitkni: I was kind of hoping that this would replace the domain object creation in the service (but not in this pr). The service implementation just seems messy. But yes, it is even worse to have two different domain object creation styles. I can copypaste the service implementation to the mapper.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -248,23 +250,32 @@ open class ApplicationService(
...
+     * delete, if the application is in Allu, and it's beyond the pending status.
+     *
+     * Furthermore, if the owning Hanke is generated and has no more applications, also deletes the
+     * Hanke.
+     */
+    @Transactional
+    open fun deleteWithOrphanGeneratedHankeRemoval(applicationId: Long, userId: String) =
+        with(getById(applicationId)) {
+            cancelIfSent(alluid, id)
+            deleteAndLog(this, userId)
```
corvidian, line 270: The beginning is exactly the same as the whole of delete. Is there a way to call delete here instead?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -291,6 +302,27 @@ open class ApplicationService(
             else -> false
         }
 
+    private fun cancelIfSent(alluid: Int?, applicationId: Long?) {
+        if (alluid == null) {
+            logger.info {
+                "Application not sent to Allu yet, simply deleting it. id=$applicationId"
```
corvidian, line 308: These log messages are no longer related to the function. I don't think there's a reason to take this out of the delete function.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
corvidian, line 16: This might need to be `@Transactional` since it calls the child entities of `entity`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: They both need to fetch the application entity. I did not want getById(applicationId) to be called twice.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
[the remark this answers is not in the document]
pitkni: Changed hanke service to use this new mapper.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Refactored, cancel and delete are now in the same function. Logging should make more sense now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
```diff
@@ -0,0 +1,84 @@
...
+                    tyomaaTyyppi = entity.tyomaaTyyppi
+                    alueet = alueList(entity.hankeTunnus, entity.listOfHankeAlueet, geometriaData)
+                    tormaystarkasteluTulos =
+                        tormaystarkasteluTulokset.firstOrNull()?.let {
+                            TormaystarkasteluTulos(it.perus, it.pyoraily, it.joukkoliikenne)
+                        }
+                }
+        }
+
+    private fun contacts(entity: HankeEntity): Map<ContactType, MutableList<HankeYhteystieto>> =
```
corvidian, line 55: Making the lists mutable isn't necessary here, since `.toMutableList()` needs to be called anyway after `.orEmpty()`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
```diff
@@ -0,0 +1,84 @@
...
+                    suunnitteluVaihe = suunnitteluVaihe,
+                    version = version,
+                    createdBy = createdByUserId,
+                    createdAt = createdAt?.zonedDateTime(),
+                    modifiedBy = modifiedByUserId,
+                    modifiedAt = modifiedAt?.zonedDateTime(),
+                    status = status,
+                    generated = generated
+                )
+                .apply {
```
corvidian, line 39: Closing `with` here would make sense, so we don't have overlapping `this`s. Like, without checking it's hard to say if `tormaystarkasteluTulokset` is a part of the `Hanke` or the `HankeEntity`.
```suggestion
                )}
                .apply {
```

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeMapperTest.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    private fun HankeEntity.yhteystietoCreatedAt(contactType: ContactType) =
+        listOfHankeYhteystieto.find { it.contactType == contactType }?.createdAt?.zonedDateTime()
+
+    private fun HankeEntity.yhteystietoModifiedAt(contactType: ContactType) =
+        listOfHankeYhteystieto.find { it.contactType == contactType }?.modifiedAt?.zonedDateTime()
+
+    private fun HankealueEntity.toDomain(): Hankealue {
```
corvidian, line 66: This is an almost exact copy of `alue()` in HankeMapper. I don't know how useful testing a method with the same code copied elsewhere is.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeMapperTest.kt
```diff
@@ -0,0 +1,92 @@
...
+            prop(Hanke::hankeTunnus).isEqualTo(entity.hankeTunnus)
+            prop(Hanke::onYKTHanke).isEqualTo(entity.onYKTHanke)
+            prop(Hanke::nimi).isEqualTo(entity.nimi)
+            prop(Hanke::kuvaus).isEqualTo(entity.kuvaus)
+            prop(Hanke::vaihe).isEqualTo(entity.vaihe)
+            prop(Hanke::suunnitteluVaihe).isEqualTo(entity.suunnitteluVaihe)
+            prop(Hanke::version).isEqualTo(entity.version)
+            prop(Hanke::createdAt).isEqualTo(entity.createdAt?.zonedDateTime())
+            prop(Hanke::createdBy).isEqualTo(entity.createdByUserId)
+            prop(Hanke::modifiedAt).isEqualTo(entity.modifiedAt?.zonedDateTime())
```
corvidian, line 45: There are no tests for `zonedDateTime()`, so these should use a literal value, not a conversion.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeFactory.kt
```diff
@@ -325,5 +374,17 @@ class HankeFactory(
...
+
+        private fun tormaysTarkastelu(
+            id: Int = 1,
+            hankeEntity: HankeEntity
+        ): TormaystarkasteluTulosEntity =
+            TormaystarkasteluTulosEntity(
+                id = id,
+                perus = Random.nextFloat(),
+                pyoraily = Random.nextFloat(),
+                joukkoliikenne = Random.nextFloat(),
```
corvidian, line 386: Using randomized values in unit tests is a bad idea. Unit tests should be repeatable and test the same thing every time. Using random values also means tests can break randomly, like in the CI pipeline.

The random values here are also outside the defined value range of the indexes. They're values between 1 and 5, when the random values are between 0 and 1.

Fuzz testing can be done separately if needed.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeFactory.kt
```diff
@@ -119,6 +133,41 @@ class HankeFactory(
...
+                    nimi = defaultNimi,
+                    kuvaus = defaultKuvaus,
+                    vaihe = SUUNNITTELU,
+                    suunnitteluVaihe = RAKENNUS_TAI_TOTEUTUS,
+                    onYKTHanke = true,
+                    version = 0,
+                    createdByUserId = defaultUser,
+                    createdAt = LocalDateTime.now().minusHours(1),
+                    modifiedByUserId = defaultUser,
+                    modifiedAt = LocalDateTime.now(),
```
corvidian, line 150: Using `now()` is effectively using a random value and I wouldn't recommend it.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
```diff
@@ -12,7 +12,12 @@ import fi.hel.haitaton.hanke.HankeService
 import fi.hel.haitaton.hanke.IntegrationTestConfiguration
 import fi.hel.haitaton.hanke.OBJECT_MAPPER
 import fi.hel.haitaton.hanke.andReturnBody
-import fi.hel.haitaton.hanke.factory.AlluDataFactory
+import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.cableReportWithoutHanke
+import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.createApplication
+import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.createApplications
+import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.createCableReportApplicationData
+import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.createCompanyCustomer
+import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.createContact
```
corvidian, line 20: I think static imports should be used sparingly. Just importing every static method you can doesn't make the code easier to read. It forces the reader to think things like "Where is this method defined?" when it could be clear from the context. This increases the cognitive load when reading code since the reader needs to keep which methods are imported in their head.

Like if you see `createApplications()` you don't know if this is a factory method or a local helper. The implications are different in each case, so this matters. So every time you encounter this call, the reader needs to remember "Oh yeah, that's a factory method."

If this is replaced with `AlluDataFactory.createApplications()` it's clear that this is a factory method and implies it produces a generic list of applications.

This doesn't mean I think static imports shouldn't be used, but there needs to be some concrete benefit to using them that outweighs the increased cognitive load.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
```diff
@@ -52,6 +59,7 @@ private const val BASE_URL = "/hakemukset"
 @WebMvcTest(ApplicationController::class)
 @Import(IntegrationTestConfiguration::class)
 @ActiveProfiles("test")
+@WithMockUser(USERNAME)
 class ApplicationControllerITest(@Autowired override val mockMvc: MockMvc) : ControllerTest {
```
corvidian, line 63: Are there any functional changes in this file or is it all just refactoring? I can't find any.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
pitkni: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
pitkni: Fixed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
pitkni: Delete checks that deleteWithOrphanGeneratedHankeRemoval is called (previously called delete).

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeMapperTest.kt
pitkni: Changed to use a different kind of function to create the expected object.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HankeMapperTest.kt
pitkni: Uses datefactory now.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeFactory.kt
pitkni: Uses datefactory now to limit randomness.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
pitkni: Brought back AlluDataFactory.createApplications.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -262,19 +262,29 @@ open class ApplicationService(
...
-                logger.info {
-                    "Application $id was the only one of a generated Hanke, removing Hanke ${hanke.hankeTunnus}."
-                }
-                hankeRepository.delete(hanke)
-                val hankeDomain = domainFrom(hanke, geometryMapFrom(hanke))
-                hankeLoggingService.logDelete(hankeDomain, userId)
+    open fun deleteWithOrphanGeneratedHankeRemoval(
+        applicationId: Long,
+        userId: String
+    ): ApplicationDeletionResultDto {
```
corvidian, line 268: It doesn't feel great to have the service produce a specific DTO like this. But returning just a boolean would be pretty misleading as well. So I guess this is fine.

pitkni: I decided to use a dto because a boolean is not very self explanatory. Is the problem in the usage of a dto, or something else? If this function may or may not delete a hanke, it is definitely beneficial to know what did or did not happen.

corvidian (approved):

PR #437, opened 2023-09-21
HAI-1872 Add validation for nimi and hankealue nimi lengths

corvidian:
# Description

The database is limited to 100 characters for nimi and hankealue. To avoid database errors, limit the length of nimi and hankealue to the same 100 characters in validation.

Also, refactor HankeValidator to reuse some of the code better. Use the Validators-methods for finding the paths with errors. Then map those paths to validator context violations.

Add comprehensive tests for HankeValidator to know the refactoring didn't change behaviour.

Refactor HankeControllerTest to use Mockk instead of Mockito. It was the last place Mockito was used. I thought I needed it for this PR, but I ended up doing a separate test class. I had already refactored it, so I decided to add the changes here anyway.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Try to create hanke with a name of over 100 characters. The error the backend gives should make sense now.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/HankeValidator.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/HankeValidator.kt
@@ -2,9 +2,17 @@ package fi.hel.haitaton.hanke.validation
 
 import fi.hel.haitaton.hanke.HankeError
 import fi.hel.haitaton.hanke.MAXIMUM_DATE
+import fi.hel.haitaton.hanke.MAXIMUM_HANKE_ALUE_NIMI_LENGTH
+import fi.hel.haitaton.hanke.MAXIMUM_HANKE_NIMI_LENGTH
 import fi.hel.haitaton.hanke.MAXIMUM_TYOMAAKATUOSOITE_LENGTH
 import fi.hel.haitaton.hanke.Vaihe
 import fi.hel.haitaton.hanke.domain.Hanke
+import fi.hel.haitaton.hanke.domain.Hankealue
+import fi.hel.haitaton.hanke.validation.Validators.isBeforeOrEqual
+import fi.hel.haitaton.hanke.validation.Validators.notLongerThan
+import fi.hel.haitaton.hanke.validation.Validators.notNull
+import fi.hel.haitaton.hanke.validation.Validators.notNullOrBlank
+import fi.hel.haitaton.hanke.validation.Validators.validate
 import jakarta.validation.ConstraintValidator
 import jakarta.validation.ConstraintValidatorContext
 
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/HankeValidator.kt
```diff
@@ -19,92 +27,41 @@ class HankeValidator : ConstraintValidator<ValidHanke, Hanke> {
...
+        val hankeResult = hanke.validate()
+        hankeResult.errorPaths().forEach { context.addViolation(HankeError.HAI1002, it) }
 
-        ok = ok && checkHankealueet(hanke, context)
-        ok = ok && checkTyomaaTiedot(hanke, context)
+        val alueResult = validate().andAllIn(hanke.alueet, "alueet", ::validateHankeAlue)
+        alueResult.errorPaths().forEach { context.addViolation(HankeError.HAI1032, it) }
 
-        return ok
+        return hankeResult.isOk() and alueResult.isOk()
```
pitkni, line 36: Nitpicking but doesn't 'and' always evaluate both sides? If hankeResult is not ok, is it necessary to check alueresult.ok?

pitkni (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/HankeValidator.kt
corvidian: Sorry, it doesn't actually matter here. I think I had the `and` between the validation calls in an earlier iteration. Now the errors are already added to the context and this just checks if both booleans are true.

corvidian: Changed to `&&`.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/validation/HankeValidatorTest.kt
```diff
@@ -0,0 +1,275 @@
...
+        verifyError(HankeError.HAI1032, "alueet[0].nimi")
+    }
+
+    @Test
+    fun `fails if haitta alku pvm is too far in the future`() {
+        val hanke = HankeFactory.create().withHankealue(haittaAlkuPvm = MAXIMUM_DATE.plusDays(1))
+
+        assertThat(hankeValidator.isValid(hanke, context)).isFalse()
+
+        verifyError(HankeError.HAI1032, "alueet[0].haittaAlkuPvm")
```
pitkni, line 199: Sonar is complaining about "Define a constant instead of duplicating this literal".

I'm not sure if it is possible for merge from dev to main will fail code quality if we have too many of these from different individual commits. So I guess they should be fixed?

corvidian: I resolved the issues in Sonarcloud. I think using a constant here would be very distracting.

PR #444, opened 2023-09-27
HAI-1876 Refactor LuokitteluRajaArvotService and PerusIndeksiPainotService

corvidian:
# Description

There's only one hardcoded implementation of LuokitteluRajaArvotService, so remove the interface and make the service an object. It doesn't need to be injected. Use when to get the classifications instead of searching through a list. This makes it clearer what's going on, at least for me. Mimic the old behaviour exactly, including negative input values. Add missing tests and test cases to LuokitteluRajaArvotServiceTest.

PerusIndeksiPainotService is just a holder for a single map, so remove the service and make the map a public value.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Checklist:

 or other location: 

# Other relevant info
This was a small distraction while working on the index split.


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/LuokitteluRajaArvotService.kt
```diff
@@ -1,47 +1,31 @@
...
-            rajaArvotBussi.find { it.minimumValue <= countOfRushHourBuses }?.luokittelu?.value ?: 0
-
+object LuokitteluRajaArvotService {
+
+    fun getHaittaAjanKestoLuokka(days: Int): Int =
+        when {
+            days >= 91 -> HaittaAjanKestoLuokittelu.YLI_3KK.value
+            days >= 14 -> HaittaAjanKestoLuokittelu.ALLE_3KK_YLI_2VK.value
+            days >= 0 -> HaittaAjanKestoLuokittelu.ALLE_2VK.value
+            else -> 0
```
pitkni, line 10: It seems that the only way to end up with zero is a negative number as an argument. Would it make sense to somehow validate the input in these functions (or is that done somewhere else)?

corvidian: This was how the previous implementation worked. Refactoring shouldn't change the behaviour.

The only way to get a negative input here is if the hanke end date is before the start date, which is validated to never happen.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/LuokitteluRajaArvotService.kt
```diff
@@ -1,47 +1,31 @@
...
-    override fun getHaittaAjanKestoLuokka(days: Int) =
-            rajaArvotHaittaAjanKesto.find { it.minimumValue <= days }?.luokittelu?.value ?: 0
-
-    override fun getLiikennemaaraLuokka(volume: Int) =
-            rajaArvotLiikennemaara.find { it.minimumValue <= volume }?.luokittelu?.value ?: 0
-
-    override fun getBussiLiikenneRuuhkaLuokka(countOfRushHourBuses: Int) =
-            rajaArvotBussi.find { it.minimumValue <= countOfRushHourBuses }?.luokittelu?.value ?: 0
-
+object LuokitteluRajaArvotService {
```
pitkni, line 3: Should this be renamed? Seeing this object elsewhere, the word service may imply (to me at least) that this is a spring component. Maybe just RajaArvoLuokittelu, RajaArvot?

corvidian: Renamed to `RajaArvoLuokittelija`.

pitkni (approved):

PR #445, opened 2023-09-27
HAI-1876 Separate buses and trams to separate indexes

corvidian:
# Description

In törmäystarkastelu, save bus and tram indexes separately. Remove the old joukkoliikenneindeksi from the DB.

Also bring the tormaystarkastelutulos table up to match the current entity. Changes to the entity have not been reflected in the table.

Return the new indexes separately from the API. Also return the old joukkoliikenneindeksi for compatibility, but calculate it's value on the fly from the bus and tram indexes.

Also remove `equals` and `hashCode` from `TormaystarkasteluTulosEntity`. The entities are not placed in sets or used as keys in a HashMap, so they are not needed. They were never called, which shows up as an error in test coverage reports.

Add tests for TormaystarkasteluLaskentaService. I perhaps added more tests than strictly necessary. I wanted to get full coverage and - perhaps more importantly - I wanted the tests to be helpful in understanding the behaviour of the classification functions.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluLaskentaServiceTest.kt
+++ b/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluLaskentaServiceTest.kt
@@ -1,5 +1,6 @@
 package fi.hel.haitaton.hanke.tormaystarkastelu
 
+import assertk.assertThat
 import assertk.assertions.isEqualTo
 import assertk.assertions.isNotNull
 import fi.hel.haitaton.hanke.HankeStatus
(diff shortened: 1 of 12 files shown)


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluLaskentaServiceTest.kt
```diff
@@ -50,16 +73,372 @@ internal class TormaystarkasteluLaskentaServiceTest {
         assertThat(perusindeksi).isEqualTo(indeksi)
     }
 
+    @Nested
+    inner class KatuluokkaLuokittelu {
+        // The parameter is only used to call mocks
+        val geometriat = listOf<Geometriat>()
+
+        @Nested
+        inner class WithYlreParts {
```
pitkni, line 82: I don't know what ylre means. Is that a term that a haitaton developer should know, or should that be renamed?

pitkni (approved):

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluLaskentaServiceTest.kt
corvidian: I think YLRE is Yleisten alueiden rekisteri: https://hri.fi/data/fi/dataset/helsingin-kaupungin-yleisten-alueiden-rekisteri.

I used the database table names in some places because the Finnish ones didn't offer much more context. But they were harder to distinguish from one another:
- Yleinen katuosa
- Yleinenkatualue katuluokka (Or just Yleinen katualue in one comment)
- Liikenteellinen katuluokka

These all just sound like the same thing to me.

I think I don't understand these wee enough to come up with good names.

PR #450, opened 2023-09-29
HAI Add userId to all logs

corvidian:
# Description

It has proven itself valuable to have the user's id present in all logs. Add them to both application and access logs.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Do something and check logs. Also try listing public hankkeet while logged out.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/UserIdLoggingFilter.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/UserIdLoggingFilter.kt
@@ -1,0 +1,31 @@
+package fi.hel.haitaton.hanke.security
+
+import jakarta.servlet.Filter
+import jakarta.servlet.FilterChain
+import jakarta.servlet.ServletRequest
+import jakarta.servlet.ServletResponse
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.slf4j.MDC
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.stereotype.Component
+
+@Component
+class UserIdLoggingFilter : Filter {
+    override fun doFilter(
+        request: ServletRequest,
+        response: ServletResponse?,
+        filterChain: FilterChain,
+    ) {
+        if (request is HttpServletRequest && response is HttpServletResponse) {
+            // Nullable for GDPR API with alternative authentication
+            val userId = SecurityContextHolder.getContext().authentication.name
+            // Application log can use MDC
+            MDC.put("userId", userId)
+            // Access log's AccessEventCompositeJsonEncoder can't use MDC, so use a request
+            // attribute instead
+            request.setAttribute("userId", userId)
+        }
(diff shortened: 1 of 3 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/UserIdLoggingFilter.kt
```diff
@@ -0,0 +1,31 @@
...
+import jakarta.servlet.ServletRequest
+import jakarta.servlet.ServletResponse
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.slf4j.MDC
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.stereotype.Component
+
+@Component
+class UserIdLoggingFilter : Filter {
```
pitkni, line 14: Would it make sense to have a unit test to verify that the userId is logged?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/UserIdLoggingFilter.kt
```diff
@@ -0,0 +1,31 @@
...
+    override fun doFilter(
+        request: ServletRequest,
+        response: ServletResponse?,
+        filterChain: FilterChain,
+    ) {
+        if (request is HttpServletRequest && response is HttpServletResponse) {
+            // Nullable for GDPR API with alternative authentication
+            val userId = SecurityContextHolder.getContext().authentication.name
```
pitkni, line 22: How about currentUserId?

corvidian: That's why this is there:
>// Nullable for GDPR API with alternative authentication

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/UserIdLoggingFilter.kt
corvidian: I'm not sure how to test this. This would be a controller test, but the integration tests have their own log format. I could maybe add the userId to that format as well if I redefine the pattern, but then I would be testing the test log pattern.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/UserIdLoggingFilter.kt
pitkni: Ok, I see.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/security/UserIdLoggingFilter.kt
pitkni: Yup, testing a test does not make a lot of sense. I guess there is not much we can do then.

pitkni (approved):

PR #453, opened 2023-10-03
HAI-1896 Create hanke users after allu update and send emails

pitkni:
# Description

When application is updated after initial send, create tokens for new contacts and send notification emails. A contact with an email address that was not present in the previous application is considered a new contact.

Additional: 
- Change application update operation so that application is saved, and the saved data is then sent to Allu (this feels a bit more logical order of operations at least for me).
- Add feature flags to prevent token creations and related emails on applications in production.
- Fix typo in hanke attachment service.
- Refactor ApplicationServiceTest class to use nested classes / have some structure.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. Send johtoselvitys to Allu.
2. Modify contacts after send. New email addresses have access to hanke and receive emails about hanke and application.

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
@@ -67,6 +67,7 @@ import fi.hel.haitaton.hanke.factory.AlluDataFactory.Companion.withCustomer
 import fi.hel.haitaton.hanke.factory.ApplicationHistoryFactory
 import fi.hel.haitaton.hanke.factory.AttachmentFactory
 import fi.hel.haitaton.hanke.factory.HankeFactory
+import fi.hel.haitaton.hanke.factory.HankeFactory.Companion.defaultNimi
 import fi.hel.haitaton.hanke.factory.HankeFactory.Companion.withHankealue
 import fi.hel.haitaton.hanke.findByType
 import fi.hel.haitaton.hanke.firstReceivedMessage
@@ -172,10 +173,12 @@ class ApplicationServiceITest : DatabaseTest() {
 
     fun createHankeEntity(
         hankeTunnus: String? = HANKE_TUNNUS,
+        hankeNimi: String = defaultNimi,
         generated: Boolean = false
-    ): HankeEntity {
-        return hankeRepository.save(HankeEntity(hankeTunnus = hankeTunnus, generated = generated))
-    }
+    ): HankeEntity =
+        hankeRepository.save(
+            HankeEntity(hankeTunnus = hankeTunnus, nimi = hankeNimi, generated = generated)
+        )
 
     @Nested
     inner class CreateApplication {
@@ -599,7 +602,7 @@ class ApplicationServiceITest : DatabaseTest() {
         }
 
         @Test
-        fun `when application is in Allu should initialize access for new contacts`() {
+        fun `when application is in Allu should provide access for new contacts`() {
             val cableReport =
                 createCableReportApplicationData(
                     customerWithContacts = hakijaCustomerContact,
@@ -646,6 +649,48 @@ class ApplicationServiceITest : DatabaseTest() {
             }
             verify(exactly = 3) { emailSenderService.sendHankeInvitationEmail(any()) }
             verify(exactly = 3) { emailSenderService.sendApplicationNotificationEmail(any()) }
+        }
+
+        @Test
+        fun `when no sender hanke user found should not send any emails`() {
+            val initialApplicationData =
+                createCableReportApplicationData(
+                    customerWithContacts = hakijaCustomerContact,
+                    contractorWithContacts = suorittajaCustomerContact,
+                )
+            val application =
+                alluDataFactory.saveApplicationEntity(
+                    USERNAME,
+                    hanke = createHankeEntity(hankeTunnus = HANKE_TUNNUS, generated = true),
+                    application =
+                        createApplication(
+                            alluid = 21,
+                            alluStatus = PENDING,
+                            applicationIdentifier = defaultApplicationIdentifier,
+                            applicationData = initialApplicationData
+                        )
+                )
+            justRun { cableReportServiceAllu.update(21, any()) }
+            justRun { cableReportServiceAllu.addAttachment(21, any()) }
+            every { cableReportServiceAllu.getApplicationInformation(any()) } returns
+                createAlluApplicationResponse(21)
+
+            val updatedApplication =
+                initialApplicationData.copy(
+                    representativeWithContacts = asianHoitajaCustomerContact,
+                    propertyDeveloperWithContacts = rakennuttajaCustomerContact,
+                    contractorWithContacts =
+                        suorittajaCustomerContact.changeContactEmails("new.mail@foo.fi")
+                )
+
+            applicationService.updateApplicationData(application.id!!, updatedApplication, USERNAME)
+
+            verifySequence {
+                cableReportServiceAllu.getApplicationInformation(any())
+                cableReportServiceAllu.update(21, any())
+                cableReportServiceAllu.addAttachment(any(), any())
+            }
+            verify { emailSenderService wasNot Called }
         }
 
         @Test
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
@@ -602,7 +602,6 @@ class ApplicationServiceITest : DatabaseTest() {
         fun `when application is in Allu should initialize access for new contacts`() {
             val cableReport =
                 createCableReportApplicationData(
-                    areas = listOf(aleksanterinpatsas),
                     customerWithContacts = hakijaCustomerContact,
                     contractorWithContacts = suorittajaCustomerContact,
                 )
@@ -613,14 +612,13 @@ class ApplicationServiceITest : DatabaseTest() {
                 )
             fakeSendAllu(applicationRepository.findById(application.id!!).orElseThrow())
             val capturedNotifications = mutableListOf<ApplicationNotificationData>()
-            with(cableReportServiceAllu) {
-                justRun { update(21, any()) }
-                every { getApplicationInformation(any()) } returns createAlluApplicationResponse(21)
-                justRun { addAttachment(21, any()) }
-            }
-            with(emailSenderService) {
-                justRun { sendHankeInvitationEmail(any()) }
-                justRun { sendApplicationNotificationEmail(capture(capturedNotifications)) }
+            justRun { cableReportServiceAllu.update(21, any()) }
+            every { cableReportServiceAllu.getApplicationInformation(any()) } returns
+                createAlluApplicationResponse(21)
+            justRun { cableReportServiceAllu.addAttachment(21, any()) }
+            justRun { emailSenderService.sendHankeInvitationEmail(any()) }
+            justRun {
+                emailSenderService.sendApplicationNotificationEmail(capture(capturedNotifications))
             }
 
             applicationService.updateApplicationData(
@@ -638,16 +636,12 @@ class ApplicationServiceITest : DatabaseTest() {
             assertThat(capturedNotifications)
                 .areValid(application.applicationType, application.hankeTunnus)
             verifySequence {
-                with(cableReportServiceAllu) {
-                    getApplicationInformation(any())
-                    update(21, any())
-                    addAttachment(any(), any())
-                }
+                cableReportServiceAllu.getApplicationInformation(any())
+                cableReportServiceAllu.update(21, any())
+                cableReportServiceAllu.addAttachment(any(), any())
             }
-            with(emailSenderService) {
-                verify(exactly = 3) { sendHankeInvitationEmail(any()) }
-                verify(exactly = 3) { sendApplicationNotificationEmail(any()) }
-            }
+            verify(exactly = 3) { emailSenderService.sendHankeInvitationEmail(any()) }
+            verify(exactly = 3) { emailSenderService.sendApplicationNotificationEmail(any()) }
         }
 
         @Test
@@ -1122,15 +1116,12 @@ class ApplicationServiceITest : DatabaseTest() {
                     USERNAME,
                 )
             val capturedEmails = mutableListOf<ApplicationNotificationData>()
-            with(cableReportServiceAllu) {
-                every { create(any()) } returns 26
-                every { getApplicationInformation(any()) } returns createAlluApplicationResponse(26)
-                justRun { addAttachment(26, any()) }
-            }
-            with(emailSenderService) {
-                justRun { sendHankeInvitationEmail(any()) }
-                justRun { sendApplicationNotificationEmail(capture(capturedEmails)) }
-            }
+            every { cableReportServiceAllu.create(any()) } returns 26
+            every { cableReportServiceAllu.getApplicationInformation(any()) } returns
+                createAlluApplicationResponse(26)
+            justRun { cableReportServiceAllu.addAttachment(26, any()) }
+            justRun { emailSenderService.sendHankeInvitationEmail(any()) }
+            justRun { emailSenderService.sendApplicationNotificationEmail(capture(capturedEmails)) }
 
             applicationService.sendApplication(application.id!!, USERNAME)
 
@@ -1138,16 +1129,12 @@ class ApplicationServiceITest : DatabaseTest() {
             assertThat(capturedEmails)
                 .areValid(application.applicationType, application.hankeTunnus)
             verifySequence {
-                with(cableReportServiceAllu) {
-                    create(any())
-                    addAttachment(any(), any())
-                    getApplicationInformation(any())
-                }
+                cableReportServiceAllu.create(any())
+                cableReportServiceAllu.addAttachment(any(), any())
+                cableReportServiceAllu.getApplicationInformation(any())
             }
-            with(emailSenderService) {
-                verify(exactly = 3) { sendHankeInvitationEmail(any()) }
-                verify(exactly = 3) { sendApplicationNotificationEmail(any()) }
-            }
+            verify(exactly = 3) { emailSenderService.sendHankeInvitationEmail(any()) }
+            verify(exactly = 3) { emailSenderService.sendApplicationNotificationEmail(any()) }
         }
 
         @Test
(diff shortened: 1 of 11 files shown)


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceTest.kt
```diff
@@ -65,37 +68,58 @@ import org.springframework.test.context.junit.jupiter.SpringExtension
...
+    @MockK private lateinit var emailSenderService: EmailSenderService
+    @MockK private lateinit var attachmentService: ApplicationAttachmentService
+
+    @MockK(relaxUnitFun = true) private lateinit var disclosureLogService: DisclosureLogService
+    @MockK(relaxUnitFun = true) private lateinit var loggingService: ApplicationLoggingService
+    @MockK(relaxUnitFun = true) private lateinit var hankeKayttajaService: HankeKayttajaService
+    @MockK(relaxUnitFun = true) private lateinit var hankeLoggingService: HankeLoggingService
+    @MockK(relaxed = true) private lateinit var featureFlags: FeatureFlags
+
+    @InjectMockKs private lateinit var applicationService: ApplicationService
```
corvidian, line 90: Why would you replace a concrete constructor with some kind of magick?

When ApplicationService dependencies change, this will give runtime errors for missing mocks. Or this might leave unnecessary mocks some dependencies are removed. Using a concrete constructor gives compile errors if the dependencies don't match, which is easier to spot. Also, it's clear which mocks are used and which are not.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -329,12 +344,63 @@ open class ApplicationService(
...
+            application = application,
+            hanke = hanke,
+            currentKayttaja = kayttaja,
+            currentUserId = currentUserId,
+            contacts = contacts
+        )
+    }
+
+    /** Creates access for new application contacts. Email address used for comparison. */
+    private fun initAccessOnAlluUpdate(
```
corvidian, line 371: I think access can be initialized just once. Calling this init feels odd.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -128,15 +132,15 @@ open class ApplicationService(
         userId: String,
     ): Application {
         val application = getById(id)
-        val applicationBefore = application.toApplication()
+        val previous = application.toApplication()
```
corvidian, line 135: Maybe `previousApplication`? Calling it just `previous` makes me check what this is every time I see it.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceTest.kt
pitkni: I thought annotations would have been a simpler way to initialise the service (ioc). If there is a missing mock, I don't think this will compile. But yea, this is somewhat ugly and and the previous style was maybe more readable. I'll revert.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -329,12 +344,63 @@ open class ApplicationService(
...
+    private fun initAccessOnInitialSend(
         application: ApplicationEntity,
         hanke: HankeEntity,
-        currentUserId: String
+        currentUserId: String,
+    ) {
+        if (featureFlags.isDisabled(Feature.USER_MANAGEMENT)) {
+            logger.info { "Feature ${Feature.USER_MANAGEMENT} disabled. No tokens created." }
+            return
+        }
```
corvidian, line 356: I think the feature was already checked in HankeKayttajaService and EmailSenderService.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceTest.kt
```diff
@@ -119,383 +155,392 @@ class ApplicationServiceTest {
...
-
-        verifySequence {
-            disclosureLogService wasNot called
-            applicationRepo.findOneById(3)
-            geometriatDao.isInsideHankeAlueet(any(), any())
-            geometriatDao.calculateCombinedArea(any())
-            geometriatDao.calculateArea(any())
-            cableReportService.create(any())
+        @Test
+        fun `when login error should not save disclosure logs`() {
```
corvidian, line 402: I would prefer proper sentences in the test names.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceTest.kt
```diff
@@ -119,383 +155,392 @@ class ApplicationServiceTest {
...
-                hankeEntity.nimi!!,
-                USERNAME,
-                sender
-            )
-            emailSenderService.sendApplicationNotificationEmail(any())
-            emailSenderService.sendApplicationNotificationEmail(any())
-            applicationRepo.save(any())
+        @ParameterizedTest
+        @CsvSource("true,Louhitaan", "false,Ei louhita")
+        fun `when sending should add rock excavation information to work description`(
```
corvidian, line 426: Does every test name have to start with 'when'? Even if there's no condition specified?

pitkni: The pattern I'm trying to achieve:

```
GIVEN a context -> this can usually be omitted if we have nested classes (it is the context)
WHEN some condition 
THEN expect some output
```

Test naming should be consistent. This may not be perfect but it is better than 95 % of haitaton test names. Maybe we should decide what is the way to name tests in haitaton.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -597,6 +598,58 @@ class ApplicationServiceITest : DatabaseTest() {
...
+                    customerWithContacts = hakijaCustomerContact,
+                    contractorWithContacts = suorittajaCustomerContact,
+                )
+            val application =
+                hankeService.generateHankeWithApplication(
+                    CableReportWithoutHanke(CABLE_REPORT, cableReport),
+                    USERNAME,
+                )
+            mockSendAllu(applicationRepository.findById(application.id!!).orElseThrow())
+            val capturedEmails = mutableListOf<ApplicationNotificationData>()
```
corvidian, line 615: Maybe capturedNotifications? To make it clearer these don't include the invitations.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -597,6 +598,58 @@ class ApplicationServiceITest : DatabaseTest() {
             }
         }
 
+        @Test
+        fun `when application is in Allu should initialize access for new contacts`() {
```
corvidian, line 602: Since we recently realized we're sending several emails to the same person if they're added to different roles in the same application, there should be a test that documents this behaviour. Even if we don't make any changes to the behaviour yet.

Also, does this one test cover all possible cases?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -597,6 +598,58 @@ class ApplicationServiceITest : DatabaseTest() {
...
+                    propertyDeveloperWithContacts = rakennuttajaCustomerContact,
+                    contractorWithContacts =
+                        suorittajaCustomerContact.changeContactEmails("new.mail@foo.fi")
+                ),
+                USERNAME
+            )
+
+            assertThat(capturedEmails).hasSize(3)
+            assertThat(capturedEmails)
+                .areValid(application.applicationType, application.hankeTunnus)
```
corvidian, line 639: This could use something like:
```
            assertThat(capturedEmails)
                .extracting { it.recipientEmail }
                .containsExactlyInAnyOrder(
                    "new.mail@foo.fi",
                    asianHoitajaCustomerContact.contacts[0].email,
                    rakennuttajaCustomerContact.contacts[0].email
                )
```
So that:
1. There's a check that we send the right emails. `areValid` doesn't check we send one per email and it uses a list with an extra email we don't want to send to here.
2. It makes it a lot easier to read this test when one can clearly see which emails are supposed to be sent.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -1754,8 +1819,24 @@ class ApplicationServiceITest : DatabaseTest() {
...
+        alluId: Int? = null,
+        applicationIdentifier: String? = null,
+    ): Application =
+        createApplication(
+            alluid = alluId,
+            applicationIdentifier = applicationIdentifier,
+            applicationData = applicationData
+        )
+
+    private fun mockSendAllu(applicationEntity: ApplicationEntity) {
```
corvidian, line 1831: This doesn't mock anything, at least not in the way we usually mean when talking about mocking in tests.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -597,6 +598,58 @@ class ApplicationServiceITest : DatabaseTest() {
             }
         }
 
+        @Test
+        fun `when application is in Allu should initialize access for new contacts`() {
+            val cableReport =
+                createCableReportApplicationData(
+                    areas = listOf(aleksanterinpatsas),
```
corvidian, line 605: The areas are irrelevant here. The factory-function creates an area by default.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -597,6 +598,58 @@ class ApplicationServiceITest : DatabaseTest() {
...
+            }
+
+            applicationService.updateApplicationData(
+                application.id!!,
+                cableReport.copy(
+                    representativeWithContacts = asianHoitajaCustomerContact,
+                    propertyDeveloperWithContacts = rakennuttajaCustomerContact,
+                    contractorWithContacts =
+                        suorittajaCustomerContact.changeContactEmails("new.mail@foo.fi")
+                ),
```
corvidian, line 633: I think constructing the new application data should be done in the arrange-stage. The act-stage should be kept simple.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -597,6 +598,58 @@ class ApplicationServiceITest : DatabaseTest() {
...
+
+            assertThat(capturedNotifications).hasSize(3)
+            assertThat(capturedNotifications)
+                .areValid(application.applicationType, application.hankeTunnus)
+            verifySequence {
+                with(cableReportServiceAllu) {
+                    getApplicationInformation(any())
+                    update(21, any())
+                    addAttachment(any(), any())
+                }
```
corvidian, line 645: Is it so bad to repeat `cableReportServiceAllu` three times? Using control structures increases cognitive load, i.e. makes code harder to understand.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: Can you elaborate what do you mean by all possible cases? If there is an email that did not exist on the previous application state -> send email. The subtractByEmail function could be more thoroughly tested, but maybe not in an integration test?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: Renamed to fakeSendAllu, could not figure out anything better.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: Fixed by removing areas.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: Fixed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
corvidian: How about setAlluFields?

Even better would be if the application was created with these values set, so there's no need to update them afterwards.

corvidian: Maybe something like:
```
            val initialApplicationData =
                createCableReportApplicationData(
                    customerWithContacts = hakijaCustomerContact,
                    contractorWithContacts = suorittajaCustomerContact,
                )
            val application =
                alluDataFactory.saveApplicationEntity(
                    USERNAME,
                    hanke = createHankeEntity(hankeTunnus = HANKE_TUNNUS),
                    application = mockApplicationWithArea()
                ) {
                    it.alluid = 21
                    it.alluStatus = PENDING
                    it.applicationIdentifier = defaultApplicationIdentifier
                    it.applicationData = initialApplicationData
                }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: I wanted to prevent the call on hankeKayttajaService.getKayttajaByUserId as there is no hankekayttajas currently. I guess it is not dangerous to do that but it serves no purpose either.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Renamed init to provide.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
pitkni: Renamed as suggested.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: This will not create the initial hankekayttaja (perustaja) which is required for sending emails. Using generateHankeWithApplication is not ideal but it does the job.

Renamed the problematic function to setAlluFields.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: Removed withs.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: Moved.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
pitkni: Added. 

Also removed:  assertThat(capturedEmails).hasSize(3) as redundant.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
corvidian: I'm not sure which cases should be tested. I'm just worried that there's just one test added.

Maybe the case where the current user doesn't have a HankeKayttaja?

Or different customer/contact combinations:
- New customer / old customer
- New contact under an existing customer
- Several contacts under the same customer

Is there anything outside the happy path that should be tested?

Of course, all the tests don't have to be top-level integration tests.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceTest.kt
corvidian: Yes, we should discuss how tests should be named.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
corvidian: Checking the feature at several steps within the same call adds complexity. The added benefits should outweigh that. If we want to check the features here, do we still need the other checks?

pitkni: I guess the flag check in hankekayttajaservice.saveNewTokensFromApplication is now redundant. Removed it.

corvidian (approved):

PR #454, opened 2023-10-05
HAI-1970 Add decision downloads to audit logs

corvidian:
# Description

Decisions contain personal information, so we need to log who we give what information to.

We can't read the personal information from inside the PDF decision, but we can log the metadata about the decision, which is the same as metadata about the application. This shows the exact version of the decision (in applicationIdentifier), which can then be (manually) matched against the content of the file if necessary.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Download a decision. The audit log should appear in the audit_logs table.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogService.kt
@@ -49,7 +49,7 @@ class DisclosureLogService(private val auditLogService: AuditLogService) {
     }
 
     /**
-     * Save disclosure logs for when a user downloads a decision.We don't know what information is
+     * Save disclosure logs for when a user downloads a decision. We don't know what information is
      * inside the PDF, but we can log the meta information about the decision (or application).
      *
      * Decisions contain private information, so their reads need to be logged.
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogService.kt
@@ -3,6 +3,7 @@ package fi.hel.haitaton.hanke.logging
 import fi.hel.haitaton.hanke.allu.CustomerType
 import fi.hel.haitaton.hanke.application.Application
 import fi.hel.haitaton.hanke.application.ApplicationData
+import fi.hel.haitaton.hanke.application.ApplicationMetaData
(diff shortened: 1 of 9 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogService.kt
```diff
@@ -47,6 +48,17 @@ class DisclosureLogService(private val auditLogService: AuditLogService) {
         saveDisclosureLogs(ALLU_AUDIT_LOG_USERID, UserRole.SERVICE, entries)
     }
 
+    /**
+     * Save disclosure logs for when a user downloads a decision.We don't know what information is
```
pitkni, line 52: a decision.We don't -> a decision. We don't

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogServiceTest.kt
```diff
@@ -367,6 +367,43 @@ internal class DisclosureLogServiceTest {
...
+                """
+                {
+                  "id": ${application.id},
+                  "alluid": ${application.alluid},
+                  "alluStatus": "${application.alluStatus}",
+                  "applicationIdentifier": "${application.applicationIdentifier}",
+                  "applicationType": "${application.applicationType}",
+                  "hankeTunnus": "${application.hankeTunnus}"
+                }"""
+                    .trimIndent()
```
pitkni, line 390: Feel free to ignore: would it make sense to trim / create these expected objects in a similar way? The other is in row 69. There could be a helper function (maybe an extension) to do that.

corvidian: I don't see what kind of helper function would help with this.

pitkni: `fun String.toExpected() = trimIndent().replace("\n", "").replace(" ", "")`

Maybe something like that. But the reason why I commented this is that there is more than one way to create the expected log objects in this file and I think it would be beneficial to have only one.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogService.kt
corvidian: Fixed.

pitkni (approved):

PR #456, opened 2023-10-09
HAI-1866 Limit generated hanke names

pitkni:
# Description

Generated hanke names get their name from the application name. There is no limit for application names, but hanke names are restricted to 100 characters.

If generated hanke name will be too long, limit the name to first 100 characters. Otherwise the name is untouched.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
- Create johtoselvitys with a name longer than 100 characters. The generated hanke name will be the first 100 characters.
- Create johtoselvitys with a name shorter than or equal to 100 characters. The generated hanke name will match hakemus name.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -999,6 +999,28 @@ class HankeServiceITests : DatabaseTest() {
         assertThat(hanke.perustaja?.email).isEqualTo("teppo@example.test")
     }
 
+    @Test
+    fun `generateHankeWithApplication sets hanke name according to limit and saves successfully`() {
+        val expectedName = "a".repeat(MAXIMUM_HANKE_NIMI_LENGTH)
+        val tooLongName = expectedName.plus("aaa")
```
corvidian, line 1005: If this test ever fails, it might be handy if the extra part is different from the expected part.
```suggestion
        val tooLongName = expectedName.plus("bbb")
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -844,10 +845,18 @@ open class HankeServiceImpl(
...
         )
 
-    private fun generatePerustajaFrom(cableReport: CableReportApplicationData): Perustaja {
+    private fun limitHankeName(input: String): String {
+        val limit = MAXIMUM_HANKE_NIMI_LENGTH
+        if (input.length > limit) {
+            logger.info { "Hanke name too long, limited to first $limit characters." }
+        }
+        return input.take(limit)
+    }
```
corvidian, line 857: Modifying the name is not necessary unless it's over the maximum length. Skipping it is not very important, but since the if is there already, we might as well use it for this as well.

Aliasing the constant only makes it less efficient. Without declaring the value, the function can use an expression body.

Since this function is custom-made to check the hanke name and not a generic limit for string length, the parameter could be named `name`.
```suggestion
    private fun limitHankeName(name: String): String =
        if (name.length > MAXIMUM_HANKE_NIMI_LENGTH) {
            logger.info {
                "Hanke name too long, limited to first $MAXIMUM_HANKE_NIMI_LENGTH characters."
            }
            name.take(MAXIMUM_HANKE_NIMI_LENGTH)
        } else {
            name
        }
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
corvidian: The `plus()` can be used as an operator:
```suggestion
        val tooLongName = expectedName + "bbb"
```

pitkni: Both fixed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
pitkni: Modified as suggested.

corvidian (approved):

PR #458, opened 2023-10-09
HAI-1971 Add application IDs to application contact audit logs

corvidian:
# Description

Add application IDs to application contact audit logs when reading contacts.

Also, give separate object types for reading application customers and contacts and for sending those to Allu.

This also changes the merging of log lines. We now log separately if identical contacts are read from different applications in one query. We still only log once if identical contacts are read from the same application in a single query.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Open an application and check the audit_logs table.

# Checklist:

 or other location: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8378417169/Lokitus#Huomioita-audit-lokien-sis%C3%A4ll%C3%B6st%C3%A4


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLogService.kt
```diff
@@ -117,63 +129,67 @@ class DisclosureLogService(private val auditLogService: AuditLogService) {
...
-        applications: List<ApplicationData>,
+        applications: List<Pair<Long?, ApplicationData>>,
+        objectType: ObjectType,
         status: Status = Status.SUCCESS,
         failureDescription: String? = null
-    ) =
+    ): List<AuditLogEntry> =
         applications
-            .flatMap { extractCustomers(it) }
+            .flatMap { (id, data) -> extractCustomers(id, data) }
```
pitkni, line 138: If extractCustomers is left untouched, this could possible be something like:

```
        applications
            .associate { (id, data) ->
                // Ignore country, since all customers have a default country atm.
                // Also, just a country can't be considered personal information.
                // I.e. check that the customer has other info besides the country.
                id to extractCustomers(data).filter { it.copy(country = "").hasInformation() }
            }
            .map { (id, data) ->
                disclosureLogEntry(objectType, id, data, status, failureDescription)
            }
```

Is toSet() really necessary? And the copy country thing looks suspicious. Do we have a lot of customers with nothing but country?

pitkni: Maybe even just:

```
        applications
            .associate { (id, data) -> id to extractCustomers(data) }
            .map { (id, data) ->
                disclosureLogEntry(objectType, id, data, status, failureDescription)
            }
```

corvidian: As it says in the comment, all customers have a default country.

When a new application is created, an empty customer and an empty contractor are created for it. But they have the country information. So yes, the country check is needed.

```
  "customerWithContacts": {
    "contacts": [
      {
        "email": "topias.heinonen+klingontes@gofore.com",
        "phone": "041234567",
        "orderer": true,
        "lastName": "Klingontes",
        "firstName": "Aaroni"
      }
    ],
    "customer": {
      "ovt": null,
      "name": "",
      "type": null,
      "email": "",
      "phone": "",
      "country": "FI",
      "registryKey": null,
      "invoicingOperator": null,
      "sapCustomerNumber": null
    }
  },
  "propertyConnectivity": false,
  "contractorWithContacts": {
    "contacts": [
      {
        "email": "",
        "phone": "",
        "orderer": false,
        "lastName": "",
        "firstName": ""
      }
    ],
    "customer": {
      "ovt": null,
      "name": "",
      "type": null,
      "email": "",
      "phone": "",
      "country": "FI",
      "registryKey": null,
      "invoicingOperator": null,
      "sapCustomerNumber": null
    }
  },
```

pitkni:
```
  "contractorWithContacts": {
    "contacts": [
      {
        "email": "",
        "phone": "",
        "orderer": false,
        "lastName": "",
        "firstName": ""
      }
    ],
    "customer": {
      "ovt": null,
      "name": "",
      "type": null,
      "email": "",
      "phone": "",
      "country": "FI",
      "registryKey": null,
      "invoicingOperator": null,
      "sapCustomerNumber": null
    }
  },
```

Having stuff like this does not make a lot of sense to me. But it is not a problem of this pr.

corvidian: Without the `.toSet()` we'll get duplicates when the same contact is used twice.

pitkni: Looks good to me. 

```
customer.copy(country = "").hasInformation()
```

Would it be ok to either remove country in the hasInformation check? This is a small issue but then there would be no need to copy these objects.

pitkni: And then maybe update the javadoc of isBlank.

corvidian: I did the changes, except I moved the emptiness-check inside `extractCustomers()`/`extractContacts()`. It's simpler there and it fits the functions well.

This is the only place that used it, so I renamed Customer's `hasInformation()` to `hasPersonalInformation()`. I removed country from the check and added explanation of why it's not included in the KDoc﻿.

pitkni (approved):

PR #462, opened 2023-10-12
HAI-1893 Make nimi always mandatory in Hanke

corvidian:
# Description

Name is one of the few things we require every hanke to have from the first draft. There needs to be a name for the hanke to distinguish it from other hanke.

Since every hanke must have a name, we can make the associated field non-nullable. This simplifies code referring to the name.

Also, make the `check` Gradle task depend on the `integrationTest` task. This way, we can run `./gradlew check` instead of `./gradlew build integrationTest`.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
@@ -95,7 +95,7 @@ class HankeKayttajaService(
 
         val hankeId = hanke.id ?: throw HankeArgumentException("Hanke without id")
         val hankeTunnus = hanke.hankeTunnus ?: throw HankeArgumentException("Hanke without tunnus")
-        val hankeNimi = hanke.nimi ?: throw HankeArgumentException("Hanke without name")
+        val hankeNimi = hanke.nimi
 
         val contacts =
             hanke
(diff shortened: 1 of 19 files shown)


services/hanke-service/src/main/resources/db/changelog/changesets/050-make-nimi-mandatory-in-hanke.yml
```diff
@@ -0,0 +1,11 @@
...
+  - changeSet:
+      id: 050-make-nimi-mandatory-in-hanke
+      author: Topias Heinonen
+      changes:
+        - addNotNullConstraint:
+            tableName: hanke
+            columnName: nimi
+            # There shouldn't be unnamed hanke, but let's add a default name
+            # just in case, instead of failing to start.
+            defaultNullValue: "Unnamed"
```
pitkni, line 11: Unnamed would show up in the ui also. As stated, there should not be unnamed instances, but could this be something that works in Finnish, Swedish and English? Would an empty string work if there is no other suitable name?

corvidian: I checked every environment and there are no hanke with a null name in any of them. I removed the default value.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaService.kt
```diff
@@ -95,7 +95,7 @@ class HankeKayttajaService(
 
         val hankeId = hanke.id ?: throw HankeArgumentException("Hanke without id")
         val hankeTunnus = hanke.hankeTunnus ?: throw HankeArgumentException("Hanke without tunnus")
-        val hankeNimi = hanke.nimi ?: throw HankeArgumentException("Hanke without name")
+        val hankeNimi = hanke.nimi
```
pitkni, line 98: This line can possibly be removed? hankeNimi was there just to make sure nimi is not null.

corvidian: Thank you for noticing. I inlined the variable.

pitkni (approved):

PR #463, opened 2023-10-12
HAI-1893 Add a new DTO class for hanke creation

corvidian:
# Description

Add a new DTO class for creating hanke. It only has the fields that the user can decide directly. I.e. the DTO doesn't have any fields that are generated or calculated from other fields. This means there's no need to sanitize its contents or select which fields should be read. It also means the OpenAPI documentation is a lot more relevant.

Using the same request for the creation of the entity enables us to make many fields in the ubiquitous Hanke class non-nullable. This is not done in this PR, however.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
1. Create a hanke.
2. It should work.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
@@ -1,18 +1,16 @@
 package fi.hel.haitaton.hanke
 
-import fi.hel.haitaton.hanke.ContactType.MUU
-import fi.hel.haitaton.hanke.ContactType.OMISTAJA
-import fi.hel.haitaton.hanke.ContactType.RAKENNUTTAJA
-import fi.hel.haitaton.hanke.ContactType.TOTEUTTAJA
 import fi.hel.haitaton.hanke.application.Application
 import fi.hel.haitaton.hanke.application.ApplicationService
 import fi.hel.haitaton.hanke.application.CableReportApplicationData
 import fi.hel.haitaton.hanke.application.CableReportWithoutHanke
 import fi.hel.haitaton.hanke.application.Contact
+import fi.hel.haitaton.hanke.domain.CreateHankeRequest
 import fi.hel.haitaton.hanke.domain.Hanke
 import fi.hel.haitaton.hanke.domain.HankeYhteystieto
 import fi.hel.haitaton.hanke.domain.Hankealue
 import fi.hel.haitaton.hanke.domain.HasId
+import fi.hel.haitaton.hanke.domain.HasYhteystiedot
 import fi.hel.haitaton.hanke.domain.Perustaja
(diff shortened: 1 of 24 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -150,6 +131,40 @@ open class HankeServiceImpl(
         }
     }
 
+    private fun createEntityFromCreateRequest(
```
pitkni, line 134: Just an idea. Could we have this as a non private function somewhere? It would simplify testing (hanke creation based on request) and provide a single source of truth on how an entity is created from a request. Currently this part is not needed anywhere else, but if it was, this would be valuable.

Or is there a reason that this has to be private?

corvidian: This method doesn't build the complete entity, since things like hanketunnus, generated, perustaja, yhteystiedot and tormaystarkastelu are handled in the parent function. Because of the limited functionality, I think it would be misleading to make this public. The visibility can be adjusted later, of course, if the need arises. But I don't expect it will.

Also, because of the call to hanketunnusService, this can be only run in integration tests without extra mocking. In integration tests, creating an entity that's not backed in the DB would be weird. So it's hard to see the value this method would have in tests.

pitkni: Fair enough.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/CreateHankeRequest.kt
```diff
@@ -0,0 +1,63 @@
...
+    override val vaihe: Vaihe? = null,
+    @field:Schema(
+        description = "Current planning stage, must be defined if vaihe = SUUNNITTELU",
+    )
+    override val suunnitteluVaihe: SuunnitteluVaihe? = null,
+    @field:Schema(
+        description =
+            "Project owners, contact information. At least one is required for the hanke to be published.",
+    )
+    override val omistajat: List<HankeYhteystieto>? = null,
```
pitkni, line 35: Would these require a request object too? It contains id, timestamps, etc.

corvidian: The same is true about alueet. And for those, we might also need a new Geometria class.

pitkni: I think HAI-1893 can cover these. But subtasking / different pr:s sound good.

pitkni (approved):

PR #466, opened 2023-10-13
HAI-1893 Make hanketunnus mandatory

corvidian:
# Description

Make hanketunnus mandatory, i.e. non-nullable, in both Hanke and HankeEntity. Now that there's a separate class for hanke creation, hanketunnus is always defined when these classes come into play.

Also, rewrite controller tests for hanke creation. Change them to use the HAnkeCreateRequest instead of just Hanke, but add one test that documents what happens when the endpoint is called with a full Hanke object. Some tests that didn't really test any functionality were removed.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
@@ -35,9 +35,9 @@ data class ApplicationEntity(
             applicationIdentifier,
             applicationType,
             applicationData,
-            hanke.hankeTunnus!!,
+            hanke.hankeTunnus,
         )
 
     /** An application must belong to a Hanke. Thus, hankeTunnus must be present. */
-    fun hankeTunnus(): String = hanke.hankeTunnus!!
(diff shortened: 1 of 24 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationEntity.kt
```diff
@@ -35,9 +35,9 @@ data class ApplicationEntity(
             applicationIdentifier,
             applicationType,
             applicationData,
-            hanke.hankeTunnus!!,
+            hanke.hankeTunnus,
         )
 
     /** An application must belong to a Hanke. Thus, hankeTunnus must be present. */
-    fun hankeTunnus(): String = hanke.hankeTunnus!!
+    fun hankeTunnus(): String = hanke.hankeTunnus
```
pitkni, line 42: I'm not sure if this is actually _needed_ anymore, but I guess it may be a convenient way to get the tunnus. Feel free to ignore.

pitkni (approved):

PR #467, opened 2023-10-13
HAI-1893 Make id mandatory in Hanke and HankeEntity

corvidian:
# Description

Make id mandatory (non-nullable) in both Hanke and HankeEntity. We give the id a default value in the entity, but it's rewritten during the initial persist. But it still needs to be a var, and not a val.

Also, remove some asserts that can never fail from HankeServiceITest.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
@@ -159,12 +159,11 @@ open class ApplicationService(
         }
 
         val hanke = application.hanke
-        val hankeId = hanke.id
         if (!hanke.generated) {
             newApplicationData.areas?.let { areas ->
-                checkApplicationAreasInsideHankealue(hankeId, areas) { applicationArea ->
+                checkApplicationAreasInsideHankealue(hanke.id, areas) { applicationArea ->
                     "Application geometry doesn't match any hankealue when updating application for user $userId, " +
-                        "hankeId = $hankeId, applicationId = ${application.id}, " +
+                        "hankeId = ${hanke.id}, applicationId = ${application.id}, " +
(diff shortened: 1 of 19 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -159,7 +159,7 @@ open class ApplicationService(
         }
 
         val hanke = application.hanke
-        val hankeId = hanke.id!!
+        val hankeId = hanke.id
```
pitkni, line 162: Can possibly be removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
```diff
@@ -197,7 +197,7 @@ open class ApplicationService(
         val application = getById(id)
 
         val hanke = application.hanke
-        val hankeId = hanke.id!!
+        val hankeId = hanke.id
```
pitkni, line 200: This can also possibly be removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
corvidian: Thank you for noticing. I inlined the value.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationService.kt
corvidian: Thank you for noticing. I inlined the value.

pitkni (approved):

PR #472, opened 2023-10-19
HAI-1893 Add DTO for creating Hankealue

corvidian:
# Description

Add a separate hankealue DTO for creating hanke.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Try to create a hanke with an alue.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -132,7 +132,7 @@ data class Hanke(
         description =
             "Hanke areas data. At least one alue is required for the hanke to be published."
     )
-    override var alueet = mutableListOf<Hankealue>()
+    override var alueet = mutableListOf<SavedHankealue>()
```
pitkni, line 135: I'm not sure what would be a better name for this, but this is also used for updating a hanke. If there is a new alue, this would be a bit misleading name? 

How about the same approach as in hanke:
- interface BaseHankeAlue
- HankeAlue for the domain object
- NewHankeAlue for request dto

corvidian: Naming interfaces or abstract classes as BaseXXX is a bad practice. I'm already regretting using it with Hanke.

The SavedHankealue name is misleading for updating a hanke because the class itself is not well suited for that purpose. There is information the user isn't able to update in the class, especially in the Geometriat attached to it.

pitkni (approved):

PR #473, opened 2023-10-19
HAI-1893 Add DTO for geometria in new hanke

corvidian:
# Description

Add a separate geometria-DTO for use in hanke creation

Also:
- refactor Tormaystarkastelu to use a set of geometria IDs, since it was all it needed from the geometries.
- Refactor GeometriaServiceImpl to maybe be a bit more understandable.
- Stop using the geometry ID given by the user when updating hanke

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Create a hanke with an alue. Can't be tested with the UI.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/geometria/GeometriatServiceImplITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/geometria/GeometriatServiceImplITest.kt
@@ -15,8 +15,8 @@ import assertk.assertions.prop
 import fi.hel.haitaton.hanke.DATABASE_TIMESTAMP_FORMAT
 import fi.hel.haitaton.hanke.DatabaseTest
 import fi.hel.haitaton.hanke.HankeService
-import fi.hel.haitaton.hanke.asJsonResource
 import fi.hel.haitaton.hanke.domain.geometriat
+import fi.hel.haitaton.hanke.factory.GeometriaFactory
 import fi.hel.haitaton.hanke.factory.HankeFactory
 import fi.hel.haitaton.hanke.factory.HankealueFactory
 import fi.hel.haitaton.hanke.test.Asserts.isRecentZDT
@@ -48,8 +48,7 @@ internal class GeometriatServiceImplITest : DatabaseTest() {
 
     @Test
     fun `save and load and update`() {
-        val geometriat: Geometriat =
-            "/fi/hel/haitaton/hanke/geometria/hankeGeometriat.json".asJsonResource()
+        val geometriat: Geometriat = GeometriaFactory.create()
         val username = SecurityContextHolder.getContext().authentication.name
 
         // For FK constraints we need a Hanke in database
@@ -131,11 +130,7 @@ internal class GeometriatServiceImplITest : DatabaseTest() {
 
     @Test
     fun `save Geometria with missing properties`() {
-        val geometriat: Geometriat =
-            "/fi/hel/haitaton/hanke/geometria/hankeGeometriat.json".asJsonResource()
-        geometriat.version = null
-        geometriat.createdAt = null
-        geometriat.modifiedAt = null
+        val geometriat = GeometriaFactory.createNew()
         geometriat.featureCollection?.crs?.properties = null
 
         assertThrows<GeometriaValidationException> {
@@ -145,11 +140,7 @@ internal class GeometriatServiceImplITest : DatabaseTest() {
 
     @Test
     fun `save Geometria with invalid coordinate system`() {
-        val geometriat: Geometriat =
-            "/fi/hel/haitaton/hanke/geometria/hankeGeometriat.json".asJsonResource()
-        geometriat.version = null
-        geometriat.createdAt = null
-        geometriat.modifiedAt = null
+        val geometriat = GeometriaFactory.createNew()
         geometriat.featureCollection?.crs?.properties?.set("name", "urn:ogc:def:crs:EPSG::0000")
 
         assertThrows<UnsupportedCoordinateSystemException> {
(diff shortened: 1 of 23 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/geometria/GeometriatServiceImplITest.kt
```diff
@@ -41,10 +48,8 @@ internal class GeometriatServiceImplITest : DatabaseTest() {
 
     @Test
     fun `save and load and update`() {
-        val geometriat =
-            "/fi/hel/haitaton/hanke/geometria/hankeGeometriat.json".asJsonResource(
-                Geometriat::class.java
-            )
+        val geometriat: Geometriat =
+            "/fi/hel/haitaton/hanke/geometria/hankeGeometriat.json".asJsonResource()
```
pitkni, line 52: Could these be a separate function somewhere? Either a private function for this class or somewhere else where it could be accessed by other test classes.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
```diff
@@ -7,6 +7,7 @@ import fi.hel.haitaton.hanke.ContactType.TOTEUTTAJA
 import fi.hel.haitaton.hanke.domain.Hanke
 import fi.hel.haitaton.hanke.domain.HankeYhteystieto
 import fi.hel.haitaton.hanke.domain.SavedHankealue
+import fi.hel.haitaton.hanke.domain.resetFeatureProperties
```
pitkni, line 10: Actually, is this reset necessary here? If this is done when creating and updating hanke, there should be no need to reset them when transforming entity to domain.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/GeometriatServiceImpl.kt
```diff
@@ -2,68 +2,72 @@ package fi.hel.haitaton.hanke.geometria
...
+        hankeGeometriaDao.deleteGeometriat(geometriat)
+        logger.info { "Deleted geometries ${geometriat.id}" }
+        return null
+    }
+
+    private fun updateGeometriat(
+        geometriat: Geometriat,
+        newFeatures: FeatureCollection
+    ): Geometriat {
+        if (geometriat.version == null) {
```
pitkni, line 58: Suggestion. No need for else branch, and removing it makes the if foldable to an elvis operator.

```
    private fun updateGeometriat(
        geometriat: Geometriat,
        newFeatures: FeatureCollection
    ): Geometriat {
        val version =
            geometriat.version
                ?: error("There is an old Geometriat ${geometriat.id} but it has no 'version'")
        geometriat.version = version + 1
        geometriat.modifiedByUserId = currentUserId()
        geometriat.modifiedAt = ZonedDateTime.now(TZ_UTC)
        geometriat.featureCollection = newFeatures
        hankeGeometriaDao.updateGeometriat(geometriat)
        logger.info { "Updated geometries ${geometriat.id}" }
        return geometriat
    }
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/geometria/GeometriatServiceImplITest.kt
corvidian: I added a factory class for creating the geometries.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
corvidian: This line is no longer needed.

I think an equally valid question is: Do we need to save the features if we can just as well reset them on the way out?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/GeometriatServiceImpl.kt
```diff
@@ -2,68 +2,72 @@ package fi.hel.haitaton.hanke.geometria
...
-                    oldGeometriat.version = oldGeometriat.version!! + 1
-                }
-                oldGeometriat.modifiedByUserId = username
-                oldGeometriat.modifiedAt = now
-                oldGeometriat.featureCollection = geometriat.featureCollection
-                hankeGeometriaDao.updateGeometriat(oldGeometriat)
-                logger.info { "Updated geometries ${oldGeometriat.id}" }
-                oldGeometriat
-            }
+            oldGeometriat != null && !hasFeatures -> deleteGeometriat(oldGeometriat)
```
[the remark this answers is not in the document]
corvidian: Renamed as suggested. I also moved the `hasFeatures()` and `resetFeatureProperties()` functions inside the HasFeatures interface, since they were declared right beside it anyway.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/geometria/GeometriatServiceImpl.kt
corvidian: I made the version non-nullable so that we can avoid the complication altogether.

I'm not sure why adding a new hankealue to an existing hanke still works, but it does.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeMapper.kt
pitkni: Hmm, I guess it does not make sense to save the properties at all.

pitkni (approved):

PR #476, opened 2023-10-25
HAI-2041 Whoami associated by hanke tunnus

pitkni:
# Description

Until now, front end has to fetch whoami data separately for each hanke. This creates unnecessary http traffic.

This commit adds a new endpoint from which a user can get whoami info for every hanke the user has a permission to in a single request.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Front end is not implemented, so:

1. Create johtoselvitys (it autogenerates a hanke)
2. Create hanke via "luo uusi hanke"
3. GET http://localhost:3001/api/hankkeet/my-permissions
4. Verify response, autogenerated hanke should have an existing hankekayttaja. For normally created hanke, it should be null. Both privileges should be at level KAIKKI_OIKEUDET.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -53,14 +53,44 @@ class HankeKayttajaController(
...
+    fun whoAmIByHanke(): Map<String, WhoAmIDto> {
+        val whoAmIAtHanke: Map<Int, WhoAmIDto> =
+            hankeKayttajaService.whoAmIByHankeId(userId = currentUserId())
+        val hankeIds = whoAmIAtHanke.keys
+
+        if (hankeIds.isEmpty()) return emptyMap()
+
+        return hankeService
+            .findIdentifiers(hankeIds)
+            .mapNotNull { (hankeId, hankeTunnus) ->
```
corvidian, line 90: I think a lot of the complexities in combining the information from different tables could be avoided by using a custom query to get the information all at once. Maybe something like (in PermissionRepository):
```kotlin
    @Query(
        "select new fi.hel.haitaton.hanke.permissions.HankePermission(" +
            "  h.hankeTunnus, hk.id, pe.kayttooikeustasoEntity" +
            ") from PermissionEntity pe " +
            "inner join HankeEntity h on pe.hankeId = h.id " +
            "left join HankeKayttajaEntity hk on hk.permission.id = pe.id " +
            "where pe.userId = :userId"
    )
    fun userPermissions(userId: String): List<HankePermission>
```

corvidian: HankePermission would also be the domain object needed to avoid using the DTO on the service level.

Maybe the kayttooikeustasoEntity could be unpacked in the query as well:
```
    @Query(
        "select new fi.hel.haitaton.hanke.permissions.HankePermission(" +
            "  h.hankeTunnus," +
            "  hk.id," +
            "  pe.kayttooikeustasoEntity.kayttooikeustaso," +
            "  pe.kayttooikeustasoEntity.permissionCode" +
            ") from PermissionEntity pe " +
            "inner join HankeEntity h on pe.hankeId = h.id " +
            "left join HankeKayttajaEntity hk on hk.permission.id = pe.id " +
            "where pe.userId = :userId"
    )
    fun userPermissions(userId: String): List<HankePermission>
```

pitkni: I agree there is some complexity to this. However, I generally consider a good practice to keep repositories focused on their respective entities. Additionally, that query is not a simple one either. I'll try to think of a way to solve this.

corvidian: > However, I generally consider a good practice to keep repositories focused on their respective entities.
Wouldn't this mean that all queries returning custom objects are forbidden? This is a big limitation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -53,14 +53,44 @@ class HankeKayttajaController(
...
+        description = "Returns a map of current users Hanke identifiers and respective permissions."
+    )
+    @ApiResponses(
+        value =
+            [
+                ApiResponse(
+                    description = "Permissions grouped by hankeTunnus.",
+                    responseCode = "200",
+                ),
+                ApiResponse(description = "Access denied.", responseCode = "401")
```
corvidian, line 78: This response (along with 500 internal server error) is automatically documented on every endpoint. It doesn't need to be repeated here.

pitkni: Ok. Removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
pitkni: No, it is just good practice. 

I changed the implementation as suggested as I did not see any other way around this. Added an index for permissions.userid also.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
```diff
@@ -157,6 +160,70 @@ class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : C
...
+
+        private val tunnus2 = "HAI23-2"
+
+        private val hankePermissions =
+            listOf(
+                HankePermission(
+                    hankeTunnus = tunnus1,
+                    hankeKayttajaId = kayttaja1,
+                    kayttooikeustaso = Kayttooikeustaso.KAIKKI_OIKEUDET,
+                    permissionCode = 1152921504606846975,
```
corvidian, line 178: To avoid the magic value:
```suggestion
                    permissionCode = PermissionCode.entries.sumOf { it.code },
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
```diff
@@ -157,6 +160,70 @@ class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : C
...
+                ),
+                HankePermission(
+                    hankeTunnus = tunnus2,
+                    hankeKayttajaId = null,
+                    kayttooikeustaso = Kayttooikeustaso.KATSELUOIKEUS,
+                    permissionCode = 1,
```
corvidian, line 184:
```suggestion
                    permissionCode = VIEW.code,
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
corvidian: Well spotted on the index 👍

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
```diff
@@ -157,6 +159,70 @@ class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : C
         }
     }
 
+    @Nested
+    inner class WhoAmIByHanke {
+        private val url = "/hankkeet/my-permissions"
+
+        private val tunnus1 = "HAI23-1"
+        private val kayttaja1 = UUID.randomUUID()
```
corvidian, line 167: This could be a pre-randomized value.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/PermissionServiceITest.kt
```diff
@@ -50,68 +54,115 @@ class PermissionServiceITest : DatabaseTest() {
...
-        val katseluoikeus = Kayttooikeustaso.KATSELUOIKEUS
-        val hankkeet = hankeFactory.saveSeveralMinimal(4)
-        listOf(kaikkiOikeudet, hankemuokkaus, hakemusasiointi, katseluoikeus).zip(hankkeet) {
-            kayttooikeustaso,
-            hanke ->
-            permissionService.create(hanke.id, username, kayttooikeustaso)
+    @Nested
+    inner class PermissionsByHanke {
+        private val canView = Kayttooikeustaso.KATSELUOIKEUS
+        private val allRights = Kayttooikeustaso.KAIKKI_OIKEUDET
```
corvidian, line 106: I don't really see the need for these aliases?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/Permissions.kt
```diff
@@ -11,10 +11,20 @@ import jakarta.persistence.Id
 import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
 import jakarta.persistence.Table
+import java.util.*
```
corvidian, line 14: Some kind of automatic import from IDE?

pitkni: It seems that this keeps happening with UUID imports no matter what I do. Fixed, thx.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/PermissionServiceITest.kt
pitkni: I had them initially to have the asserts on a single line per assert. Nicer for my eyes. But now it looks ok without them. Removed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
pitkni: Fixed.

corvidian (approved):

PR #481, opened 2023-10-31
HAI-2038 Add a hankealue for every application area in generated hanke

corvidian:
# Description

Create a new hankealue for each application area in the cable report application. The geometry of the hankealueet is the same as the application area. The start and end times of every hankealue are the same as the application dates.

Also, fix some compiler warnings left over from PR #473.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
I tested this by:
- I authenticated to the Swagger UI
- In the UI, I opened a cable report application I had made earlier.
- I copied the response to the application GET request from the network tab.
- I pasted the application JSON to http://localhost:3001/api/swagger-ui/index.html#/application-controller/createWithGeneratedHanke.
- I opened the newly created hanke to see that it had hankealue, and also that the hanke had start and end dates.

# Checklist:

 or other location: 

# Other relevant info
This doesn't work from the UI yet, since the UI adds the areas on a separate page to where the application is created. Support for that is coming in HAI-2039.

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
@@ -894,61 +894,82 @@ class HankeServiceITests : DatabaseTest() {
         assertThat(alue.geometriat).isNotNull
     }
 
-    @Test
-    fun `generateHankeWithApplication generates hanke based on application`() {
-        val inputApplication = AlluDataFactory.cableReportWithoutHanke()
+    @Nested
+    inner class GenerateHankeWithApplication {
 
-        val application = hankeService.generateHankeWithApplication(inputApplication, USER_NAME)
+        @Test
+        fun `generates hanke based on application`() {
+            val inputApplication = AlluDataFactory.cableReportWithoutHanke()
 
-        assertThat(application.applicationData.name)
-            .isEqualTo(inputApplication.applicationData.name)
-        val hanke = hankeRepository.findByHankeTunnus(application.hankeTunnus)!!
-        assertThat(hanke.generated).isTrue
-        assertThat(hanke.status).isEqualTo(HankeStatus.DRAFT)
-        assertThat(hanke.hankeTunnus).isEqualTo(application.hankeTunnus)
-        assertThat(hanke.nimi).isEqualTo(application.applicationData.name)
-        assertThat(hanke.perustaja?.nimi).isEqualTo("Teppo Testihenkilö")
-        assertThat(hanke.perustaja?.email).isEqualTo("teppo@example.test")
-    }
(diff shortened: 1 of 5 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
```diff
@@ -860,12 +863,29 @@ open class HankeServiceImpl(
      * - Hanke name is same as application name (limited to first 100 characters).
      * - Perustaja generated from application data orderer.
      */
-    private fun generateHankeFrom(cableReport: CableReportWithoutHanke): Hanke =
-        createHankeInternal(
-            CreateHankeRequest(nimi = limitHankeName(cableReport.applicationData.name)),
+    private fun generateHankeFrom(cableReport: CableReportWithoutHanke): Hanke {
```
pitkni, line 866: The comment could mention the stuff about areas. Alternatively, the comment can be removed and let the code explain itself.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -894,61 +894,82 @@ class HankeServiceITests : DatabaseTest() {
...
-        assertThat(hanke.generated).isTrue
-        assertThat(hanke.status).isEqualTo(HankeStatus.DRAFT)
-        assertThat(hanke.hankeTunnus).isEqualTo(application.hankeTunnus)
-        assertThat(hanke.perustaja?.nimi).isEqualTo("Teppo Testihenkilö")
-        assertThat(hanke.perustaja?.email).isEqualTo("teppo@example.test")
-    }
+        @Test
+        fun `creates hankealueet from the application areas`() {
+            val inputApplication = AlluDataFactory.cableReportWithoutHanke()
+            assertThat(inputApplication.applicationData.areas).isNotNull.isNotEmpty
```
pitkni, line 942: Would this work without the null check? I mean isNotEmpty should should fail for nulls.

Same thing for features size check.

corvidian: I was worried they wouldn't give a good error message in the null case. But they have an explicit null assert buried inside, so my worries were unfounded. Looks like I'm already starting to forget how AssertJ works. I dropped the null asserts.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeServiceImpl.kt
corvidian: I added a line about the areas.

pitkni (approved):

PR #483, opened 2023-10-31
HAI-2039 Seperate hankealue stuff from HankeService to a new service

corvidian:
# Description

Separate hankealue stuff from HankeService to a new service

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Refactoring, nothing should change.

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
@@ -42,20 +42,20 @@ class HankealueService(
         // Assuming the incoming date, while being zoned date and time, is in UTC and time value can
         // be simply dropped here.
         // Note, .toLocalDate() does not do any time zone conversion.
-        source.haittaLoppuPvm?.let { result.haittaLoppuPvm = source.haittaLoppuPvm?.toLocalDate() }
-        source.haittaAlkuPvm?.let { result.haittaAlkuPvm = source.haittaAlkuPvm?.toLocalDate() }
+        source.haittaLoppuPvm?.let { result.haittaLoppuPvm = it.toLocalDate() }
+        source.haittaAlkuPvm?.let { result.haittaAlkuPvm = it.toLocalDate() }
 
-        source.kaistaHaitta?.let { result.kaistaHaitta = source.kaistaHaitta }
-        source.kaistaPituusHaitta?.let { result.kaistaPituusHaitta = source.kaistaPituusHaitta }
-        source.meluHaitta?.let { result.meluHaitta = source.meluHaitta }
-        source.polyHaitta?.let { result.polyHaitta = source.polyHaitta }
-        source.tarinaHaitta?.let { result.tarinaHaitta = source.tarinaHaitta }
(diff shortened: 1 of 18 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
```diff
@@ -0,0 +1,157 @@
...
+ * }
+ * - Transforms data from hanke.alueet into entity.listOfHankeAlueet
+ * - Does the transformation with the last lambda
+ *
+ * Source list is not modified. Target container is sorted by order of source container.
+ *
+ * converter takes additional parameter for existing Target object in case it exists in target collection.
+ *
+ */
+private fun <Source : HasId<Int>, Target : HasId<Int>> mergeDataInto(
```
umeetiusbaar, line 32: If HankealueService is the only class that needs `mergeDataInto` then could it be a member function? Otherwise, it might be better to put it into `Utils.kt` to keep it as a general utility method?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
```diff
@@ -0,0 +1,157 @@
...
+        hankeTunnus: String,
+        source: Hankealue,
+        target: HankealueEntity?
+    ): HankealueEntity {
+        val result = target ?: HankealueEntity()
+
+        // Assuming the incoming date, while being zoned date and time, is in UTC and time value can
+        // be simply dropped here.
+        // Note, .toLocalDate() does not do any time zone conversion.
+        source.haittaLoppuPvm?.let { result.haittaLoppuPvm = source.haittaLoppuPvm?.toLocalDate() }
```
pitkni, line 76: Would it make sense to use it for all of these? 

`source.haittaLoppuPvm?.let { result.haittaLoppuPvm = it.toLocalDate() }`

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
corvidian: I thought I would try to make this more useful and less weird, but I don't know if I succeeded.

Looking at this led me to some refactoring in commit 49b3486:
- I made the type parameter in HasId exact, so the implementation
  specifies if the id property is nullable or not.
- I changed the id property in HankealueEntity to be non-nullable.
- I changed the mergeDataInto function to accept only target types with
  non-nullable IDs. This removes the need to handle cases where the
  target ID would be null. Move the function to Utils.kt and add tests
  for it.
- Upgrade ktfmt used by the Spotless formatter. The old version crashed
  when using Kotlin's definitely non-nullable types feature (X & Any).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
corvidian: I changed them.

umeetiusbaar (approved):
Looks good to me. Renaming of `listOfHankeAlueet` to `alueet` is an excellent choice!

PR #488, opened 2023-11-03
HAI-2039 Recreate hankealue in generated hanke on update

corvidian:
# Description

When a cable report application is updated, recreate the hankealueet of the associated hanke, if it's generated.

Also, give every generated hankealue a name. The names follow the same pattern as UI: "Hankealue 1", "Hankealue 2" and so forth. Because the hankealueet are recreated at every update, the indexing always starts at one.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
- Start to create a cable report application through the UI.
- After filling in areas on the area page save and suspend. 
- Go to the hanke that was generated for the application
- Check that the hanke has the same areas you gave the application.
- Edit the application areas. Save and suspend again.
- Check that the generated hanke has the same changes in its areas.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
@@ -3,6 +3,7 @@ package fi.hel.haitaton.hanke
 import assertk.all
 import assertk.assertFailure
 import assertk.assertions.each
+import assertk.assertions.first
 import assertk.assertions.hasClass
 import assertk.assertions.hasSize
 import assertk.assertions.isEmpty
@@ -48,6 +49,7 @@ import fi.hel.haitaton.hanke.permissions.HankeKayttajaRepository
 import fi.hel.haitaton.hanke.permissions.KayttajaTunnisteRepository
 import fi.hel.haitaton.hanke.permissions.PermissionCode
 import fi.hel.haitaton.hanke.permissions.PermissionService
+import fi.hel.haitaton.hanke.test.Asserts.hasSingleGeometryWithCoordinates
 import fi.hel.haitaton.hanke.test.Asserts.isRecent
 import fi.hel.haitaton.hanke.test.TestUtils
 import fi.hel.haitaton.hanke.test.TestUtils.nextYear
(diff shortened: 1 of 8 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -938,18 +940,19 @@ class HankeServiceITests : DatabaseTest() {
...
 
             val hanke = hankeService.loadHanke(application.hankeTunnus)!!
-            hanke.alueet.forEach { hankealue ->
-                val features = hankealue.geometriat?.featureCollection?.features
-                assertThat(features).hasSize(1)
-                val polygon = features!![0].geometry as Polygon
-                assertThat(polygon.coordinates)
-                    .isEqualTo(inputApplication.applicationData.areas!![0].geometry.coordinates)
+            assertThat(hanke.alueet).hasSize(1)
+            assertk.assertThat(hanke.alueet).first().all {
```
umeetiusbaar, line 953: Why `assertk.assertThat` instead of just `assertThat`?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
```diff
@@ -13,6 +13,8 @@ import org.geojson.Feature
 import org.geojson.FeatureCollection
 import org.springframework.stereotype.Service
 
+const val HANKEALUE_DEFAULT_NAME = "Hankealue"
```
umeetiusbaar, line 16: Should `HANKEALUE_DEFAULT_NAME` be in Constants.kt?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
corvidian: In this file `assertThat` refers to AssertJ, and I couldn't find a clean way to do this in AssertJ.

umeetiusbaar: Good point. It might be possible to get rid of AssertJ in this test class but maybe no need to do that now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
corvidian: It isn't used anywhere else, so I moved it to a private value in the companion object.

umeetiusbaar (approved):

PR #492, opened 2023-11-07
HAI-1967 Add a client for Azure Blob

corvidian:
# Description

Add a client for Azure Blob. The client can upload, download and delete files.

In upload, Content-Type and Content-Disposition headers are set for the files. These might come useful later when we want to download directly from Blob storage.

Download returns a Azure's BinaryData object for the content, because the bytes can be streamed from that object. This might be useful, if we can stream the data directly to the HTTP response. It would reduce the memory footprint of downloads significantly. Content length is also returned separately to facilitate the streaming. If the streaming is not feasible, the data can be easily accessed as a ByteArray.

Ways to connect in different scenarios:
- In the cloud, AZURE_CLIENT_ID etc. environment variables are read in application.properties to form a connection string.
- When run in Docker Compose, a HAITATON_BLOB_CONNECTION_STRING variable is set in docker-compose.yml. HAITATON_BLOB_CONNECTION_STRING overrides the default connection string in application.properties.
- If started with `./gradlew bootRun`, HAITATON_BLOB_CONNECTION_STRING is set in build.gradle.kts.
- In BlobFileClientITest, an Azurite container is started with Testcontainers. The connection string is handcrafted in the same file, along with the client.
- Other integration tests don't exist yet, but the hope is that they could use MockFileClient to reduce the number of different Spring configurations.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
I added this temp piece of code in BlobFileClient to test the client with Docker Compose and `./gradlew bootRun`.
```
    @EventListener(ApplicationReadyEvent::class)
    fun atStart() {
        val i = 4
        val testPath = "fake/test$i.txt"
        val originalFilename = "test$i.txt"
        val contentType = MediaType.TEXT_PLAIN
        val content = "This is test file #$i".toByteArray()

        println("<AtStart> Trying to delete a file.")
        val deleteResult = delete(Container.HANKE_LIITTEET, testPath)
        println("<AtStart> deleteResult $deleteResult")

        println("<AtStart> Uploading a file.")
        val options = BlobParallelUploadOptions(BinaryData.fromBytes(content))
        options.headers = BlobHttpHeaders()
        options.headers.setContentType(contentType.toString())
        options.headers.setContentDisposition("attachment; filename=$originalFilename")
        upload(Container.HANKE_LIITTEET, testPath, originalFilename, contentType, content)

        hankeAttachmentClient.listBlobs().forEach {
            println("<AtStart> ${it.name} ${it.properties.contentLength} ${it.properties.contentType} ${it.properties.contentDisposition}")
        }

        println("<AtStart> Downloading a file.")
        val downloaded = download(Container.HANKE_LIITTEET, testPath)
        println("<AtStart> ${downloaded.content.toStream().readAllBytes().toString(Charsets.UTF_8)}")
        println("<AtStart> Downloaded: $downloaded")
    }
```

# Checklist:

 or other location:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -100,6 +100,7 @@ services:
       HAITATON_PASSWORD: haitaton
       HAITATON_OAUTH2_CLIENT_ID: https://api.hel.fi/auth/haitatonapidev
       HAITATON_OAUTH2_USER_INFO_URI: https://tunnistamo.test.hel.ninja/openid/userinfo
+      HAITATON_BLOB_CONNECTION_STRING: BlobEndpoint=http://azurite:10000/devstoreaccount1;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;
       #      HAITATON_CORS_ALLOWED_ORIGINS: http://localhost:8000
       HAITATON_GDPR_DISABLED: ${HAITATON_GDPR_DISABLED:-true}
       ALLU_BASEURL: ${ALLU_BASEURL}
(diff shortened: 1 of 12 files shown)


umeetiusbaar (approved):
Looks good to me. Tested with the mock code in description.

PR #495, opened 2023-11-08
HAI-1969 Clean test annotations

corvidian:
# Description

`@Testcontainers` is an inheritable annotation, so if it's added to DatabaseTest, it doesn't need to be repeated in test classes that inherit DatabaseTest.

`webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT` in `@SpringBootTest` doesn't seem to do anything, so those can be removed.

`@TestPropertySource(locations =
["classpath:application-test.properties"])` is no longer needed, since all tests use the test profile, so `application-test.properties` is used automatically.

Also, change the name of the container property in DatabaseTest to something harder to collide with accidentally. Collisions can cause misleading error messages.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Tests still pass.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/DatabaseTest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/DatabaseTest.kt
@@ -5,6 +5,7 @@ import org.springframework.test.context.DynamicPropertySource
 import org.springframework.test.context.jdbc.Sql
 import org.springframework.test.context.jdbc.SqlMergeMode
 import org.testcontainers.junit.jupiter.Container
+import org.testcontainers.junit.jupiter.Testcontainers
 import org.testcontainers.utility.MountableFile
 
 /**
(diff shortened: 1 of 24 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/DatabaseTest.kt
```diff
@@ -19,10 +20,11 @@ import org.testcontainers.utility.MountableFile
  */
 @Sql("/clear-db.sql")
 @SqlMergeMode(SqlMergeMode.MergeMode.MERGE)
+@Testcontainers
 abstract class DatabaseTest {
```
umeetiusbaar, line 24: Could DatabaseTest be an interface instead of an abstract class?

corvidian: Looks like the container doesn't get started if it's an interface.

pitkni (approved):

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/DatabaseTest.kt
pitkni: The kdoc should be revised however.

PR #498, opened 2023-11-10
HAI-1969 Separate authorization and business logic in hanke attachments

corvidian:
# Description

Authorization and business logic have been intertwined in hanke attachments. Authorization has checked that the user has the correct permission for the hanketunnus in the URL, but checking that the application belongs to that hanke has been done in HankeAttachmentService.

Separate authorization and business logic by adding a new authorizer for hanke attachments. The authorizer checks that the user has the correct permission for the hanke and that the attachment belongs to that hanke. So we don't need to pass the hanketunnus to the service anymore.

Also, enable PR builds also for branches that target any branch starting
with HAI.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Downloading and deleting attachments works like before.

# Checklist:

 or other location:

### DIFF
--- a/azure-pipelines-hanke-service-pr.yml
+++ b/azure-pipelines-hanke-service-pr.yml
@@ -17,6 +17,7 @@ pr:
     - main
     - dev
     - releases/*
+    - HAI*
 
 # By default, use self-hosted agents
 pool: Default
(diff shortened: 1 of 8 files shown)


umeetiusbaar (approved):
Looks good to me

PR #499, opened 2023-11-10
HAI-1994 Split AttachmentContentService in two

corvidian:
# Description

AttachmentContentService can be split cleanly in two, leaving one class to handle application-related content and the other to handle hanke-related content.

The motivation is that the content service is getting larger with the addition of the Blob Storage support. Some of it is temporary while we support both Blob Storage and database storage, but some are permanent since we have to be more explicit with Blob Storage than with the database.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar (approved):
Looks good to me

PR #503, opened 2023-11-13
HAI-1967 Use default credential builder for cloud environments

corvidian:
# Description

Blob Storage can't use a connection string to connect to Azure AD (https://stackoverflow.com/questions/67715786/blob-storage-java-sdk-authorize-by-azure-ad).

Use the default credential builder when the connection string is not explicitly defined, i.e. in cloud environments. Also, add some diagnostic logging to startup.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/azure/BlobFileClientITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/azure/BlobFileClientITest.kt
@@ -41,7 +41,7 @@ class BlobFileClientITest : FileClientTest() {
         azuriteContainer.start()
         val connectionString =
             "BlobEndpoint=http://${azuriteContainer.host}:${azuriteContainer.firstMappedPort}/devstoreaccount1;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;"
-        serviceClient = AzureContainerServiceClient(connectionString).blobServiceClient()
+        serviceClient = AzureContainerServiceClient(connectionString, "").blobServiceClient()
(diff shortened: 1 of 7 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/azure/BlobFileClientITest.kt
```diff
@@ -41,7 +41,7 @@ class BlobFileClientITest : FileClientTest() {
         azuriteContainer.start()
         val connectionString =
             "BlobEndpoint=http://${azuriteContainer.host}:${azuriteContainer.firstMappedPort}/devstoreaccount1;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;"
-        serviceClient = AzureContainerServiceClient(connectionString).blobServiceClient()
+        serviceClient = AzureContainerServiceClient(connectionString, "").blobServiceClient()
```
umeetiusbaar, line 44: Could the second parameter (`endpoint`) have a default value or could there be a second constructor in order to use it here like `AzureContainerServiceClient(connectionString)`?

corvidian: Adding a default value causes the application startup to fail because Spring can't find a default constructor. I think the default value gets compiled into several constructors.

To make this neater, there would have to be a default (empty) constructor in AzureContainerServiceClient, which would mean it would have to use field injection instead of constructor injection. I don't see this as an overall improvement.

umeetiusbaar: Sounds reasonable

umeetiusbaar (approved):
Looks good to me

PR #505, opened 2023-11-14
HAI-1994 Temp endpoint for moving attachment content to Blob

corvidian:
# Description

Add a temporary endpoint for moving attachment content to Blob Storage. The endpoint will take an attachment ID as its parameter and move the content of that attachment to the cloud.

The motivation is to make testing Blob functionality easier. Adding a single attachment to the cloud required 4 steps earlier, now it needs 2 and the steps are easier.

At least some of the service implementation can be used later for the background migration process (HAI-1964).

After using the UI to upload an attachment, look up its ID from the network tab, and use the Swagger UI to call http://localhost:3001/api/swagger-ui/index.html#/test-data-controller/replaceContent. Authorization is needed as usual. The attachment will be uploaded to Blob Storage and the content will be removed from the database.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar (approved):
This is a great help for testing

PR #512, opened 2023-11-16
HAI Remove explicit Spring Security version

corvidian:
# Description

The Spring Security version was specified explicitly while preparing for the Spring 3.0 upgrade. A special transitional version of Spring Security made the upgrade easier. When the upgrade to 3.1 was completed, this specification should've been deleted, but the version was updated instead.

Removing the version will make Spring Security follow Spring Boot's recommended version. So it will update as we update Spring Boot.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other


umeetiusbaar (approved):
👍

PR #514, opened 2023-11-20
HAI-2013 Delete Blobs when hanke is deleted

corvidian:
# Description

Use a batch operation to delete all the attachment Blobs of a hanke when the hanke is removed. The operation removes all Blobs matching the directory the hanke attachments are in, so it also removes any that might have been orphaned earlier. It can't handle subdirectories, though, since we don't have a use case for them. Implementing deleting objects from subdirectories seemed to require a recursive search for them.

Refactor HankeAttachmentFactory. It was confusing to have it as an interface with default functions. To enable the use of a fluid interface for creating content for the attachments, use a builder that stores the required dependencies and the saved entity. Remove functionality related to hanke attachments from AttachmentFactory and use HankeAttachmentFactory instead.

Mark Hanke and HankeEntity as implementations of HankeIdentifier, i.e. they have id and hanketunnus fields. This enables using HankeIdentifier as a convenient parameter type when we need just the two pieces of information from a hanke. Also, it enables the use of `logString()` from HankeIdentifier for both Hanke or HankeEntity, which is convenient since logging both the ID and hanketunnus of a hanke is very common and recommended.

Add a `createAll()` method to HankeLoggingService as a delegate for `AuditLogService.createAll()`. With this change, HankeService no longer needs a separate dependency on AuditLogService.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Add attachments to a hanke. Create a new hanke if necessary.
2. Use http://localhost:3001/api/swagger-ui/index.html#/test-data-controller/moveHankeAttachmentToCloud on some of them to move their content to Blob Storage (Azurite locally).
3. Check that there are files in Blob Storage using Azure Storage Explorer.
4. Remove the hanke.
5. Check that the files are gone from Azure Storage Explorer.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/azure/BlobFileClient.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/azure/BlobFileClient.kt
@@ -3,9 +3,11 @@ package fi.hel.haitaton.hanke.attachment.azure
 import com.azure.core.http.HttpHeaderName
 import com.azure.core.util.BinaryData
 import com.azure.storage.blob.BlobServiceClient
+import com.azure.storage.blob.batch.BlobBatchClientBuilder
 import com.azure.storage.blob.models.BlobErrorCode
 import com.azure.storage.blob.models.BlobHttpHeaders
 import com.azure.storage.blob.models.BlobStorageException
+import com.azure.storage.blob.models.DeleteSnapshotsOptionType
 import com.azure.storage.blob.options.BlobParallelUploadOptions
 import fi.hel.haitaton.hanke.attachment.common.DownloadNotFoundException
 import fi.hel.haitaton.hanke.attachment.common.DownloadResponse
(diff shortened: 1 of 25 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/azure/BlobFileClient.kt
```diff
@@ -84,10 +88,33 @@ class BlobFileClient(blobServiceClient: BlobServiceClient, containers: Container
         return response
     }
 
+    override fun deleteAllByPrefix(container: Container, prefix: String) {
```
pitkni, line 91: Suggestion: would it be better if the prefix was of type Number, representing either hanke.id or application.id? 

I guess it would be pretty difficult to find a suitable name for the function deleteAllByX(container: Container, id: Number), but that way the caller would not need to know about the inner functionality of the deletion logic.

corvidian: I think BlobFileClient shouldn't need to know the context where it's being used. The content service decides where the content is saved, so that should be the component that decides what to delete.

pitkni: Fair enough.

pitkni (approved):

PR #517, opened 2023-11-21
HAI-1963 Add ability to upload hanke attachments as blob

pitkni:
# Description

HAI-1963 Add ability to upload hanke attachments as blob

New hanke attachments are uploaded to an azure blob container. Furthermore, the required http calls, such as clamav and blob upload are moved outside of a database transaction.

Additional: remove Dockerfile as it is not needed (we have Dockerfile-platta and Dockerfile-local).

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Add attachment(s) for hanke, they should be saved and be retrievable from hanke blob container. From UI perspective, everything should work as before.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -110,3 +112,9 @@ fun String?.isValidBusinessId(): Boolean {
         false
     }
 }
+
+fun MultipartFile.validNameAndType() =
+    Pair(
+        AttachmentValidator.validFilename(originalFilename),
+        AttachmentValidator.ensureMediaType(contentType)
+    )
```
corvidian, line 120: Shouldn't this be in AttachmentValidator?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentPersister.kt
```diff
@@ -0,0 +1,53 @@
...
+import java.time.OffsetDateTime
+import mu.KotlinLogging
+import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Transactional
+
+private val logger = KotlinLogging.logger {}
+
+/** Needed for attachment services transaction management. See usages for reasoning. */
+@Component
+class AttachmentPersister(
```
corvidian, line 16: The responsibilities of AttachmentPersister and HankeAttachmentService are not very clearly separated. Both access the database using the same repositories. There's also very similar code in both of them regarding finding a hanke.

Did you consider adding a new service between HankeAttachmentController and HankeAttachmentService instead of adding a service under HankeAttachmentService?

So something like a new HankeAttachmentUploadService, which has the addAttachment function. It could delegate to HankeAttachmentService for database access. It could do the calls to scanClient and the content service. It could also handle the functionality that's been added to the controller, i.e. finding the hanke ID and validating the attachment name and type. The functionality in the persistor could be moved to HankeAttachmentService.

pitkni: I just wanted to have a simple implementation that handles the transactions without http calls. 

I guess it would be possible to handle the transaction programmatically, but it would not be very spring boot style.

I think the separation of responsibilities in  HankeAttachmentUploadService and HankeAttachmentService would not be very intuitive / clear either. 

In short, this is a bit weird looking extra class but it is needed for transactional proxying to work. 

This being said, I will try to see if a new service between the controller and service would work.

pitkni: Fixed as suggested (as I understood it). How about now?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
pitkni: Moved.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentPersister.kt
corvidian: I think it's cleaner now.

corvidian (approved):

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentUploadServiceITest.kt
```diff
@@ -0,0 +1,98 @@
...
+@ActiveProfiles("test")
+@WithMockUser(USERNAME)
+class AttachmentUploadServiceITest(
+    @Autowired private val attachmentUploadService: AttachmentUploadService,
+    @Autowired private val attachmentRepository: HankeAttachmentRepository,
+    @Autowired private val hankeFactory: HankeFactory,
+    @Autowired private val fileClient: MockFileClient
+) : DatabaseTest() {
+
+    private lateinit var mockClamAv: MockWebServer
```
corvidian, line 40: mockClamAv can be removed from HankeAttachmentServiceITests.

pitkni: I don't think it can. Uploadattachment scans the input file.

corvidian: I was talking about this, but there was no change on the line to comment on:
https://github.com/City-of-Helsinki/haitaton-backend/blob/8ea7659a520212ed3a49c3cf8b8fa79086801b24/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/hanke/HankeAttachmentServiceITests.kt#L64

I should had been clearer. I'll remove it in #515.

PR #525, opened 2023-11-27
HAI Upgrade Spring Boot to 3.2.8

corvidian:
# Description

Some fixes were needed to get the upgrade to work:

- Add an exception handler for NoResourceFoundException. After the
  update, TestDataController endpoints started to return 500 Internal
  error when they were called with the controller disabled by environment
  variables. Before the update, they returned 404 Not Found. The added
  exception handler returns the expected behaviour.
- Disable AlluUpdateService and LockService in tests. They're not
  used, but they cause problems when the test context is shutting down.
  There's a problem with TestContainer and Spring lifetimes, where the
  test container is shutdown before the application context, causing the
  database connections to time out. The thing trying to call the
  database at shutdown is the JdbcLockRegistry.
- Use reusable containers. This is an alternative solution to the
  previous problem.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/build.gradle.kts
+++ b/services/hanke-service/build.gradle.kts
@@ -49,7 +49,7 @@ spotless {
 
 plugins {
     val kotlinVersion = "2.0.0"
-    id("org.springframework.boot") version "3.1.12"
+    id("org.springframework.boot") version "3.2.8"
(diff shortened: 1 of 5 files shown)


umeetiusbaar (approved):
Looking great!

PR #545, opened 2023-12-07
HAI-1987 Refactor application attachment services

corvidian:
# Description

Split DB operations from ApplicationAttachmentService to a new service, ApplicationAttachmentMetadataService. This allows us to do Allu and Blob operations without having a DB transaction running.

ApplicationAttachmentMetadataService is only called from ApplicationAttachmentService, so other services don't have to know the difference between the two.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Nothing should change.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/Domain.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/Domain.kt
@@ -1,0 +1,50 @@
+package fi.hel.haitaton.hanke.attachment.common
+
+import fi.hel.haitaton.hanke.allu.Attachment as AlluAttachment
+import fi.hel.haitaton.hanke.allu.AttachmentMetadata
+import java.time.OffsetDateTime
+import java.util.UUID
+
+sealed interface Attachment {
+    val id: UUID
+    val fileName: String
+    val contentType: String
+    val createdByUserId: String
+    val createdAt: OffsetDateTime
+    val blobLocation: String?
+}
+
+data class ApplicationAttachment(
+    override val id: UUID,
+    override val fileName: String,
+    override val contentType: String,
+    override val createdByUserId: String,
+    override val createdAt: OffsetDateTime,
+    override val blobLocation: String?,
+    val applicationId: Long,
+    val attachmentType: ApplicationAttachmentType,
+) : Attachment {
+    fun toDto(): ApplicationAttachmentMetadata {
+        return ApplicationAttachmentMetadata(
+            id = id,
+            fileName = fileName,
+            createdAt = createdAt,
+            createdByUserId = createdByUserId,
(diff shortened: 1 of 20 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/Domain.kt
```diff
@@ -0,0 +1,50 @@
...
+sealed interface Attachment {
+    val id: UUID
+    val fileName: String
+    val contentType: String
+    val createdByUserId: String
+    val createdAt: OffsetDateTime
+    val blobLocation: String?
+}
+
+data class ApplicationAttachment(
```
pitkni, line 17: The stuff in common/AttachmentMetaData should probably be also moved here. Or vice versa. And there may be a need for some renaming.

I like to think that an Attachment should have both metadata and content. So there would be three object types: AttachmentMetadata, AttachmentContent and together they create Attachment. If there is no use cases for having a whole Attachment, we could just have metadata and content separately.

corvidian: The classes in AttachmentMetadata are used as DTOs, where these are purely domain classes. So there's a clear distinction between them. Maybe the difference could be emphasized by renaming AttachmentData.kt to Dto.kt. Or by renaming the AttachmentMetadata classes to AttachmentDto, AttachmentMetadataDto or AttachmentMetadataResponse. Then we could use the AttachmentMetadata name for the domain class.

corvidian: The `AttachmentContent` and `HankeAttachmentWithContent` are used as domain classes. I'll move them to Domain.kt.

umeetiusbaar: Should we use separate classes instead of Domain.kt? Maybe `domain` package?

corvidian: In https://github.com/City-of-Helsinki/haitaton-backend/pull/545/commits/da7c583de17ce747acf6cd7031b70bde8ca02872, I renamed:
- HankeAttachment to HankeAttachmentMetadataDto
- ApplicationAttachmentMetadata to ApplicationAttachmentMetadataDto
- ApplicationAttachment to ApplicationAttachmentMetadata
- AttachmentMetadata.kt-file to `Dto.kt`

I also moved AttachmentContent and HankeAttachmentWithContent to `Domain.kt`.

Is this ok?

pitkni (approved):

umeetiusbaar (approved):

PR #552, opened 2023-12-12
HAI-2173 Adjust bus traffic nuisance index calculation

umeetiusbaar:
# Description

Adjusted bus traffic hindrance calculation based on new specification: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8236367985/Linja-autojen+paikallisliikenne

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing
This should be first merged into [HAI-2110/adjust-car-traffic-hindrance-index-calculation](https://github.com/City-of-Helsinki/haitaton-backend/tree/HAI-2110/adjust-car-traffic-hindrance-index-calculation) and then tested with it.
Testing in UI is only possible if first check the bus lines and rush hour traffic in database with QGIS.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTormaysService.kt
```diff
@@ -168,9 +168,9 @@ enum class TormaystarkasteluBussiRunkolinja(val runkolinja: String) {
 
     fun toLinjaautoliikenneluokittelu(): Linjaautoliikenneluokittelu =
         when (this) {
-            ON -> Linjaautoliikenneluokittelu.RUNKOLINJA
-            LAHES -> Linjaautoliikenneluokittelu.RUNKOLINJAMAINEN
-            EI -> Linjaautoliikenneluokittelu.PERUS
+            ON -> Linjaautoliikenneluokittelu.RUNKOLINJA_TAI_ENINTAAN_20_VUOROA_RUUHKATUNNISSA
+            LAHES -> Linjaautoliikenneluokittelu.ENINTAAN_10_VUOROA_RUUHKATUNNISSA
+            EI -> Linjaautoliikenneluokittelu.EI_RUUHKAAIKANA
```
corvidian, line 173: This doesn't make much sense anymore.

umeetiusbaar: How so?

corvidian: Why does being almost a trunk line lead to having 10 buses?
```kotlin
LAHES -> Linjaautoliikenneluokittelu.ENINTAAN_10_VUOROA_RUUHKATUNNISSA
```

Does this mean this is not a trunk line at rush hour?
```kotlin
EI -> Linjaautoliikenneluokittelu.EI_RUUHKAAIKANA
```

umeetiusbaar: This is mostly from old implementation but there should not be much changes to it. New specification says:
```
2 = "Linja-autojen paikallisliikennettä muulloin, kuin ruuhka-aikana"
3 = "Linja-autojen paikallisliikenteen linjojen vuoromäärä enintään 10 ruuhkatunnissa"
4 = "Linja-autojen paikallisliikenteen runkolinja tai vuoromäärä enintään 20 ruuhkatunnissa"
```
That `when` logic follows this specification in my opinion: 2 means that there is no "runkolinja", 3 means that there is "almost" some "runkolinja" and "yes" means that there is "runkolinja".

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/Enums.kt
```diff
@@ -35,9 +35,10 @@ enum class Linjaautoliikenneluokittelu(override val value: Int) : Luokittelu {
...
         5
     ),
-    RUNKOLINJA_TAI_ENINTAAN_20_VUOROA_RUUHKAAIKANA(4),
-    RUNKOLINJAMAINEN_LINJA_TAI_ENINTAAN_10_VUOROA_RUUHKAAIKANA(3),
-    ENINTAAN_5_VUOROA_RUUHKAAIKANA_TAI_LINJOJA_MUUNA_AIKANA(2),
+    RUNKOLINJA_TAI_ENINTAAN_20_VUOROA_RUUHKATUNNISSA(4),
+    ENINTAAN_10_VUOROA_RUUHKATUNNISSA(3),
+    EI_RUUHKAAIKANA(2),
+    MAHDOLLINEN_POIKKEUSREITTI(1),
     EI_VAIKUTA_LINJAAUTOLIIKENTEESEEN(0)
```
corvidian, line 42: The enum names could be made simpler, with more detailed descriptions as documentation.
```kotlin
    /**
     * Rautatientori, Kamppi, Mannerheimintie, Kaisaniemenkatu, Hämeentie or over 20 buses per hour.
     */
    TARKEIMMAT_JOUKKOLIIKENNEKADUT(5),
    /** Trunk line (runkolinja) or over 10 buses per hour. */
    RUNKOLINJA(4),
    /** At most 10 buses per hour during rush hour. */
    VUOROJA_RUUHKAAIKANA(3),
    /** No regular buses during rush hour. The might be buses at other times, like night buses. */
    EI_VUOROJA_RUUHKAAIKANA(2),
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTormaysService.kt
corvidian: I can't find anything about the "almost" trunk lines from the new specification. Collecting those values seems to have been removed from the material update scripts. Maybe we can remove the `LAHES` value completely?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/RajaArvoLuokittelija.kt
```diff
@@ -22,16 +22,14 @@ object RajaArvoLuokittelija {
 
     fun linjaautoliikenteenRuuhkavuoroluokka(countOfRushHourBuses: Int) =
         when {
-            countOfRushHourBuses >= 21 ->
+            countOfRushHourBuses > 20 ->
```
corvidian, line 25: Doesn't this need editing as well?

https://github.com/City-of-Helsinki/haitaton-backend/blob/3a938e130d9fc21c90fc68b24f135fa7175b2efd/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluLaskentaService.kt#L130-L131

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTormaysService.kt
umeetiusbaar: You are correct. I removed "almost".

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/Enums.kt
umeetiusbaar: I renamed enum values and added comments.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/RajaArvoLuokittelija.kt
umeetiusbaar: You are right! I missed this. Removed the unnecessary if-statement.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluLaskentaService.kt
```diff
@@ -132,15 +132,13 @@ class TormaystarkasteluLaskentaService(
...
-                .KAMPPI_RAUTATIENTORI_MANNERHEIMINTIE_KAISANIEMENKATU_HAMEENTIE_TAI_YLI_20_VUOROA_RUUHKATUNNISSA
-                .value
+            return Linjaautoliikenneluokittelu.TARKEIMMAT_JOUKKOLIIKENNEKADUT.value
         }
 
         val bussireitit = tormaysService.getIntersectingBusRoutes(geometriaIds)
 
         val valueByRunkolinja =
-            bussireitit.maxOfOrNull { it.runkolinja.toLinjaautoliikenneluokittelu().value }
+            bussireitit.maxOfOrNull { it.runkolinja?.toLinjaautoliikenneluokittelu()?.value ?: 0 }
```
corvidian, line 141: Can we safely assume zero is a good default value in case the value is somehow missing from the database? Would it be better to fail the operation if our database is so badly messed up?

Why was this change made in an unrelated "PR comment changes" commit?

umeetiusbaar: I was thinking of those "almost" values that possibly are in the database. At least I have them in my current database. Default value EI could be used for those cases.

corvidian: I think it would be better for the code to fail than to continue with possible faulty values.

If I understood this correctly, the only this can fail is if the bus data in the database is from before "almost" was dropped from the processing result, i.e. from before 9.11.2022. Or if someone has added bad data directly to the DB. So if this fails in an environment, we definitely should update the bus data in that environment.

umeetiusbaar: "almost" values are still present in at least dev and test environments:
```
haitatonhanketietokantatest=> select distinct trunk from tormays_buses_polys;
 trunk  
--------
 no
 yes
 almost
(3 rows)
```

umeetiusbaar: I will try to generate a new git package and see what kind of data it holds.

umeetiusbaar: Okay, new gpkg generated does not contain any "almost" rows. So, it needs to be updated as well.

umeetiusbaar: I made it throw an exception when there is other value than 'yes' or 'no'.

corvidian (approved):

PR #553, opened 2023-12-13
HAI-2154 Add client for Profiili GraphQL API

corvidian:
# Description

Add a client for getting the verified name of the user from Profiili GraphQL API. This allows us to get the verified, official name of the user.

All the GraphQL libraries I found were really complicated with separate Gradle plugins, code generation and that fun stuff. We need just one static query, so using code generation to accomplish that felt like an overkill. Instead, add the GraphQL query in a resource file as a literal query and send it to the API server wrapped in an almost static JSON object.

Also, add a test data controller endpoint for querying the caller's own verified name. This was added to facilitate testing the endpoint in different environments, but also because that information is not readily available elsewhere.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Call the new test data controller from Swagger-UI. Make sure you have Suomi.fi login enabled.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/profiili/ProfiiliClientITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/profiili/ProfiiliClientITest.kt
@@ -82,7 +82,6 @@ class ProfiiliClientITest {
 
         @Test
         fun `throws exception when API tokens not found`() {
-            clearMocks(securityContext)
             mockAccessToken()
             mockApiTokensApi.enqueue(MockResponse().setResponseCode(404))
 
@@ -98,7 +97,6 @@ class ProfiiliClientITest {
 
         @Test
         fun `throws exception when API tokens is empty`() {
-            clearMocks(securityContext)
             mockAccessToken()
             mockApiTokensApi.enqueueSuccess("{}")
 
@@ -114,7 +112,6 @@ class ProfiiliClientITest {
 
         @Test
         fun `throws exception when there's no API token for the correct audience`() {
-            clearMocks(securityContext)
             mockAccessToken()
             mockApiTokensApi.enqueueSuccess(mapOf("some-other-audience" to "some-other-token"))
 
@@ -130,7 +127,6 @@ class ProfiiliClientITest {
 
         @Test
         fun `calls the API token endpoint with correct url and headers`() {
-            clearMocks(securityContext)
             mockAccessToken()
             mockApiTokensApi.enqueueSuccess(mapOf("some-other-audience" to "some-other-token"))
 
@@ -148,7 +144,6 @@ class ProfiiliClientITest {
 
         @Test
         fun `throws exception when GraphQL request fails`() {
-            clearMocks(securityContext)
             mockAccessToken()
             mockApiToken()
             mockGraphQl.enqueue(MockResponse().setResponseCode(404))
@@ -181,7 +176,6 @@ class ProfiiliClientITest {
 
         @Test
         fun `returns null when user's profile doesn't have verified information`() {
-            clearMocks(securityContext)
             mockAccessToken()
             mockApiToken()
             mockGraphQl.enqueueSuccess(ProfiiliResponse(ProfiiliData(MyProfile(null))))
@@ -196,7 +190,6 @@ class ProfiiliClientITest {
 
         @Test
         fun `returns name when user's profile has verified information`() {
-            clearMocks(securityContext)
             mockAccessToken()
             mockApiToken()
             mockGraphQl.enqueueSuccess(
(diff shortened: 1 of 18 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/profiili/ProfiiliClientITest.kt
```diff
@@ -0,0 +1,316 @@
...
+                hasClass(VerifiedNameNotFound::class)
+                messageContains("User not authenticated")
+            }
+            assertThat(mockApiTokensApi.requestCount).isEqualTo(0)
+            verify { securityContext.accessToken() }
+        }
+
+        @Test
+        fun `throws exception when API tokens not found`() {
+            clearMocks(securityContext)
```
umeetiusbaar, line 85: Is `clearMocks` needed here when there is `clearAllMocks` in `setUp`?

services/hanke-service/src/test/resources/application-test.yml
```diff
@@ -13,6 +13,9 @@ haitaton:
   features:
     hanke-editing: true
     user-management: true
+  profiili-api:
```
umeetiusbaar, line 16: Where are these used?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/profiili/ProfiiliData.kt
```diff
@@ -0,0 +1,9 @@
+package fi.hel.haitaton.hanke.profiili
+
+data class ProfiiliResponse(val data: ProfiiliData)
+
+data class ProfiiliData(val myProfile: MyProfile?)
+
+data class MyProfile(val verifiedPersonalInformation: Names?)
+
+data class Names(val firstName: String, val lastName: String, val givenName: String)
```
pitkni, line 9: The difference of firstName and givenName could be explained in a kdoc or something.

services/hanke-service/src/test/resources/application-test.yml
corvidian: They are loaded into ProfiiliProperties.

umeetiusbaar: I meant the values `http://localhost:14678/*`. What service is this? And why these properties are under `test` when there is no other changes there? All other test changes are under `integrationTest`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/PublicHankeController.kt
```diff
@@ -40,5 +40,7 @@ class PublicHankeController(private val hankeService: HankeService) {
                 )
             ]
     )
-    fun getAll(): List<PublicHanke> = hankeService.loadPublicHanke().map { hankeToPublic(it) }
+    fun getAll(): List<PublicHanke> {
```
pitkni, line 43: Why the change? It's the same thing but with three lines.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/profiili/ProfiiliClientITest.kt
corvidian: These were left over from some debugging I did earlier. Removed now.

services/hanke-service/src/test/resources/application-test.yml
corvidian: I was planning on using that port for the mockWebService in ProfiiliClientITest, but I used two random ports instead. These could probably be replaced with tokens that are clearly non-functional. There should be something here to override the default value in application.yml, so that we don't accidentally connect to the actual backend from integration tests.

Both tests and integration tests use the test Spring profile, so they use the same application-test.yml. Integration tests can see this file from the src/test/ directory, but tests can't access anything in src/intergationTest/, so the file needs to be here under src/test/.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/profiili/ProfiiliData.kt
corvidian: I added documentation here.

umeetiusbaar (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/PublicHankeController.kt
corvidian: I used the endpoint for some early prototyping and didn't notice I left a change. Reverted this. Thank you for noticing.

pitkni (approved):

PR #554, opened 2023-12-13
HAI-1996 Delete application attachment blobs

pitkni:
# Description

Add logic to delete application attachment blobs if the blob location is defined.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing
The blob upload functionality for applications is not yet implemented. Hence, there is no blobs to delete. The attachment deletion should however work as before from the UI perspective.

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
@@ -116,7 +116,7 @@ class ApplicationAttachmentServiceITest(
         }
 
         @Test
-        fun `Return empty when application doesn't have attachments`() {
+        fun `Returns empty when application doesn't have attachments`() {
             val application = alluDataFactory.saveApplicationEntity(USERNAME)
 
             val result = attachmentService.getMetadataList(application.id!!)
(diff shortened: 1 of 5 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
```diff
@@ -103,30 +107,30 @@ class ApplicationAttachmentServiceITest(
...
 
-            val result = applicationAttachmentService.getMetadataList(applicationId)
+            val result = attachmentService.getMetadataList(applicationId)
 
             assertThat(result).isEmpty()
         }
 
         @Test
-        fun `return empty when application doesn't have attachments`() {
+        fun `Return empty when application doesn't have attachments`() {
```
umeetiusbaar, line 119: "Returns", "Return" "Returns". Just a small detail but since you changed r --> R I noticed this.

umeetiusbaar:
Tested by creating a new application and adding an attachment to it in UI. Then added the attachment to azure by Azure Container Explorer and changed the database manually by adding `blob_location` to the right `application_attachment` row. Then when deleting the attachment in UI the file was deleted in blob storage as well as the database rows. So, it works ok.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
pitkni: Typo fixed.

umeetiusbaar (approved):

PR #557, opened 2023-12-15
HAI-2154 Make hanke creation simpler

corvidian:
# Description

Make hanke creation endpoint take only hanke name and founder email and phone number as data. Anything else is ignored. Founder name will be fetched from Profiili API.

Create hanke request and the create method in hanke service have been very widely used in test data initialization. This means a lot of changes, since most of the code relating to creating hanke entity for testing needed to be rewritten. The new service method for creating a hanke needs a mocked security context and doesn't produce very useful object on it's own, so I created a builder to hide the complexities. I considered adopting the builder for all hanke creation in tests, but this PR is too big already.

Since CreateHankeRequest no longer implemented BaseHanke, the interface only had one implementation, so it was removed. This leaves other classes in the file `BaseHanke.kt`. They all have better places elsewhere, but again, this PR is too large even without any additional refactoring.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
0. Make sure you are using Suomi.fi identification for logging in.
1. Use [Swagger-UI](http://localhost:3001/api/swagger-ui/index.html#/hanke-controller/createHanke) or cURL to create a hanke.
2. You should be able to find the hanke from omat hankkeet.
3. There should be a user with the name from Suomi.fi and the email and phone number you provided when creating the hanke.
4. You should be able to edit the hanke from the UI.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/CreateHankeRequest.kt
```diff
@@ -15,94 +14,14 @@ data class CreateHankeRequest(
...
-    )
-    val tyomaaTyyppi: Set<TyomaaTyyppi>? = null,
-    @field:Schema(
-        description =
-            "Hanke areas data. At least one alue is required for the hanke to be published.",
-    )
-    override val alueet: List<NewHankealue>? = null,
-) : BaseHanke
+    val nimi: String,
+    val perustaja: HankePerustaja
```
pitkni, line 18: This could have a swagger description also.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/CreateHankeRequestValidator.kt
```diff
@@ -0,0 +1,62 @@
...
+    val payload: Array<KClass<out Payload>> = []
+)
+
+class CreateHankeRequestValidator :
+    ConstraintValidator<ValidCreateHankeRequest, CreateHankeRequest> {
+    override fun isValid(
+        request: CreateHankeRequest?,
+        context: ConstraintValidatorContext
+    ): Boolean {
+        if (request == null) {
```
pitkni, line 32: Is it a possible scenario that the request is null? It is non nullable in the controller endpoint.

pitkni: Or is this executed before that? In that case this should be ok to provide a more specific error.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -138,3 +140,8 @@ fun String?.isValidBusinessId(): Boolean {
         false
     }
 }
+
+fun <T, C : Collection<T>> C.inspect(f: (T) -> Unit): C {
```
pitkni, line 144: Would it be better to rename this to for example applyEach or something similar? I was having a hard time to understand what inspect is supposed to do.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeService.kt
```diff
@@ -70,77 +71,61 @@ class HankeService(
...
     ): Hanke {
-        val userId = currentUserId()
-
-        val entity: HankeEntity = createEntityFromCreateRequest(request, generated, userId)
+        val entity: HankeEntity =
+            createEntityFromCreateRequest(request, generated, securityContext.userId())
 
-        val loggingEntryHolder = prepareLogging(entity)
-        copyYhteystietosToEntity(request, entity, userId, loggingEntryHolder, mutableMapOf())
+        alueet?.let {
```
pitkni, line 81: Creating a new hanke through hanke controller will not have any areas.

Would it make sense to have two different functions, one for creating a regular hanke and one for cable report situations? These of course have common functionalities which could also be in a separate function.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/CreateHankeRequest.kt
corvidian: Thank you for noticing. I added a description here and for the fields in HankePerustaja.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/CreateHankeRequestValidator.kt
corvidian: I couldn't find a scenario where the request would be null. When I tried with a broken or an empty body, I got an `org.springframework.http.converter.HttpMessageNotReadableException` exception before the validator was called. I made the request non-nullable, which made this method much simpler. Good catch.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
corvidian: Turns out there's a premade method for this: onEach. Using that instead.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeService.kt
corvidian: I fiddled around a bit with this and split the core of createHanke into two parts, `createNewEntity` and `saveCreatedHanke`. `createHanke` now only does these two steps. `generateHankeFrom` calls those methods and adds the areas in-between. I like that now `generateHankeFrom` doesn't have to create a DTO in the middle of processing, and the transactions are clearer.

pitkni (approved):

corvidian:
Waiting for UI, HAI-2153.

PR #560, opened 2023-12-21
HAI-2014 Delete application attachments from blob storage

umeetiusbaar:
# Description

Application attachment content is deleted from blob storage instead of database.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
* Create an application or use an existing application
* Add an attachment into blob storage's `haitaton-hakemusliitteet-local` container (by using Microsoft Azure Storage Explorer or Azure CLI)
* Connect the attachment to the application by adding `blob_location` for the correct `application_attachment` row. `blob_location` is the path of the attachment in Azure Storage Explorer.
* Delete the attachment in UI
* Check that the attachment is no longer in blob storage and the `application_attachment` row is also deleted.

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
@@ -40,6 +40,9 @@ import fi.hel.haitaton.hanke.allu.ApplicationStatus.PENDING_CLIENT
 import fi.hel.haitaton.hanke.allu.CableReportService
 import fi.hel.haitaton.hanke.asJsonResource
 import fi.hel.haitaton.hanke.asUtc
+import fi.hel.haitaton.hanke.attachment.application.ApplicationAttachmentContentService
+import fi.hel.haitaton.hanke.attachment.common.ApplicationAttachmentRepository
+import fi.hel.haitaton.hanke.attachment.common.MockFileClientExtension
 import fi.hel.haitaton.hanke.domain.Hankealue
 import fi.hel.haitaton.hanke.email.EmailSenderService.Companion.translations
 import fi.hel.haitaton.hanke.email.textBody
@@ -112,6 +115,7 @@ import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.api.extension.RegisterExtension
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.EnumSource
@@ -131,6 +135,7 @@ private val dataWithoutAreas = createCableReportApplicationData(areas = listOf()
 @SpringBootTest
 @ActiveProfiles("test")
 @WithMockUser(USERNAME)
+@ExtendWith(MockFileClientExtension::class)
 class ApplicationServiceITest : DatabaseTest() {
 
     @Autowired private lateinit var applicationService: ApplicationService
@@ -139,6 +144,9 @@ class ApplicationServiceITest : DatabaseTest() {
     @Autowired private lateinit var cableReportServiceAllu: CableReportService
 
     @Autowired private lateinit var applicationRepository: ApplicationRepository
+    @Autowired
+    private lateinit var applicationAttachmentContentService: ApplicationAttachmentContentService
+    @Autowired private lateinit var applicationAttachmentRepository: ApplicationAttachmentRepository
     @Autowired private lateinit var hankeRepository: HankeRepository
     @Autowired private lateinit var alluStatusRepository: AlluStatusRepository
     @Autowired private lateinit var auditLogRepository: AuditLogRepository
@@ -1292,7 +1300,7 @@ class ApplicationServiceITest : DatabaseTest() {
                     hanke = initializedHanke(),
                     applicationData = mockApplicationDataWithArea()
                 )
-            attachmentFactory.saveAttachment(application.id!!)
+            attachmentFactory.saveAttachmentToDb(application.id!!)
             val applicationData = application.applicationData as CableReportApplicationData
             val pendingApplicationData = applicationData.copy(pendingOnClient = false)
             val expectedAlluRequest = pendingApplicationData.toAlluData(HANKE_TUNNUS)
@@ -1362,7 +1370,7 @@ class ApplicationServiceITest : DatabaseTest() {
                     hanke = initializedHanke(),
                     applicationData = mockApplicationDataWithArea()
                 )
-            attachmentFactory.saveAttachment(application.id!!)
+            attachmentFactory.saveAttachmentToDb(application.id!!)
             val applicationData = application.applicationData as CableReportApplicationData
             val pendingApplicationData = applicationData.copy(pendingOnClient = false)
             val expectedAlluRequest = pendingApplicationData.toAlluData(HANKE_TUNNUS)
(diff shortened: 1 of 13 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentService.kt
```diff
@@ -48,9 +53,14 @@ class ApplicationAttachmentContentService(
                 throw AttachmentNotFoundException(attachmentId)
             }
 
+    fun exists(location: String): Boolean = fileClient.exists(Container.HAKEMUS_LIITTEET, location)
+
+    fun existsForApplication(applicationId: Long): Boolean =
+        fileClient.existsByPrefix(Container.HAKEMUS_LIITTEET, prefix(applicationId))
```
corvidian, line 59: These are not called from outside tests. Tests can use the FileClient directly.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -1423,16 +1431,31 @@ class ApplicationServiceITest : DatabaseTest() {
...
             val hanke = hankeFactory.saveMinimal()
             val application =
                 applicationFactory.saveApplicationEntity(USERNAME, hanke = hanke, alluId = null)
+            // assert that application is saved to database
             assertThat(applicationRepository.findAll()).hasSize(1)
+            val attachment1 = attachmentFactory.saveAttachment(application.id!!)
+            val attachment2 = attachmentFactory.saveAttachment(application.id!!)
+            // assert that attachments are saved to cloud
+            assertTrue(applicationAttachmentContentService.exists(attachment1.blobLocation!!))
+            assertTrue(applicationAttachmentContentService.exists(attachment2.blobLocation!!))
```
corvidian, line 1444: I think it's sufficient to check that there are 2 attachments. 
```suggestion
            assertThat(fileClient.listBlobs(Container.HAKEMUS_LIITTEET)).hasSize(2)
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -1423,16 +1431,31 @@ class ApplicationServiceITest : DatabaseTest() {
...
+            assertThat(applicationAttachmentRepository.findByApplicationId(application.id!!))
+                .hasSize(2)
 
             applicationService.delete(application.id!!, USERNAME)
 
+            // assert that application is deleted from database
             assertThat(applicationRepository.findAll()).isEmpty()
             verify { cableReportServiceAllu wasNot Called }
+            // assert that attachments are deleted from cloud
+            assertFalse(applicationAttachmentContentService.existsForApplication(application.id!!))
```
corvidian, line 1455: If there are no attachments, there are none for the application either.
```suggestion
            assertThat(fileClient.listBlobs(Container.HAKEMUS_LIITTEET)).isEmpty()
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
```diff
@@ -1423,16 +1431,31 @@ class ApplicationServiceITest : DatabaseTest() {
...
+            // assert that attachments are saved to cloud
+            assertTrue(applicationAttachmentContentService.exists(attachment1.blobLocation!!))
+            assertTrue(applicationAttachmentContentService.exists(attachment2.blobLocation!!))
+            // assert that attachment metadata is saved to database
+            assertThat(applicationAttachmentRepository.findByApplicationId(application.id!!))
+                .hasSize(2)
 
             applicationService.delete(application.id!!, USERNAME)
 
+            // assert that application is deleted from database
```
corvidian, line 1451: I think the asserts themselves are pretty self-explanatory. Commenting on what the next line of code does is more distracting than useful.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/ApplicationAttachmentFactory.kt
```diff
@@ -58,18 +58,25 @@ class ApplicationAttachmentFactory(
         return ApplicationAttachmentBuilder(entity, attachmentRepository, this)
     }
 
-    fun saveAttachment(applicationId: Long): ApplicationAttachmentEntity {
+    fun saveAttachmentToDb(applicationId: Long): ApplicationAttachmentEntity {
```
corvidian, line 61: There's nowhere else to save this than to the DB.

corvidian: I think I should have removed this method in #537 when I introduced the builder. Now there are several ways to create the same attachment. Maybe the uses of this method could be replaced with `save(application = application).withDbContent()`?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/ApplicationAttachmentFactory.kt
```diff
@@ -58,18 +58,25 @@ class ApplicationAttachmentFactory(
...
+        return applicationAttachmentEntity
+    }
+
     fun saveContentToDb(attachmentId: UUID, bytes: ByteArray): ApplicationAttachmentContentEntity {
         val entity = ApplicationAttachmentContentEntity(attachmentId, bytes)
         return contentRepository.save(entity)
     }
 
-    fun saveContentToCloud(
+    fun saveContent(
```
corvidian, line 79: Renaming this method could wait until there's just one place to store the content in.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentService.kt
umeetiusbaar: Removed "exists" methods

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
umeetiusbaar: Refactored to use fileClient. Removed unnecessary `applicationAttachmentContentService` from this test class.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
umeetiusbaar: Added `list` method to FileClient:
```
fun list(container: Container, prefix: String): List<String>
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
umeetiusbaar: You're right. Removed comments. I added them for myself mostly to make things more clear in tests.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/ApplicationAttachmentFactory.kt
umeetiusbaar: Removed saveAttachmentToDb and saveAttachmentToCloud methods and replaced them with what you suggested.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/ApplicationAttachmentFactory.kt
umeetiusbaar: Yes, you're right, I went too far in this task.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
corvidian: If there are no attachments, there are none for the application. So the prefix isn't needed, at least not here.

We don't want to add test code to production whenever we can avoid it. Adding the method to FileClient means that BlobFileClient will have the same method, but it's not used anywhere. If the method is needed, it could be added to MockFileClient, so it doesn't cause changes on the production side. The test class can inject MockFileClient instead of FileClient since we know it's going to get a MockFileClient anyway.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/ApplicationAttachmentFactory.kt
corvidian: You renamed `saveContentToDb` to `saveContent` instead?

umeetiusbaar: Sorry, that was also a mistake.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationServiceITest.kt
umeetiusbaar: Good point. Thank you!

corvidian (approved):

PR #561, opened 2023-12-21
HAI-2159 Add endpoint for adding a new user

corvidian:
# Description

Add an endpoint for adding a new user. The endpoint adds a hankekäyttäjä and a käyttäjäkutsu for the user and sends an invitation email to the hanke. The request is declined if a user already has the same email.

The invitation email is only sent if the caller has a hankekäyttäjä, and not just a permission. This limitation will not matter with a new hanke created with the kortisto-model.

Add first and last names as new fields to HankeKayttajaDto. With this, the response fields match the request fields. Keep the full name field in the DTO for compatibility.

Add a new permission code for adding a new user. The existing codes weren't a good match. The only one with the correct käyttöoikeustaso is RESEND_INVITATION, which is not descriptive of this task.

Override Sentry DSN to empty value in tests to disable Sentry completely.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Create a hanke and call the endpoint from [Swagger-UI](http://localhost:3001/api/swagger-ui/index.html#/hanke-kayttaja-controller/createNewUser). Check that the new user is added from the access rights management page.

For checking that the invitation email is sent, use a generated hanke.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/PermissionUpdate.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/PermissionUpdate.kt
@@ -11,3 +11,25 @@ data class PermissionDto(
     @field:Schema(description = "HankeKayttaja ID") val id: UUID,
     @field:Schema(description = "New access level in Hanke") val kayttooikeustaso: Kayttooikeustaso,
 )
+
+data class NewUserRequest(
+    @field:Schema(
+        description = "The first name of the user",
+    )
+    val etunimi: String,
+    @field:Schema(
+        description = "The last name of the user",
+    )
+    val sukunimi: String,
+    @field:Schema(
+        description = "The email address of the user. The invitation will be sent to this address.",
+    )
+    val sahkoposti: String,
+    @field:Schema(
+        description = "The phone number of the user.",
+    )
+    val puhelinnumero: String,
+) {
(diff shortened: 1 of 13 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/PermissionUpdate.kt
```diff
@@ -11,3 +11,25 @@ data class PermissionDto(
...
+    @field:Schema(
+        description = "The email address of the user. The invitation will be sent to this address.",
+    )
+    val sahkoposti: String,
+    @field:Schema(
+        description = "The phone number of the user.",
+    )
+    val puhelinnumero: String,
+) {
+    fun toDomain(): HankekayttajaInput =
```
pitkni, line 33: toDomain may be a bit misleading. HankeKayttaja seems to be the actual domain object. Would it make sense to remove HankekayttajaInput and use this class instead (they hold the same data) in hankekayttajaservice?

corvidian: `HankekayttajaInput` is a domain object that's used in many places and this is a DTO, so tightly coupling them might lead to problems later. I'll rename this to `toHankekayttajaInput` to be explicit.

pitkni (approved):

PR #569, opened 2024-01-09
HAI-1983 Add application attachments to blob storage instead of database

umeetiusbaar:
# Description

Application attachments are stored in Azure blob storage instead of database.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. Add an attachment for an application
2. Check blob storage with e.g. MS Azure Storage Explorer - the attachment should be there under `haitaton-hakemusliitteet-local` in path `<applicationId>/<uuid>`
3. Check the database - there should not be a new row in `application_attachment_content` table
4. Check the database - there should be a new row in `application_attachment` table with `blob_location` set to `<applicationId>/<uuid>` which is the path of the file in blob storage.
5. [OPTIONAL] Check that deletion of the attachment works as expected

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentServiceITest.kt
@@ -74,7 +74,7 @@ class ApplicationAttachmentContentServiceITest(
             assertThat(fileClient.listBlobs(HAKEMUS_LIITTEET).map { it.path })
                 .containsExactly(attachment.blobLocation!!)
 
-            attachmentContentService.delete(attachment)
+            attachmentContentService.delete(attachment.blobLocation!!)
 
             assertThat(fileClient.listBlobs(HAKEMUS_LIITTEET)).isEmpty()
         }
@@ -83,7 +83,8 @@ class ApplicationAttachmentContentServiceITest(
         fun `Doesn't throw an error if content is missing`() {
             val attachment = attachmentWithoutCloudContent().copy(blobLocation = "missing")
 
-            assertThat(runCatching { attachmentContentService.delete(attachment) }).isSuccess()
+            assertThat(runCatching { attachmentContentService.delete(attachment.blobLocation!!) })
+                .isSuccess()
         }
 
         @Test
@@ -91,7 +92,8 @@ class ApplicationAttachmentContentServiceITest(
             val attachment = attachmentWithoutCloudContent()
             assertThat(attachment.blobLocation).isNull()
 
-            assertThat(runCatching { attachmentContentService.delete(attachment) }).isSuccess()
+            assertThat(runCatching { attachmentContentService.delete(attachment.blobLocation) })
+                .isSuccess()
         }
     }
 
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentServiceITest.kt
@@ -8,8 +8,8 @@ import assertk.assertions.hasClass
 import assertk.assertions.hasMessage
 import assertk.assertions.isEmpty
 import assertk.assertions.isEqualTo
-import assertk.assertions.isNull
-import assertk.assertions.isSuccess
+import assertk.assertions.isFalse
+import assertk.assertions.isTrue
 import fi.hel.haitaton.hanke.DatabaseTest
 import fi.hel.haitaton.hanke.attachment.USERNAME
 import fi.hel.haitaton.hanke.attachment.azure.Container.HAKEMUS_LIITTEET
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentServiceITest.kt
```diff
@@ -83,15 +83,17 @@ class ApplicationAttachmentContentServiceITest(
         fun `Doesn't throw an error if content is missing`() {
             val attachment = attachmentWithoutCloudContent().copy(blobLocation = "missing")
 
-            assertThat(runCatching { attachmentContentService.delete(attachment) }).isSuccess()
+            assertThat(runCatching { attachmentContentService.delete(attachment.blobLocation!!) })
+                .isSuccess()
```
corvidian, line 87: This seems overly complicated, the test will fail anyway if there's an unexpected exception.
```suggestion
            attachmentContentService.delete(attachment.blobLocation!!)
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentServiceITest.kt
```diff
@@ -221,11 +224,13 @@ class ApplicationAttachmentServiceITest(
             assertThat(attachmentInDb.createdAt).isRecent()
             assertThat(attachmentInDb.applicationId).isEqualTo(application.id)
             assertThat(attachmentInDb.attachmentType).isEqualTo(typeInput)
+            assertThat(attachmentInDb.blobLocation!!.startsWith("${application.id!!}/")).isTrue()
```
corvidian, line 227: There's a more specific assertion for this, which gives a better error message if it fails.
```suggestion
            assertThat(attachmentInDb.blobLocation).isNotNull().startsWith("${application.id!!}/")
```
This gives an error like `expected to start with:<"1/"> but was:<"2/e71729a3-3f11-4744-848c-548b5fb99c3d">` instead of the generic `expected to be true`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentService.kt
```diff
@@ -19,18 +19,26 @@ class ApplicationAttachmentContentService(
...
-        with(attachment) {
-            logger.info { "Deleting attachment $id if blob defined. Blob=$blobLocation." }
-            blobLocation?.let { fileClient.delete(Container.HAKEMUS_LIITTEET, it) }
+    fun delete(blobPath: String?) {
+        if (blobPath == null) {
+            logger.info { "Blob path is null, nothing to delete" }
+            return
         }
+        fileClient.delete(Container.HAKEMUS_LIITTEET, blobPath)
+    }
```
corvidian, line 39: I think using `?.let` and the Elvis operator gives us cleaner code here.
```suggestion
    fun delete(blobPath: String?) {
        blobPath?.let { fileClient.delete(Container.HAKEMUS_LIITTEET, it) }
            ?: logger.info { "Blob path is null, nothing to delete" }
    }
```

If the log row isn't mandatory, this could be almost a oneliner:
```
    fun delete(blobPath: String?) {
        blobPath?.let { fileClient.delete(Container.HAKEMUS_LIITTEET, blobPath) }
    }
```

Then again, this is called from 3 places, and the blob location can be null in only one of those. So the null case could be also checked at the callsite.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentService.kt
```diff
@@ -19,18 +19,26 @@ class ApplicationAttachmentContentService(
...
 
     fun deleteAllForApplication(applicationId: Long) {
-        logger.info { "Deleting all attachments from application $applicationId" }
```
corvidian, line 33: Was there something wrong with this? Why was it removed?

umeetiusbaar: This is called from only one place in `ApplicationAttachmentService` and there is logging there:
```
    fun deleteAllAttachments(application: Application) {
        attachmentContentService.deleteAllForApplication(application.id!!)
        metadataService.deleteAllAttachments(application.id)
        logger.info { "Deleted all attachments from application ${application.id}" }
    }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentService.kt
umeetiusbaar: Made blobPath non-nullable and added a check in that one place where it is needed (`ApplicationAttachmentService.deleteAttachmend(attachmentId: UUID)`):
```
        attachment.blobLocation?.let { attachmentContentService.delete(it) }
            ?: logger.warn { "Attachment ${attachment.id} has no blob content" }
```
Not sure whether it should log at WARN level or INFO.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentServiceITest.kt
umeetiusbaar: Adjusted the test to test whether anything is actually deleted or not.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentService.kt
corvidian: That logging happens after the operation. It's also useful to log before doing something, especially when that something is related to external services. So if there's a connection error or something, the logs have some context on what the service was trying to do when the error happened.

umeetiusbaar: On lower lever at `BlobFileClient.deleteAllByPrefix(..)` this is already logged:
```
    override fun deleteAllByPrefix(container: Container, prefix: String) {
        logger.info { "Deleting all Blobs from container $container with prefix $prefix" }
        val containerClient = getContainerClient(container)
        val blobUrls =
            containerClient.listBlobsByHierarchy(prefix).map {
                if (it.isPrefix) {
                    throw UnexpectedSubdirectoryException(container, it.name)
                }
                containerClient.getBlobClient(it.name).blobUrl
            }
        if (blobUrls.isEmpty()) {
            return
        }
        batchClient.deleteBlobs(blobUrls, DeleteSnapshotsOptionType.INCLUDE).forEach {
            logger.info { "Deleted blob ${it.request.url} with status ${it.statusCode}" }
        }
        logger.info { "Deleted all Blobs from container $container with prefix $prefix" }
    }
```

I have it little difficult to pinpoint the best location for logging but I did some adjustments here and  there.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentService.kt
umeetiusbaar: Ok, changed it to INFO

corvidian (approved):

PR #570, opened 2024-01-11
HAI-2184 Add size to attachment metadata

umeetiusbaar:
# Description

Add attachment size (in bytes) in database and in attachment metadata API. Migrate existing application attachments to have correct size based on the content size in database. For Hanke attachments the size will be 0 because their content are already migrated to use blob storage.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. For testing existing application attachment add a new attachment BEFORE running this version
2. After starting this version check the database that the new columns `application_attachment.size` and `hanke_attachment.size` are in place and that `application_attachment.size` has correct value (if you had existing application attachment before running this version).
3. Check that API returns `size` and `contentType` for both `GET /hankkeet/{hankeTunnus}/liitteet` and `GET /hakemukset/{hakemusId}/liitteet`

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/resources/db/changelog/changesets/065-add-attachment-sizes.sql
```diff
@@ -0,0 +1,23 @@
...
+
+UPDATE application_attachment
+SET size = length(content)
+FROM application_attachment_content
+WHERE application_attachment.id = application_attachment_content.attachment_id;
+
+
+-- Hanke attachment size
+ALTER TABLE hanke_attachment
+    ADD COLUMN size BIGINT default 0;
```
corvidian, line 19: The default values could be dropped after the existing data has been initialized.

corvidian: Should these be NOT NULL?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentTestUtil.kt
```diff
@@ -26,6 +26,7 @@ val DUMMY_DATA = "ABC".toByteArray()
 val DEFAULT_DATA by lazy {
     "/fi/hel/haitaton/hanke/decision/fake-decision.pdf".getResourceAsBytes()
 }
+const val DEFAULT_SIZE = 18959L
```
corvidian, line 29: Could this be `DEFAULT_DATA.len()`? To explain, where the magic number comes from.

umeetiusbaar: Now that I moved fake-decision.pdf under test (was under itest) this could be done.  Before moving it there were errors in unit tests that used DEFAULT_SIZE (mainly attachment Factory companion object initialization).

services/hanke-service/src/main/resources/db/changelog/changesets/065-add-attachment-sizes.sql
umeetiusbaar: `content_type` column seems to have a default value `default 'application/octet-stream'::text` - should that be removed as well? I now dropped default for size only.

I added NOT NULL for `size`

corvidian: Dropping the default value from `content_type` columns wouldn't hurt. Not mandatory, though.

corvidian (approved):

PR #575, opened 2024-01-15
HAI-2162 Add link between Hankeyhteystieto and Hankekayttaja

corvidian:
# Description

Create a link table between Hankeyhteystieto and Hankekayttaja: Hankeyhteyshenkilo. Add the new relations to the entities.

Remove alikontaktit from yhteystiedot as they will be replaced by the new relations in later PRs. Also remove the logic of creating hankekayttaja from alikontaktit.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Can't really be tested yet, at least not easily.

# Checklist:

 or other location:


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
```diff
@@ -166,17 +161,19 @@ class HankeKayttajaServiceITest : DatabaseTest() {
...
-                hankeKayttajaService.getKayttajatByHankeId(hankeToFind.id)
+                hankeKayttajaService.getKayttajatByHankeId(hankeToFind)
 
-            assertThat(result).hasSize(5)
-            val correctIds = hankeKayttajaRepository.findByHankeId(hankeToFind.id).map { it.id }
-            assertThat(result).each { dto ->
-                dto.prop(HankeKayttajaDto::id).isIn(*correctIds.toTypedArray())
-            }
+            assertThat(result).hasSize(3)
+            val correctIds = hankeKayttajaRepository.findByHankeId(hankeToFind).map { it.id }
```
umeetiusbaar, line 175: This makes me wonder whether it would be better to have the correct ids taken from factory methods (if that is possible). If the service uses this same repository finder method.

services/hanke-service/src/main/resources/db/changelog/changesets/067-create-hankeyhteyshenkilo-table.sql
```diff
@@ -0,0 +1,31 @@
...
+            REFERENCES hankeyhteystieto (id) ON DELETE CASCADE
+);
+
+CREATE UNIQUE INDEX uk_hankekayttaja_hankeyhteystieto ON hankeyhteyshenkilo (hankekayttaja_id, hankeyhteystieto_id);
+CREATE INDEX idx_hankekayttaja ON hankeyhteyshenkilo (hankekayttaja_id);
+CREATE INDEX idx_hankeyhteystieto ON hankeyhteyshenkilo (hankeyhteystieto_id);
+
+COMMENT ON TABLE hankeyhteyshenkilo IS 'Table for linking the hanke users and hanke customers as contact persons.';
+COMMENT ON COLUMN hankeyhteyshenkilo.hankekayttaja_id IS 'The kayttaja who this contact is.';
+COMMENT ON COLUMN hankeyhteyshenkilo.hankeyhteystieto_id IS 'The customer this contact is represents.';
```
umeetiusbaar, line 25: "this contact represents" sounds better?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeYhteystietoEntity.kt
```diff
@@ -136,6 +141,7 @@ class HankeYhteystietoEntity(
 
 @Schema(description = "Contact person")
 data class Yhteyshenkilo(
```
umeetiusbaar, line 143: Should this class be in separate file or in `HankeYhteyshenkiloEntity.kt`? Also, method `toHankekayttajaInput()` seems to be unnecessary.

services/hanke-service/src/main/resources/db/changelog/changesets/067-create-hankeyhteyshenkilo-table.sql
corvidian: Yes, that's a typo. Thank you for noticing.

umeetiusbaar:
Should something be done about code coverage complains?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
corvidian: The problem is the founder since we can't get their ID directly. I added a database fetch to get it before creating the other users, so we know it's the founder. I think the test is better now. Thank you for the suggestion.

This also revealed a bug in `HankeBuilder.saveEntity()`. It created duplicates of the hanke.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeYhteystietoEntity.kt
corvidian: I moved this to a separate class under domain-package. I've been thinking about extracting the yhteystieto stuff from HankeService to a new service and creating a new package around that with the entities and domain classes, but I haven't found the right time to do that.

corvidian:
> Should something be done about code coverage complains?

The missing lines are covered in #579, so I don't think we need to react to them here.

umeetiusbaar (approved):

PR #576, opened 2024-01-16
HAI-1991 Move application attachments from db to blob storage

umeetiusbaar:
# Description

Migrates application attachment content from database to Azure Blob Storage. This is done in a scheduled way so that once a minute a single application attachment is handled. When there are no more application attachments that have content in the database, the scheduled function does nothing.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
* before running this version have an application attachment and its content in database (if cannot create via UI, create manually with e.g. `insert into application_attachment_content values (<attachment-id>, pg_read_binary_file('attachment.pdf'))`. The file has to be in `DB_VOLUME` directory). Also, remember to set `application_attachment.blob_location` to `null` for this attachment.
* set env `HAITATON_ATTACHMENT_MIGRATION_APPLICATION_ENABLED` to `true`.
* start this version
* after 10 secs (`HAITATON_ATTACHMENT_MIGRATION_APPLICATION_INITIAL_DELAY`) the process should start with an INFO message `There are un-migrated application attachments left, count = X` where X is the number of un-migrated attachments
* after the migration process has ran there should not be `application_attachment_content` row for the attachment and `application_attachment.blob_location` should be set and point to a file in Azure Blob Storage.
* loading the attachment via UI should work
* after all attachment content has been migrated there should be `No application attachments to migrate` INFO message in logs when the scheduled task is triggered.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -33,7 +35,54 @@ class ApplicationAttachmentService(
...
 ) {
+
+    internal val lockName = "applicationAttachmentBlobTransfer"
+
+    @Scheduled(
+        fixedDelayString = "\${haitaton.application.attachments.blobTransferIntervalMilliSeconds}",
+        initialDelayString =
+            "\${haitaton.application.attachments.blobTransferInitialDelayMilliSeconds}"
+    )
+    fun transferAttachmentContentToBlobStorage() {
```
corvidian, line 49: I think this should be in a separate service, conditional on an environment variable so we can turn the scheduled method off when it's completed.

Maybe similar to the hanke attachment migration scheduler: https://github.com/City-of-Helsinki/haitaton-backend/commit/40dcc7dac5e54c4a8086722c97c07debcb1bbd44

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
```diff
@@ -33,7 +35,54 @@ class ApplicationAttachmentService(
...
+            "Trying to obtain lock $lockName to start transferring application attachment content from database to Azure Blob Storage."
+        )
+        lockService.doIfUnlocked(lockName) { transferAttachmentContent() }
+    }
+
+    private fun transferAttachmentContent() {
+        logger.info {
+            "Starting to transfer application attachment content from database to Azure Blob Storage"
+        }
+        contentTransferService.nextTransferableAttachment()?.let { attachmentEntity ->
```
corvidian, line 60: This would be easier to read if everything inside the let was in a function, so this would be:
```kotlin
        contentTransferService.nextTransferableAttachment()?.let { migrate(it) }
            ?: logger.info { "All application attachment content has been migrated." }
```
Having the Elvis operator after a long block makes it harder to understand what the null condition refers to.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentEntity.kt
```diff
@@ -170,6 +170,8 @@ interface HankeAttachmentRepository : JpaRepository<HankeAttachmentEntity, UUID>
 interface ApplicationAttachmentRepository : JpaRepository<ApplicationAttachmentEntity, UUID> {
     fun findByApplicationId(applicationId: Long): List<ApplicationAttachmentEntity>
 
+    fun findByBlobLocationIsNull(): List<ApplicationAttachmentEntity>
```
corvidian, line 173:
```suggestion
    fun findFirstByBlobLocationIsNull(): ApplicationAttachmentEntity?
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentTransferService.kt
```diff
@@ -0,0 +1,55 @@
...
+private val logger = KotlinLogging.logger {}
+
+@Service
+class ApplicationAttachmentContentTransferService(
+    private val attachmentRepository: ApplicationAttachmentRepository,
+    private val attachmentContentRepository: ApplicationAttachmentContentRepository,
+    private val fileClient: FileClient,
+) {
+    fun nextTransferableAttachment(): ApplicationAttachmentEntity? =
+        attachmentRepository.findByBlobLocationIsNull().firstOrNull()
```
corvidian, line 25: Passing entities beyond transaction boundaries can lead to difficult bugs. I think this might be safe, but I think the code would work just as well if this method returned the domain object.
```suggestion
    fun nextTransferableAttachment(): ApplicationAttachmentMetadata? =
        attachmentRepository.findFirstByBlobLocationIsNull()?.toDomain()
```

corvidian: The difference would be that in the end in `cleanUpDatabase`, you would need to load the entity from the database, instead of an explicit save:
```kotlin
        val attachmentEntity = attachmentRepository.getReferenceById(attachment.id)
        logger.info { "Updating attachment ${attachmentEntity.id} with blob location $blobPath" }
        attachmentEntity.blobLocation = blobPath
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentTransferService.kt
```diff
@@ -0,0 +1,55 @@
...
+        return blobPath
+    }
+
+    @Transactional
+    fun cleanUpDatabase(attachmentEntity: ApplicationAttachmentEntity, blobPath: String) {
+        logger.info { "Cleaning up database for attachment ${attachmentEntity.id}" }
+        attachmentEntity.blobLocation = blobPath
+        logger.info { "Updating attachment ${attachmentEntity.id} with blob location $blobPath" }
+        attachmentRepository.save(attachmentEntity)
+        logger.info { "Deleting attachment ${attachmentEntity.id} content from database" }
```
corvidian, line 51: I think it's a good practice to separate the variables from the message when logging. So that searching for the event is easier. E.g.
```suggestion
        logger.info { "Deleting attachment content from database, id = ${attachmentEntity.id}" }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
umeetiusbaar: Oh, I wish I had thought to look into that Hanke attachment migration. But, I will do it now.

umeetiusbaar: I refactored the whole thing to be more like what was done with Hanke attachments.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentService.kt
umeetiusbaar: Thank you for pointing this out. I hope there are no such things any more.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/AttachmentEntity.kt
umeetiusbaar: Thank you for this, I did not know about `First`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentTransferService.kt
umeetiusbaar: You are of course right. I haven't been working with JPA and Entities for while and it shows in things like this. After refactoring to follow Hanke attachment migration I hope this kind of errors are no more.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentContentTransferService.kt
umeetiusbaar: Changed logging to follow this kind of syntax.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
```diff
@@ -0,0 +1,109 @@
...
+                    applicationId = attachmentEntity.applicationId,
+                    content =
+                        AttachmentContent(
+                            fileName = attachmentEntity.fileName,
+                            contentType = attachmentEntity.contentType,
+                            bytes = contentEntity.content,
+                        ),
+                )
+            }
+        }
```
corvidian, line 50: There's some unnecessary nesting here.
```suggestion
        val amount = attachmentAmountToMigrate()
        if (amount == 0L) {
            return null
        }
        logger.info {
            logger.info { "There are un-migrated application attachments left, count = $amount" }
        }
        val (attachmentEntity, contentEntity) = pickForMigration()
        return ApplicationAttachmentWithContent(
            id = attachmentEntity.id!!,
            applicationId = attachmentEntity.applicationId,
            content =
                AttachmentContent(
                    fileName = attachmentEntity.fileName,
                    contentType = attachmentEntity.contentType,
                    bytes = contentEntity.content,
                ),
        )
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
```diff
@@ -0,0 +1,109 @@
...
+    /** Find an attachment that is not yet migrated, i.e. still has content in the db. */
+    @Transactional(readOnly = true)
+    fun findAttachmentWithDatabaseContent(): ApplicationAttachmentWithContent? {
+        return attachmentAmountToMigrate().let {
+            if (it == 0L) {
+                return null
+            }
+            logger.info {
+                logger.info { "There are un-migrated application attachments left, count = $it" }
+            }
```
corvidian, line 37: I assume these nested `logger.info`s are a copy-paste error or similar.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    private fun attachmentAmountToMigrate() =
+        contentRepository.count().also {
+            logger.info { "There are un-migrated application attachments left, count = $it" }
+        }
+
+    /**
+     * An attachment with its content. Content references attachment data. Thus, the relation must
+     * exist.
+     */
```
corvidian, line 92: This comment is not valid for this implementation.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/ApplicationAttachmentMigrationSchedulerTest.kt
```diff
@@ -0,0 +1,209 @@
...
+            every { migrator.findAttachmentWithDatabaseContent() } returns null
+
+            scheduler.scheduleMigrate()
+
+            verifyOrder {
+                jdbcLockRegistry.obtain(LOCK_NAME)
+                migrator.findAttachmentWithDatabaseContent()
+            }
+            verify(exactly = 0) { migrator.migrate(any()) }
+            verify(exactly = 0) { migrator.setBlobPathAndCleanup(any(), any()) }
```
corvidian, line 67: VerifySequence checks that no other calls were made to the mocks mentioned in the block.
```suggestion
            verifySequence {
                jdbcLockRegistry.obtain(LOCK_NAME)
                migrator.findAttachmentWithDatabaseContent()
            }
```
This also checks the order of the calls, even though it's not documented directly in the function's KDoc.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
umeetiusbaar: Fixed as suggested.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
umeetiusbaar: Something like that maybe. Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
umeetiusbaar: Fixed.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/ApplicationAttachmentMigrationSchedulerTest.kt
umeetiusbaar: Thank you for teaching me a new thing again! Fixed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
```diff
@@ -0,0 +1,146 @@
...
+            val attachmentWithContent = attachmentMigrator.findAttachmentWithDatabaseContent()!!
+
+            val blobPath = attachmentMigrator.migrate(attachmentWithContent)
+
+            assertThat(attachmentRepository.findAll()).hasSize(1)
+            assertThat(contentRepository.findAll()).hasSize(1)
+            assertThat(fileClient.listBlobs(Container.HAKEMUS_LIITTEET)).hasSize(1)
+            assertThat(fileClient.listBlobs(Container.HAKEMUS_LIITTEET).first()).all {
+                prop(TestFile::path).isEqualTo(blobPath)
+            }
```
corvidian, line 109: `all {}` isn't necessary with just one assertation.
```suggestion
            assertThat(fileClient.listBlobs(Container.HAKEMUS_LIITTEET).first())
                .prop(TestFile::path)
                .isEqualTo(blobPath)
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
```diff
@@ -0,0 +1,146 @@
...
+            val attachmentWithContent = attachmentMigrator.findAttachmentWithDatabaseContent()
+
+            assertThat(attachmentWithContent).isNotNull().all {
+                prop(ApplicationAttachmentWithContent::id).isEqualTo(attachment.id!!)
+                prop(ApplicationAttachmentWithContent::applicationId)
+                    .isEqualTo(attachment.applicationId)
+                prop(ApplicationAttachmentWithContent::content).all {
+                    prop(AttachmentContent::fileName).isEqualTo(attachment.fileName)
+                    prop(AttachmentContent::contentType).isEqualTo(attachment.contentType)
+                    prop(AttachmentContent::bytes).isEqualTo(content.content)
```
corvidian, line 86: Couldn't this check against the data the factory uses?
```suggestion
                    prop(AttachmentContent::bytes).isEqualTo(DEFAULT_DATA)
```

If we want it to be explicit, we can specify the data the factory should use.
```kotlin
val attachment = attachmentFactory.save().withDbContent(DEFAULT_DATA).value
```

Either way, we should know what the content is supposed to be without getting it from the DB.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    @Nested
+    @Transactional
+    inner class SetBlobPathAndCleanup {
+        @Test
+        fun `Should throw AttachmentNotFoundException if attachment is not in db`() {
+            val attachment = attachmentFactory.save().withDbContent().value
+            val blobPath = generateBlobPath(attachment.applicationId)
+            attachmentRepository.delete(attachment)
+            assertThat(attachmentRepository.findAll()).isEmpty()
```
corvidian, line 121: Is it necessary to save an attachment and then delete it? It's the same as if it never existed. Except that `attachmentFactory.save()` saves an application, but `setBlobPathAndCleanup` doesn't use the application for anything.
```suggestion
            val attachmentId = UUID.fromString("1695d310-0602-4822-9770-ff76847a4479")
            val blobPath = "15/$attachmentId"
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
```diff
@@ -0,0 +1,146 @@
...
+                attachmentMigrator.setBlobPathAndCleanup(attachment.id!!, blobPath)
+            }
+
+            assertThat(attachmentRepository.findAll()).isEmpty()
+        }
+
+        @Test
+        fun `Should set blob path and delete content from db`() {
+            val attachment = attachmentFactory.save().withDbContent().value
+            val blobPath = generateBlobPath(attachment.applicationId)
```
corvidian, line 132: I think it would be better to be explicit with blobPath here. So we know `setBlobPathAndCleanup` uses the parameter and doesn't compute it independently.
```suggestion
            val blobPath = "some/path"
```

services/hanke-service/src/main/resources/application.yml
```diff
@@ -7,6 +7,11 @@ haitaton:
     updateInitialDelayMilliSeconds: ${ALLU_UPDATE_INITIAL_DELAY:60000}
     updateIntervalMilliSeconds: ${ALLU_UPDATE_INTERVAL:60000}
     username: ${ALLU_USERNAME:fake_user}
+  attachment-migration:
+    application:
+      enabled: ${HAITATON_ATTACHMENT_MIGRATION_APPLICATION_ENABLED:false}
+      interval: ${HAITATON_ATTACHMENT_MIGRATION_APPLICATION_INTERVAL:60000}
+      initial-delay: ${HAITATON_ATTACHMENT_MIGRATION_APPLICATION_INITIAL_DELAY:10000}
```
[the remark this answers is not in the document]
umeetiusbaar: I used the same value from Hanke attachment migration. I increased it to 60000 as it was in my original implementation.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
umeetiusbaar: Fixed

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
umeetiusbaar: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
```diff
@@ -0,0 +1,102 @@
...
+    private val contentRepository: ApplicationAttachmentContentRepository,
+) {
+    /** Find an attachment that is not yet migrated, i.e. still has content in the db. */
+    @Transactional(readOnly = true)
+    fun findAttachmentWithDatabaseContent(): ApplicationAttachmentWithContent? {
+        val amount = attachmentAmountToMigrate()
+        if (amount == 0L) {
+            return null
+        }
+        logger.info { "There are un-migrated application attachments left, count = $amount" }
```
corvidian, line 35: This already logged inside `attachmentAmountToMigrate()`.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
umeetiusbaar: Fixed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigratorITest.kt
umeetiusbaar: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/application/ApplicationAttachmentMigrator.kt
umeetiusbaar: I removed the whole `attachmentAmountToMigrate()` as its only purpose was to log the number. I do not think that logging 0 is helpful as `No application attachments to migrate` is logged in `ApplicationAttachmentMigrationScheduler`.

corvidian (approved):

PR #600, opened 2024-02-01
HAI-2168 Update user's name from Profiili

umeetiusbaar:
# Description

When user logs in via the invitation link, retrieve user's verified name from Profiili and if it differs from original name, update the name. Use given name as firstname. Do not alter names used in invitation.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. Using an access login of logged in Hanke user with `CREATE_USER` rights, create a new user ` POST /hankkeet/{hankeTunnus}/kayttajat` for Hanke. Take note of the data you provide, e.g. 
`{
  "etunimi": "Jaska",
  "sukunimi": "Jokunen",
  "sahkoposti": "jaska.jokunen@gmail.com",
  "puhelinnumero": "1234567"
}`
2. Log in with the invitation link. Use Suomi.fi test user that is not yet used in Hanke. 
3. Using the access token of this new log in, check that the user name is updated with Profiili data `GET /hankkeet/{hankeTunnus}/kayttajat` (There should also be an INFO log message starting `Updated user's name from Profiili`)

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
@@ -286,7 +286,7 @@ Responds with information about the activated user and the hanke associated with
                 ),
                 ApiResponse(
                     description = "Name not found in Profiili",
-                    responseCode = "404",
+                    responseCode = "500",
                     content = [Content(schema = Schema(implementation = HankeError::class))]
                 ),
                 ApiResponse(
@@ -457,7 +457,7 @@ Returns the updated hankekayttaja.
     }
 
     @ExceptionHandler
-    @ResponseStatus(HttpStatus.NOT_FOUND)
+    @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
     @Hidden
     fun verifiedNameNotFoundException(ex: VerifiedNameNotFound): HankeError {
         logger.warn(ex) { ex.message }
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -411,6 +424,14 @@ of the token and link will be reset.
         return HankeError.HAI4004
     }
 
+    @ExceptionHandler
+    @ResponseStatus(HttpStatus.NOT_FOUND)
```
corvidian, line 428: I don't think this should be a 404. Not found means that the URL for the request had something wrong with it.

Here, we expect every authorized user to have a verified name, so if it's missing it's a problem with Profiili. Maybe this should be a 500 Internal server error?

umeetiusbaar: You're right. I changed it to 500.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -280,6 +284,11 @@ Responds with information about the activated user and the hanke associated with
                     responseCode = "404",
                     content = [Content(schema = Schema(implementation = HankeError::class))]
                 ),
+                ApiResponse(
+                    description = "Name not found in Profiili",
```
corvidian, line 288: The OpenAPI description can have only one response section per error code. Currently, this doesn't show up in the Swagger UI or the api-doc json, as it's being overwritten by the next response. ("Token doesn't have a user associated with it").

This description can be merged with that one.

umeetiusbaar: Good to know, thanks!

corvidian (approved):

PR #608, opened 2024-02-08
POC for maps in pdfs

corvidian:
# Description

I made a POC of adding map images to PDF files. This uses the johtoselvitys-PDF as an example.

Some things, like the map boundaries, are hard coded.

The code has not been cleaned, so the functions are not well divided and there are many extra prints everywhere.


corvidian:
Useful parts of this have now been merged to dev in other PRs.

PR #616, opened 2024-02-15
HAI-2243 New API for getting hakemukset for hanke

umeetiusbaar:
# Description
Move API endpoint `/hankkeet/<hankeid>/hakemukset` from HankeController to new HakemusController.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Endpoint `/hankkeet/<hankeid>/hakemukset` should work as before.

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
@@ -54,7 +54,7 @@ import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
 
 private const val USERNAME = "testUser"
 private const val HANKE_TUNNUS = "HAI-1234"
-private const val HAKEMUKSET_BASE_URL = "/hakemukset"
+private const val BASE_URL = "/hakemukset"
 private const val HANKKEET_BASE_URL = "/hankkeet"
 
 @WebMvcTest(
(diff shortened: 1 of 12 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationController.kt
```diff
@@ -54,7 +52,26 @@ class ApplicationController(
     private val hankeService: HankeService,
     private val disclosureLogService: DisclosureLogService,
 ) {
-    @GetMapping
+    @GetMapping("/hankkeet/{hankeTunnus}/hakemukset")
+    @Operation(
+        summary = "Get hanke applications",
+        description = "Returns list of applications belonging to a given hanke."
+    )
+    @PreAuthorize("@hankeAuthorizer.authorizeHankeTunnus(#hankeTunnus, 'VIEW')")
```
corvidian, line 60: `authorizeHankeTunnus` is a part of the base `Authorizer` class, so it can be called through any authorizer. Using the same authorizer as the other endpoints makes tests simpler.
```suggestion
    @PreAuthorize("@applicationAuthorizer.authorizeHankeTunnus(#hankeTunnus, 'VIEW')")
```

umeetiusbaar: Good to know! I changed it here and elsewhere where I had used both hankeAuthorizer and applicationAuthorizer.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/application/ApplicationControllerITest.kt
```diff
@@ -64,7 +68,7 @@ class ApplicationControllerITest(@Autowired override val mockMvc: MockMvc) : Con
 
     @Autowired private lateinit var applicationService: ApplicationService
     @Autowired private lateinit var hankeService: HankeService
-    @Autowired private lateinit var authorizer: ApplicationAuthorizer
+    @Autowired private lateinit var applicationAuthorizer: ApplicationAuthorizer
```
corvidian, line 71: Could you revert this change? This causes a lot of noise to the history and it's no longer required.

umeetiusbaar: You're right! Reverted the change and also BASE_URL (well, that could have been better to be the new HAKEMUKSET_BASE_URL but maybe later that change)

corvidian (approved):

PR #623, opened 2024-02-20
HAI-2289 gis-validate-deploy process

petri-timlin:
# Description

Added common validation and deploing methods for automation.
Added own handling modules for each "tormays" material.
Added "tormays" data configurations for each material.
Added new gis-validate-deploy container using own haitaton_gis_validate_deploy volume.
Added logging configuration to config.yaml
Updated scripts/gis-material-update/README.md

### Jira Issue: 
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2289 (main ticket)
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2287
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2288
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2291
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2292
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2293
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2294
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2295
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2297
https://helsinkisolutionoffice.atlassian.net/browse/HAI-2298

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

Check instructions from scripts/gis-material-update/readme starting from `Run validate-deploy` chapter. There is also some changes in beginning of the README (Need to have volumes created, images build and copy script files to external volumes).
There need to be at least one of the following "tormays" tables in the testing enviroment:
- tormays_buses_polys
- tormays_central_business_area_polys
- tormays_cycle_infra_polys
- tormays_street_classes_polys
- tormays_tram_infra_polys
- tormays_tram_lines_polys
- tormays_volumes15_polys
- tormays_volumes30_polys
- tormays_ylre_classes_polys
- tormays_ylre_parts_polys	

Check that .env file contains values for:
- HELSINKI_EXTRANET_USERNAME
- HELSINKI_EXTRANET_PASSWORD
- HAITATON_USER
- HAITATON_USER_PW
- HAITATON_DB
- GIS_DATABASE_PORT

Then you can run following:

```sh
docker-compose run --rm gis-fetch <source>
docker-compose run --rm gis-process <source>
docker-compose run --rm gis-validate-deploy <source>
```

where `<source>` is currently one of:

- `hsl`
- `ylre_katualueet`
- `ylre_katuosat`
- `maka_autoliikennemaarat`
- `tram_infra`  
- `tram_lines`
- `cycle_infra`
- `central_business_area`
- `liikennevaylat`

and each source is having it's own prerequisites for gis-fetch (check `Run processing` chapter).
Use that `<source>` which is having "tormays" table in the testing environment.

After run results should be something like below where the example is from `central_business_area`:
```sh
Validating data.
Validating deployment item: central_business_area
2024-02-13 11:11:28,613 - INFO - Data amount validation started
2024-02-13 11:11:28,614 - INFO - Module: central_business_area
2024-02-13 11:11:28,614 - INFO - New data amount (tormays_central_business_area_polys_temp): 123
2024-02-13 11:11:28,614 - INFO - Old data amount (tormays_central_business_area_polys): 123
2024-02-13 11:11:28,614 - INFO - Limits: 120 <= new amount <= 136
2024-02-13 11:11:28,614 - INFO - Data amount validation result: Valid
2024-02-13 11:11:28,623 - INFO - Deploy ...
2024-02-13 11:11:28,624 - INFO - tormays_central_business_area_polys: 123 rows deleted.
2024-02-13 11:11:28,628 - INFO - tormays_central_business_area_polys: 123 rows inserted.
```

# Checklist:

 or other location: 
https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8292401164/Paikkatietoaineiston+k+sittely#Paikkatietoaineiston-automaatio-(kontitettu-ymp%C3%A4rist%C3%B6)

# Other relevant info

### DIFF
--- a/scripts/gis-material-update/process/modules/autoliikennemaarat.py
+++ b/scripts/gis-material-update/process/modules/autoliikennemaarat.py
@@ -13,8 +13,10 @@ class MakaAutoliikennemaarat:
         self._module = "maka_autoliikennemaarat"
 
         filename = cfg.local_file(self._module)
+        self._store_original_data = cfg.store_orinal_data(self._module)
         layer = cfg.layer(self._module)
         df = gpd.read_file(filename, layer=layer)
+        self._orig = df
         self._df = (
             df.loc[:, ["autot", "geometry"]]
             .rename(columns={"autot": "volume"})
@@ -37,37 +39,17 @@ class MakaAutoliikennemaarat:
     def persist_to_database(self) -> None:
         connection = create_engine(self._cfg.pg_conn_uri())
 
-        # persist traffic volume lines to database
-        self._df.to_postgis(
-            "volume_lines",
-            connection,
-            "public",
-            if_exists="replace",
-            index=True,
-            index_label="fid",
-        )
-
-        # persist traffic volume polygons to database
-        #for buffer_size, polygon_data in self._process_result.items():
-        #    polygon_data.to_postgis(
-        #        "volume_{}".format(buffer_size),
-        #        connection,
-        #        "public",
-        #        if_exists="replace",
-        #        index=True,
-        #        index_label="fid",
-        #    )
-
-        # persist results to temp table
-        for buffer_size, polygon_data in self._process_result.items():
-            polygon_data.to_postgis(
-                self._cfg.tormays_table_temp(self._module).format(buffer_size),
+        if self._store_original_data is not False:
+            self._orig.rename_geometry('geom', inplace=True)
+            # persist original data
+            self._orig.to_postgis(
+                self._store_original_data,
                 connection,
                 "public",
                 if_exists="replace",
                 index=True,
                 index_label="fid",
-            )
+                )
 
     def save_to_file(self):
         target_lines_file_name = self._cfg.target_file(self._module)
--- a/scripts/gis-material-update/process/modules/autoliikennemaarat.py
+++ b/scripts/gis-material-update/process/modules/autoliikennemaarat.py
@@ -48,15 +48,15 @@ class MakaAutoliikennemaarat:
         )
 
         # persist traffic volume polygons to database
-        for buffer_size, polygon_data in self._process_result.items():
-            polygon_data.to_postgis(
-                "volume_{}".format(buffer_size),
-                connection,
-                "public",
-                if_exists="replace",
-                index=True,
-                index_label="fid",
-            )
+        #for buffer_size, polygon_data in self._process_result.items():
+        #    polygon_data.to_postgis(
+        #        "volume_{}".format(buffer_size),
+        #        connection,
+        #        "public",
+        #        if_exists="replace",
+        #        index=True,
+        #        index_label="fid",
+        #    )
 
         # persist results to temp table
         for buffer_size, polygon_data in self._process_result.items():
--- a/scripts/gis-material-update/process/modules/autoliikennemaarat.py
+++ b/scripts/gis-material-update/process/modules/autoliikennemaarat.py
@@ -58,6 +58,17 @@ class MakaAutoliikennemaarat:
                 index_label="fid",
             )
 
+        # persist results to temp table
+        for buffer_size, polygon_data in self._process_result.items():
+            polygon_data.to_postgis(
+                self._cfg.tormays_table_temp(self._module).format(buffer_size),
+                connection,
+                "public",
+                if_exists="replace",
+                index=True,
+                index_label="fid",
+            )
+
     def save_to_file(self):
         target_lines_file_name = self._cfg.target_file(self._module)
         self._df.to_file(target_lines_file_name, driver="GPKG")
--- a/scripts/gis-material-update/process/modules/cycling_infra.py
+++ b/scripts/gis-material-update/process/modules/cycling_infra.py
@@ -17,8 +17,10 @@ class CycleInfra(GisProcessor):
         self._module = "cycle_infra"
 
         file_name = cfg.local_file(self._module)
+        self._store_original_data = cfg.store_orinal_data(self._module)
 
         self._lines = gpd.read_file(file_name)
+        self._orig = self._lines
 
     def process(self):
         self._process_result_lines = self._lines
@@ -36,38 +38,20 @@ class CycleInfra(GisProcessor):
         self._process_result_polygons = target_infra_polys
 
     def persist_to_database(self):
-        engine = create_engine(self._cfg.pg_conn_uri(), future=True)
+        connection = create_engine(self._cfg.pg_conn_uri(), future=True)
 
-        # persist route lines to database
-        self._process_result_lines.to_postgis(
-            "cycle_infra",
-            engine,
-            "public",
-            if_exists="replace",
-            index=True,
-            index_label="fid",
-        )
+        if self._store_original_data is not False:
+            self._orig.rename_geometry('geom', inplace=True)
+            # persist original data
+            self._orig.to_postgis(
+                self._store_original_data,
+                connection,
+                "public",
+                if_exists="replace",
+                index=True,
+                index_label="fid",
+                )
 
-        # persist polygons to database
-        #self._process_result_polygons.to_postgis(
-        #    "cycle_infra_polys",
-        #    engine,
-        #    "public",
-        #    if_exists="replace",
-        #    index=True,
-        #    index_label="fid",
-        #)
-
-        # persist results to temp table
-        self._process_result_polygons.to_postgis(
-            self._cfg.tormays_table_temp(self._module),
-            engine,
-            "public",
-            if_exists="replace",
-            index=True,
-            index_label="fid",
-        )
-        
     def save_to_file(self):
         """Save processing results to file."""
         # cycle line infra as debug material
--- a/scripts/gis-material-update/process/modules/cycling_infra.py
+++ b/scripts/gis-material-update/process/modules/cycling_infra.py
@@ -49,14 +49,14 @@ class CycleInfra(GisProcessor):
         )
 
         # persist polygons to database
-        self._process_result_polygons.to_postgis(
-            "cycle_infra_polys",
-            engine,
-            "public",
-            if_exists="replace",
-            index=True,
-            index_label="fid",
-        )
+        #self._process_result_polygons.to_postgis(
+        #    "cycle_infra_polys",
+        #    engine,
+        #    "public",
+        #    if_exists="replace",
+        #    index=True,
+        #    index_label="fid",
+        #)
 
         # persist results to temp table
         self._process_result_polygons.to_postgis(
--- a/scripts/gis-material-update/process/modules/cycling_infra.py
+++ b/scripts/gis-material-update/process/modules/cycling_infra.py
@@ -14,8 +14,9 @@ class CycleInfra(GisProcessor):
         self._process_result_lines = None
         self._process_result_polygons = None
         self._debug_result_lines = None
+        self._module = "cycle_infra"
 
-        file_name = cfg.local_file("cycle_infra")
+        file_name = cfg.local_file(self._module)
 
         self._lines = gpd.read_file(file_name)
 
@@ -23,7 +24,7 @@ class CycleInfra(GisProcessor):
         self._process_result_lines = self._lines
 
         # Buffering configuration
-        buffers = self._cfg.buffer("cycle_infra")
+        buffers = self._cfg.buffer(self._module)
         if len(buffers) != 1:
             raise ValueError("Unknown number of buffer values")
 
@@ -57,17 +58,27 @@ class CycleInfra(GisProcessor):
             index_label="fid",
         )
 
+        # persist results to temp table
+        self._process_result_polygons.to_postgis(
+            self._cfg.tormays_table_temp(self._module),
+            engine,
+            "public",
+            if_exists="replace",
+            index=True,
+            index_label="fid",
+        )
+        
     def save_to_file(self):
         """Save processing results to file."""
         # cycle line infra as debug material
-        target_infra_file_name = self._cfg.target_file("cycle_infra")
+        target_infra_file_name = self._cfg.target_file(self._module)
 
         cycle_lines = self._process_result_lines.reset_index(drop=True)
 
         cycle_lines.to_file(target_infra_file_name, driver="GPKG")
 
         # tormays GIS material
-        target_buffer_file_name = self._cfg.target_buffer_file("cycle_infra")
+        target_buffer_file_name = self._cfg.target_buffer_file(self._module)
 
         # instruct Geopandas for correct data type in file write
         # fid is originally as index, obtain fid as column...
--- a/scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
+++ b/scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
@@ -1,0 +1,52 @@
+import logging
+
+from modules.config import Config
+from modules.common import *
+from modules.gis_validate_deploy import GisProcessor
+
+class MakaAutoliikennemaarat(GisProcessor):
+    """Process traffic (car) volumes."""
+
+    def __init__(self, cfg: Config):
+        self._cfg = cfg
+        self._module = "maka_autoliikennemaarat"
+        self._tormays_table_org = self._cfg.tormays_table_org(self._module)
+        self._tormays_table_temp = self._cfg.tormays_table_temp(self._module)
+        self._validate_limit_min = self._cfg.validate_limit_min(self._module)
+        self._validate_limit_max = self._cfg.validate_limit_max(self._module)
+        self._pg_conn_uri = self._cfg.pg_conn_uri()
+        logging.basicConfig(
+            filename=self._cfg.logging_filename().format(self._module), 
+            filemode=self._cfg.logging_filemode(), 
+            format=self._cfg.logging_format()
+            )
+        self._logger = logging.getLogger()
+        self._logger.setLevel(self._cfg.logging_level())
+                            
+    def validate_deploy(self):
+
+        # validate data amount: is it between given limits
+        buffers = self._cfg.buffer(self._module)
+        self._validate_result = {}
+        for buffer in buffers:
+            self._validate_result[buffer] = validate_data_count_limits(
+                self._module,
+                self._pg_conn_uri, 
+                self._tormays_table_org.format(buffer), 
+                self._tormays_table_temp.format(buffer), 
+                self._validate_limit_min, 
+                self._validate_limit_max, 
+                self._logger
+                )
+        
+        # Valid --> deploy
+        for buffer_size, result in self._validate_result.items():
+            if result == "Valid":
+                self._deploy_result = deploy(
+                    self._pg_conn_uri, 
+                    self._tormays_table_org.format(buffer_size), 
+                    self._tormays_table_temp.format(buffer_size), 
+                    self._logger
+                    )
+            else:
+                logging.error("Validation failed(" + self._module + "_" + str(buffer_size) + "). No deploy")
--- a/scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
+++ b/scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
@@ -11,11 +11,11 @@ class MakaAutoliikennemaarat(GisProcessor):
     def __init__(self, cfg: Config):
         self._module = "maka_autoliikennemaarat"
         self._buffers = cfg.buffer(self._module)
+        self._filename = cfg.target_buffer_file(self._module)
+        self._tormays_files_temp = {}
         GisProcessor.__init__(self, cfg)
 
-    def get_temp_data(self, cfg):
-        self._tormays_files_temp = {}
-        self._filename = cfg.target_buffer_file(self._module)
+    def get_temp_data(self):
         for buffer in self._buffers:
             buffer_data = gpd.read_file(self._filename.format(buffer))
             buffer_data.rename_geometry('geom', inplace=True)
@@ -24,11 +24,10 @@ class MakaAutoliikennemaarat(GisProcessor):
     def validate_deploy(self):
 
         # validate data amount: is it between given limits
-        buffers = self._buffers
-        self._validate_result = False
+        validate_result = False
 
-        for buffer in buffers:
-            self._validate_result = validate_data_count_limits(
+        for buffer in self._buffers:
+            validate_result = validate_data_count_limits(
                 self._module,
                 self._pg_conn_uri,
                 self._tormays_table_org.format(buffer),
@@ -38,7 +37,7 @@ class MakaAutoliikennemaarat(GisProcessor):
                 self._filename.format(buffer),
                 self._logger
                 )
-            if self._validate_result is True:
+            if validate_result is True:
                 self._deploy_result = deploy(
                     self._pg_conn_uri,
                     self._tormays_table_org.format(buffer),
--- a/scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
+++ b/scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
@@ -3,50 +3,47 @@ import logging
 from modules.config import Config
 from modules.common import *
 from modules.gis_validate_deploy import GisProcessor
+import geopandas as gpd
 
 class MakaAutoliikennemaarat(GisProcessor):
     """Process traffic (car) volumes."""
 
     def __init__(self, cfg: Config):
-        self._cfg = cfg
         self._module = "maka_autoliikennemaarat"
-        self._tormays_table_org = self._cfg.tormays_table_org(self._module)
-        self._tormays_table_temp = self._cfg.tormays_table_temp(self._module)
-        self._validate_limit_min = self._cfg.validate_limit_min(self._module)
-        self._validate_limit_max = self._cfg.validate_limit_max(self._module)
-        self._pg_conn_uri = self._cfg.pg_conn_uri()
-        logging.basicConfig(
-            filename=self._cfg.logging_filename().format(self._module), 
-            filemode=self._cfg.logging_filemode(), 
-            format=self._cfg.logging_format()
-            )
-        self._logger = logging.getLogger()
-        self._logger.setLevel(self._cfg.logging_level())
-                            
+        self._buffers = cfg.buffer(self._module)
+        GisProcessor.__init__(self, cfg)
+
+    def get_temp_data(self, cfg):
+        self._tormays_files_temp = {}
+        self._filename = cfg.target_buffer_file(self._module)
+        for buffer in self._buffers:
+            buffer_data = gpd.read_file(self._filename.format(buffer))
+            buffer_data.rename_geometry('geom', inplace=True)
+            self._tormays_files_temp[buffer] = buffer_data
+
     def validate_deploy(self):
 
         # validate data amount: is it between given limits
-        buffers = self._cfg.buffer(self._module)
-        self._validate_result = {}
+        buffers = self._buffers
+        self._validate_result = False
+
         for buffer in buffers:
-            self._validate_result[buffer] = validate_data_count_limits(
+            self._validate_result = validate_data_count_limits(
                 self._module,
-                self._pg_conn_uri, 
-                self._tormays_table_org.format(buffer), 
-                self._tormays_table_temp.format(buffer), 
-                self._validate_limit_min, 
-                self._validate_limit_max, 
+                self._pg_conn_uri,
+                self._tormays_table_org.format(buffer),
+                self._tormays_files_temp[buffer],
+                self._validate_limit_min,
+                self._validate_limit_max,
+                self._filename.format(buffer),
                 self._logger
                 )
-        
-        # Valid --> deploy
-        for buffer_size, result in self._validate_result.items():
-            if result == "Valid":
+            if self._validate_result is True:
                 self._deploy_result = deploy(
-                    self._pg_conn_uri, 
-                    self._tormays_table_org.format(buffer_size), 
-                    self._tormays_table_temp.format(buffer_size), 
+                    self._pg_conn_uri,
+                    self._tormays_table_org.format(buffer),
+                    self._tormays_files_temp[buffer],
                     self._logger
                     )
             else:
-                logging.error("Validation failed(" + self._module + "_" + str(buffer_size) + "). No deploy")
+                logging.error("Validation failed(" + self._module + "_" + str(buffer) + "). No deploy")
--- a/scripts/gis-material-update/validate-deploy/modules/common.py
+++ b/scripts/gis-material-update/validate-deploy/modules/common.py
@@ -10,7 +10,6 @@ def deploy(pg_conn_uri, tormays_table_org, tormays_table_temp, logger=None):
         delete_sql = "delete from " + tormays_table_org
         try:
             logger.info("Deploy ...")
-            # Future development: Add JDBC Lock Registry functionality by inserting line in the table "int_lock" before actual deleting transaction and removing lock line after insert transaction.
             delete_result = connection.execute(text(delete_sql))
             logger.info(tormays_table_org + ": " + f"{delete_result.rowcount} rows deleted.")
         except Exception as e:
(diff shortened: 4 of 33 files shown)


scripts/gis-material-update/process/modules/autoliikennemaarat.py
```diff
@@ -48,9 +48,20 @@ def persist_to_database(self) -> None:
         )
 
         # persist traffic volume polygons to database
+        #for buffer_size, polygon_data in self._process_result.items():
```
corvidian, line 51: Why has a bunch of out-commented code been added to the process scripts?

scripts/gis-material-update/validate-deploy/modules/hsl.py
```diff
@@ -0,0 +1,48 @@
+import logging
+
+from modules.config import Config
+from modules.common import *
+from modules.gis_validate_deploy import GisProcessor
+
+class HslBuses(GisProcessor):
```
corvidian, line 7: The only differences between e.g. this file and `liikennevaylat.py` are the class name, the docstring under it and the module name. Having several copies of the same code makes maintenance harder since any change has to be applied to every copy of the code or they will start diverging.

Is it possible to add the common code to the abstract base class?

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+    old_amount = None
+    new_amount = None
+    tormays_sql = "select count(*) lkm from " + tormays_table_org + " where geometry is not null"
+    temp_sql = "select count(*) lkm from " + tormays_table_temp + " where geometry is not null"
+    engine = create_engine(pg_conn_uri)
+    with engine.connect() as connection:
+        try:
+            tormays_result = connection.execute(text(tormays_sql)).fetchall()
+        except Exception as e:
+            # Transaction implicitly rolls back if an exception occurred within the "with" block
```
corvidian, line 43: There's no transaction here.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+        max_limit = "NA"
+        
+    logger.info("Limits: " + str(low_limit) + " <= new amount <= " + str(max_limit))
+    
+    if retval == "Valid":
+        logger.info("Data amount validation result: " + retval)
+    else:
+        logger.error("Data amount validation result: " + retval)
+        
+    return retval
```
corvidian, line 91: The error is already logged, so this function could return a simple boolean.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+    temp_sql = "select count(*) lkm from " + tormays_table_temp + " where geometry is not null"
+    engine = create_engine(pg_conn_uri)
+    with engine.connect() as connection:
+        try:
+            tormays_result = connection.execute(text(tormays_sql)).fetchall()
+        except Exception as e:
+            # Transaction implicitly rolls back if an exception occurred within the "with" block
+            logger.error(f"Error: {e}")
+
+    with engine.connect() as connection:
```
corvidian, line 46: Getting the `old_amount` and `new_amount` is identical except for the table name. Getting the count could be extracted to a separate function that's called twice. Once to get the `old_amount` and once to get the `new_amount`.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+        elif new_amount<math.floor(validate_limit_min*old_amount):
+            retval = "Data amount is Below given limits"
+        elif new_amount>math.ceil(validate_limit_max*old_amount):
+            retval = "Data amount is Above given limits"
+        else:
+            retval = "Not valid"
+    else:
+        retval = "Not valid"
+
+    logger.info("Data amount validation started")
```
corvidian, line 73: The validation has already been done at this point. Starting the validation could be logged at the top of this function.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+    logger.info("Module: " + module)
+    logger.info("New data amount (" + tormays_table_temp + "): " + str(new_amount))
+    logger.info("Old data amount (" + tormays_table_org + "): " + str(old_amount))
+    if old_amount and validate_limit_min and validate_limit_max:
+        low_limit = math.floor(validate_limit_min*old_amount)
+        max_limit = math.ceil(validate_limit_max*old_amount)
```
corvidian, line 79: Calculating the limits is done repeatedly. These variables could be defined before they are first used.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+    
+    if temp_result:
+        for row in temp_result:
+            new_amount = row.lkm
+    
+    if new_amount and old_amount:
+        if new_amount>=math.floor(validate_limit_min*old_amount) and new_amount<=math.ceil(validate_limit_max*old_amount):
+            retval = "Valid"
+        elif new_amount<math.floor(validate_limit_min*old_amount):
+            retval = "Data amount is Below given limits"
```
corvidian, line 65: Instead of setting a special variable for the return value, these could log the error and return from the function.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+        except Exception as e:
+            # Transaction implicitly rolls back if an exception occurred within the "with" block
+            logger.error(f"Error: {e}")
+
+    with engine.connect() as connection:
+        try:
+            temp_result = connection.execute(text(temp_sql)).fetchall()
+        except Exception as e:
+            # Transaction implicitly rolls back if an exception occurred within the "with" block
+            logger.error(f"Error: {e}")
```
corvidian, line 51: Would it be better if exceptions was not caught here? The execution can't continue without this result anyway. Or maybe log the exception and re-throw it?

scripts/gis-material-update/validate-deploy/modules/hsl.py
```diff
@@ -0,0 +1,48 @@
...
+    """Process HSL bus lines."""
+
+    def __init__(self, cfg: Config):
+        self._cfg = cfg
```
corvidian, line 11: This class variable isn't used outside `__init__`, so there's no need to initialize it. The `cfg` parameter can be used instead.

scripts/gis-material-update/validate-deploy/validate_deploy_data.py
```diff
@@ -0,0 +1,73 @@
...
+        return CycleInfra(cfg)
+    elif item == "liikennevaylat":
+        return Liikennevaylat(cfg)
+    elif item == "central_business_area":
+        return CentralBusinessAreas(cfg)
+    else:
+        try:
+            raise RuntimeError("Configuration not recognized: {}".format(item))
+        except Exception as e:
+            print("{}".format(e))
```
corvidian, line 51: Why is there a try that always throws an exception?

petri-timlin: Is source given as parameter for gis-validate-deploy? And is the source one of supported sources?

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: There is only "select" -clauses used here so no need for transaction.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+    engine = create_engine(pg_conn_uri)
+    with engine.connect() as connection:
+        transaction = connection.begin()
+        delete_sql = "delete from " + tormays_table_org
+        try:
+            logger.info("Deploy ...")
+            delete_result = connection.execute(text(delete_sql))
+            logger.info(tormays_table_org + ": " + f"{delete_result.rowcount} rows deleted.")
+        except Exception as e:
+            # Transaction implicitly rolls back if an exception occurred within the "with" block
```
corvidian, line 16: The exception needs to be raised when the `with` block ends for the rollback to happen. If the exception is handled in `except`-block, it's no longer raised and the transaction will be committed.

scripts/gis-material-update/validate-deploy/modules/common.py
corvidian: I know, but the comment is misleading.

scripts/gis-material-update/validate-deploy/validate_deploy_data.py
corvidian: Is there a difference between raising an exception in a `try` and printing it in an except as opposed to just printing it directly? I don't know of any, but I'm not very familiar with Python.

```suggestion
        print("{}".format(RuntimeError("Configuration not recognized: {}".format(item))))
```

If there's no difference, we could skip the exception altogether and just print the error directly.
```python
        print("Configuration not recognized: {}".format(item))
```

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+                logger.error(f"Error: {e}")
+            
+def validate_data_count_limits(module, pg_conn_uri, tormays_table_org, tormays_table_temp, validate_limit_min, validate_limit_max, logger=None):
+    # validate data amount: is it between given limits
+    tormays_result = None
+    temp_result = None
+    retval = None
+    old_amount = None
+    new_amount = None
+    tormays_sql = "select count(*) lkm from " + tormays_table_org + " where geometry is not null"
```
corvidian, line 36: There's no geometry columns in the tormays tables. They are called `geom`.

```
haitatonhanketietokantatest=> \d tormays_buses_polys
                                       Table "public.tormays_buses_polys"
    Column    |          Type          | Collation | Nullable |                     Default                      
--------------+------------------------+-----------+----------+--------------------------------------------------
 fid          | integer                |           | not null | nextval('tormays_buses_polys_fid_seq'::regclass)
 route_id     | character varying      |           |          | 
 direction_id | integer                |           |          | 
 rush_hour    | integer                |           |          | 
 trunk        | character varying      |           |          | 
 geom         | geometry(Polygon,3879) |           |          | 
Indexes:
    "tormays_buses_polys_pkey" PRIMARY KEY, btree (fid)
    "tormays_buses_polys_geom_geom_idx" gist (geom)

haitatonhanketietokantatest=> 
```

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+    temp_sql = "select count(*) lkm from " + tormays_table_temp + " where geometry is not null"
```
corvidian, line 37: I tried running the code with hsl data, and I got the error:
```
psycopg2.errors.UndefinedTable: relation "tormays_buses_polys_temp" does not exist
```

When looking at the code I can't find anything that would create that table or anything that would insert data to it. `validate_data_count_limits` checks that there's data in the temp table. `deploy` copies the data from the temp table to the production table. But I can't see anything that would insert anything to the temp table. Am I missing something?

corvidian: I see, the temp table is created in the process stage. I changed my DB config to reference the same database I run the backend against and tried to run gis-validate-deploy again. That's why it didn't have the table in place.

This doesn't feel right in terms of responsibility separation. The process stage also does a part of the deployment. It feels like the processing should finish before the deployment starts. We're leveraging the functionality meant for writing debug data to a temporary table for writing the production data to the production database.

I think there should be a separate configuration for the production database, using the same `HAITATON_HOST`, `HAITATON_PORT`, `HAITATON_USER` and `HAITATON_PASSWORD` environment variables the cloud environments use.

scripts/gis-material-update/process/modules/cycling_infra.py
```diff
@@ -48,26 +49,36 @@ def persist_to_database(self):
...
+        #    "public",
+        #    if_exists="replace",
+        #    index=True,
+        #    index_label="fid",
+        #)
+
+        # persist results to temp table
         self._process_result_polygons.to_postgis(
-            "cycle_infra_polys",
+            self._cfg.tormays_table_temp(self._module),
```
corvidian, line 63: Extra tables are created to the database. E.g. in this file, there's also `cycle_infra` and in hsl.py, there's `buses_lines`. There are similar hardcoded debug tables in almost every process file. If the database configuration points to the main haitaton database, these debug tables will be written there as well. And we clearly don't want them there.

scripts/gis-material-update/validate-deploy/modules/common.py
corvidian: Could the temp table be created at the beginning of the validate-deploy stage from the output of the process stage? I.e. the `.gpkg` file, so in this case, `haitaton-gis-output/tormays_buses_polys.gpkg`.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+        try:
+            logger.info("Deploy ...")
+            delete_result = connection.execute(text(delete_sql))
+            logger.info(tormays_table_org + ": " + f"{delete_result.rowcount} rows deleted.")
+        except Exception as e:
+            # Transaction implicitly rolls back if an exception occurred within the "with" block
+            logger.error(f"Error: {e}")
+        
+        if delete_result:    
+            copy_sql = "insert into " + tormays_table_org + " select * from " + tormays_table_temp
```
corvidian, line 20: Using `SELECT *` with `INSERT` is a bit dangerous, since there's a chance that the columns are in a different order, which could either fail or insert data to the wrong columns. E.g. `rush_hour` to `direction_id` and vice versa for `tormays_buses_polys`.

With delete and insert, there's also no way to change the columns of the target table. Dropping the table and recreating it would enable that.

But this might be something for future development.

scripts/gis-material-update/validate-deploy/modules/common.py
corvidian: After playing around with the transactions for a while I think I understand now: this works correctly, but not quite in the way these comments suggest.

1. There's no autocommit here, so the `transaction.commit()` needs to be called for the commit to happen. This wouldn't be the case if the transaction was defined as a `with` block.
2. If the connection ends without a commit, the transaction is rolled back.
3. If there's an error in the SQL, the database automatically rolls back the transaction.

So this works because we don't let the execution reach the commit unless both statements are executed without errors.

scripts/gis-material-update/validate-deploy/modules/hsl.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+        
+        # Valid --> deploy
+        if self._validate_result == "Valid":
+            self._deploy_result = deploy(
+                self._pg_conn_uri, 
+                self._tormays_table_org, 
+                self._tormays_table_temp, 
+                self._logger
+                )
```
corvidian, line 46: I think the temp table should be deleted after the deployment is completed. This could be done as a separate function call here since it's not critical if it fails. But we want to clean up after ourselves, and not leave the production database with any unnecessary tables.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+    # validate data amount: is it between given limits
+    delete_result = None
+    engine = create_engine(pg_conn_uri)
+    with engine.connect() as connection:
+        transaction = connection.begin()
+        delete_sql = "delete from " + tormays_table_org
+        try:
+            logger.info("Deploy ...")
+            delete_result = connection.execute(text(delete_sql))
+            logger.info(tormays_table_org + ": " + f"{delete_result.rowcount} rows deleted.")
```
corvidian, line 14: Using both concatenation and formatting on the same line is a bit funny.
```suggestion
            logger.info(f"{tormays_table_org}: {delete_result.rowcount} rows deleted.")
```

Overall, I think the formatting style is clearer than concatenation.

scripts/gis-material-update/process/modules/autoliikennemaarat.py
petri-timlin: Actually not added. Those were already earlier in the code but now commented out because those parts are only for as local testing purposes.

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: I will fix this

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Comment removed.

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,91 @@
...
+            new_amount = row.lkm
+    
+    if new_amount and old_amount:
+        if new_amount>=math.floor(validate_limit_min*old_amount) and new_amount<=math.ceil(validate_limit_max*old_amount):
+            retval = "Valid"
+        elif new_amount<math.floor(validate_limit_min*old_amount):
+            retval = "Data amount is Below given limits"
+        elif new_amount>math.ceil(validate_limit_max*old_amount):
+            retval = "Data amount is Above given limits"
+        else:
```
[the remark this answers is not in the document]
petri-timlin: Removed else part

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Totally changed this part. Now uses file not table.

scripts/gis-material-update/validate-deploy/modules/hsl.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/validate_deploy_data.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Now temp table is not created at all because data is coming from file.

scripts/gis-material-update/process/modules/cycling_infra.py
petri-timlin: Removed all unnecessary table creations and added possibility to add original data tables if needed.

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: This part also changed so that tormays table data is replaced data from file and the tormays table is recreated always.

scripts/gis-material-update/validate-deploy/modules/hsl.py
petri-timlin: This part is changed. No unnecessary tables anymore.

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
```diff
@@ -0,0 +1,56 @@
...
+from modules.common import *
+import geopandas as gpd
+
+
+class GisProcessor(ABC):
+    """Abstract base class for GIS processing classes.
+
+    This class helps keeping interface consistent."""
+
+    @abstractmethod
```
corvidian, line 13: The methods are no longer abstract, since they have implementations. Also, the class is no longer an abstract base class, just a base class.

scripts/gis-material-update/validate-deploy/modules/common.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if textval == "Valid":
+        logger.info("Data amount validation result: " + textval)
+        return True
+    elif textval == "Tormays table {} not exists.":
+        logger.error(textval.format(tormays_table_org))
+        return False
+    else:
+        logger.error("Data amount validation result: " + textval)
+        return False
```
corvidian, line 63: I think this would be cleaner if the different branches exited the function instead of setting a special variable.

Logging could be done as soon as possible. This makes it clearer where the logged values come from and removes the need for some variables. The log can also be used to see how far in the process we got if we get an unexpected error.
```suggestion
def validate_data_count_limits(module, pg_conn_uri, tormays_table_org, tormays_file_temp, validate_limit_min, validate_limit_max, filename, logger=None):
    """validate data amount: is it between given limits"""
    logger.info("Data amount validation started")
    logger.info(f"Module: {module}")

    old_amount = get_data_count(pg_conn_uri, tormays_table_org, logger)
    logger.info(f"Old data amount ({tormays_table_org}): {old_amount}")
    new_amount = len(tormays_file_temp)
    logger.info(f"New data amount ({filename}): {new_amount}")

    if new_amount and old_amount and validate_limit_min and validate_limit_max and old_amount != -1:
        min_limit = math.floor(validate_limit_min*old_amount)
        max_limit = math.ceil(validate_limit_max*old_amount)
        logger.info(f"Limits: {min_limit} <= new amount <= {max_limit}")

        if new_amount < min_limit:
            logger.error("Data amount validation failed: Data amount is Below given limits")
            return False
        elif new_amount > max_limit:
            logger.error("Data amount validation failed: Data amount is Above given limits")
            return False
        else:
            logger.info("Data amount is within given limits")
            return True

    elif old_amount == -1:
        logger.error(f"Tormays table {tormays_table_org} not exists.")
        return False
    else:
        logger.error("Data amount validation failed because of missing configuration.")
        return False
```

scripts/gis-material-update/validate-deploy/modules/common.py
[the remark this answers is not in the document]
corvidian: This is a bit weird without the else. There's just an if and too else ifs. 

We can use math here: !(x < a) && !(x > b) <=> a <= x <= b. So if the new amount is not below the minimum or above the maximum, it has to be in between the limits.

```
        if new_amount < min_limit:
            logger.error("Data amount validation failed: Data amount is Below given limits")
            return False
        elif new_amount > max_limit:
            logger.error("Data amount validation failed: Data amount is Above given limits")
            return False
        else:
            logger.info("Data amount is within given limits")
            return True
```

scripts/gis-material-update/validate-deploy/modules/hsl.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, cfg: Config):
+        self._module = "hsl"
+        self._filename = cfg.target_buffer_file(self._module)
+        GisProcessor.__init__(self, cfg)
+
+    def get_temp_data(self, cfg: Config):
+        GisProcessor.get_temp_data(self, cfg)
+
+    def validate_deploy(self):
+        GisProcessor.validate_deploy(self)
```
corvidian, line 19: If the GisProcessor methods are not marked as `@abstractmethod`, the subclasses don't have to implement them, unless they need a different implementation. They inherit the parent class's methods directly.

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
```diff
@@ -0,0 +1,56 @@
...
+        logging.basicConfig(
+            filename=cfg.logging_filename().format(self._module),
+            filemode=cfg.logging_filemode(),
+            format=cfg.logging_format()
+            )
+        self._logger = logging.getLogger()
+        self._logger.setLevel(cfg.logging_level())
+
+    @abstractmethod
+    def get_temp_data(self, cfg):
```
corvidian, line 28: `cfg` parameter is not used.

scripts/gis-material-update/validate-deploy/modules/config.py
```diff
@@ -0,0 +1,143 @@
...
+
+        if directory_name.exists():
+            return str(directory_name)
+        else:
+            return None
+
+    def deployment_profile(self) -> str:
+        return self._deployment_profile
+
+    def local_file(self, item: str) -> str:
```
corvidian, line 71: This is never called.

scripts/gis-material-update/validate-deploy/modules/config.py
```diff
@@ -0,0 +1,143 @@
...
+        file_path = self._file_directory("output_dir")
+        return "/".join([file_path, self._cfg.get(item, {}).get("target_buffer_file")])
+
+    def crs(self) -> str:
+        """Return CRS information from config file."""
+        return self._cfg.get("common").get("crs")
+
+    def layer(self, item: str) -> str:
+        """Return layer name of source data from configuration."""
+        return self._cfg.get(item, {}).get("layer")
```
corvidian, line 92: These are never called.

scripts/gis-material-update/validate-deploy/modules/config.py
```diff
@@ -0,0 +1,143 @@
...
+        """Return buffer value list from configuration."""
+        return self._cfg.get(item, {}).get("validate_limit_min")
+
+    def validate_limit_max(self, item: str):
+        """Return buffer value list from configuration."""
+        return self._cfg.get(item, {}).get("validate_limit_max")
+
+    def buffer_class_values(self, item: str) -> list[int]:
+        """Return buffer value list from configuration."""
+        return self._cfg.get(item, {}).get("buffer_class_values")
```
corvidian, line 112: This is never called.

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
corvidian: I see, it's used in `autoliikennemaart.py`. But it doesn't need to be used there. The line that uses `cfg` can be moved to `__init__`.
```
self._filename = cfg.target_buffer_file(self._module)
```

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
```diff
@@ -0,0 +1,56 @@
...
+            self._tormays_file_temp,
+            self._validate_limit_min,
+            self._validate_limit_max,
+            self._filename,
+            self._logger
+            )
+
+        # Valid --> deploy
+        if self._validate_result is True:
+            self._deploy_result = deploy(
```
corvidian, line 49: The value _deploy_result is never used.
```suggestion
            deploy(
```

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
```diff
@@ -0,0 +1,56 @@
...
+    @abstractmethod
+    def get_temp_data(self, cfg):
+        self._tormays_file_temp = gpd.read_file(self._filename)
+        self._tormays_file_temp.rename_geometry('geom', inplace=True)
+
+    @abstractmethod
+    def validate_deploy(self):
+
+        # validate data amount: is it between given limits
+        self._validate_result = validate_data_count_limits(
```
corvidian, line 36: `self._validate_result` can be a local variable.
```suggestion
        validate_result = validate_data_count_limits(
```

scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
```diff
@@ -0,0 +1,49 @@
...
+        for buffer in self._buffers:
+            buffer_data = gpd.read_file(self._filename.format(buffer))
+            buffer_data.rename_geometry('geom', inplace=True)
+            self._tormays_files_temp[buffer] = buffer_data
+
+    def validate_deploy(self):
+
+        # validate data amount: is it between given limits
+        buffers = self._buffers
+        self._validate_result = False
```
corvidian, line 28: `self._validate_result` can be a local variable.
```suggestion
        validate_result = False
```

scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
```diff
@@ -0,0 +1,49 @@
...
+
+        for buffer in buffers:
```
corvidian, line 30: This can use `self._buffers` directly. There's no need to assign it to the `buffers` variable.
```suggestion
        for buffer in self._buffers:
```

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
```diff
@@ -0,0 +1,56 @@
+import logging
+
+from abc import ABC, abstractmethod
+from modules.common import *
```
corvidian, line 4: Using wildcard imports is generally discouraged. They make it hard to see where a particular function is defined and can easily cause name clashes.

You can either import the methods one-by-one:
```suggestion
from modules.common import validate_data_count_limits
from modules.common import deploy
```

Or use common as a module, and call the methods with `common.deploy(` etc.
```suggestion
import modules.common as common
```
Since common is a module, maybe it should be used as one?

scripts/gis-material-update/validate-deploy/modules/config.py
corvidian: The preview is a bit unclear, but I mean the `local_file` method.

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
corvidian: The other `self.` value could be moved to `__init__` as well. It seems there's a convention in Python to define all class variables in `__init__`.

scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/autoliikennemaarat.py
petri-timlin: changed

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/config.py
petri-timlin: Removed

scripts/gis-material-update/validate-deploy/modules/config.py
petri-timlin: Removed

scripts/gis-material-update/validate-deploy/modules/config.py
petri-timlin: Removed

scripts/gis-material-update/validate-deploy/modules/gis_validate_deploy.py
petri-timlin: Changed to this:
from modules.common import validate_data_count_limits, deploy

scripts/gis-material-update/validate-deploy/modules/common.py
petri-timlin: Changed

scripts/gis-material-update/validate-deploy/modules/hsl.py
petri-timlin: Removed methods. Same done also other modules except autoliikennemaarat.

corvidian (approved):

PR #628, opened 2024-02-23
HAI-2253 DTO for application update and checks for whether it contains updated data

umeetiusbaar:
# Description

A new simpler dto for application update at endpoint `PUT /hakemukset/{hakemusId}`. This PR contains only the dto and checking whether it contains updated data compared to old. Future PRs will add service and controller functionality.

Also, adds environment variables for features in docker-compose.yml.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
This PR does not add any user-testable functionality.

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -0,0 +1,146 @@
...
+            constructionWork != applicationData.constructionWork ||
+            maintenanceWork != applicationData.maintenanceWork ||
+            propertyConnectivity != applicationData.propertyConnectivity ||
+            emergencyWork != applicationData.emergencyWork ||
+            rockExcavation != applicationData.rockExcavation ||
+            workDescription != applicationData.workDescription ||
+            startTime != applicationData.startTime ||
+            endTime != applicationData.endTime ||
+            areas != applicationData.areas ||
+            customerWithContacts?.hasChanges(applicationData.customerWithContacts)
```
corvidian, line 85: With kortisto, `CableReportApplicationData` will not have `customerWithContacts`, since the customers and contacts will be saved separately to DB.

I think the comparison should be done against ApplicationEntity.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -0,0 +1,146 @@
...
+    fun hasChanges(customerWithContacts: CustomerWithContacts?): Boolean =
+        customerWithContacts?.let {
+            customer.hasChanges(it.customer) ||
+                contacts.size != customerWithContacts.contacts.size ||
+                contacts.map { contact -> contact.id }.toSet() !=
+                    customerWithContacts.contacts.map { contact -> contact.id }.toSet()
+        } ?: !customer.isEmpty()
+}
+
+data class HakemusUpdateRequestCustomer(
```
corvidian, line 112: This will want an `id` field, so we know if the user wants to update an existing customer or create a new one.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -0,0 +1,146 @@
...
+            emergencyWork != applicationData.emergencyWork ||
+            rockExcavation != applicationData.rockExcavation ||
+            workDescription != applicationData.workDescription ||
+            startTime != applicationData.startTime ||
+            endTime != applicationData.endTime ||
+            areas != applicationData.areas ||
+            customerWithContacts?.hasChanges(applicationData.customerWithContacts)
+                ?: !applicationData.customerWithContacts.isNullOrBlank() ||
+            contractorWithContacts?.hasChanges(applicationData.contractorWithContacts)
+                ?: !applicationData.contractorWithContacts.isNullOrBlank() ||
```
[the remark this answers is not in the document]
umeetiusbaar: I refactored these comparisons.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: Ah, yes! Added the id. Thanks!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
corvidian: Changing the name fields is nice if we also change the frontend to match.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (this.size != hakemusyhteyshenkilot.size) return true
+    hakemusyhteyshenkilot
+        .map { it.hankekayttaja.id }
+        .toSet()
+        .let { existingIds ->
+            return this.any { hakemusyhteyshenkiloRequest ->
+                hakemusyhteyshenkiloRequest.id !in existingIds
+            }
+        }
```
corvidian, line 121: I think this can be simplified. This method could be tested separately to cover all execution paths.
```suggestion
    if (this == null) {
        return hakemusyhteyshenkilot.isNotEmpty()
    }
    val requestIds = map { it.id }.toSet()
    val existingIds = hakemusyhteyshenkilot.map { it.hankekayttaja.id }.toSet()
    return requestIds != existingIds
```

corvidian: Oh, that doesn't cover the tilaaja field.
```suggestion
    if (this == null) {
        return hakemusyhteyshenkilot.isNotEmpty()
    }
    val requests = this.toSet()
    val existing =
        hakemusyhteyshenkilot.map { HakemusyhteyshenkiloRequest(it.id, it.tilaaja) }.toSet()
    return requests != existing
```

corvidian: Come to think of it, the tilaaja field shouldn't be updatable by the user. We're supposed to set it when the the application is sent to Allu.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -0,0 +1,122 @@
...
+fun Map<ApplicationContactType, HakemusyhteystietoRequest>?.hasChanges(
+    hakemusyhteystiedot: Map<ApplicationContactType, HakemusyhteystietoEntity>
+): Boolean {
+    if (this == null) {
+        return hakemusyhteystiedot.isNotEmpty()
+    }
+    if (this.size != hakemusyhteystiedot.size) return true
+    return this.entries.any { (role, yhteystietoRequest) ->
+        hakemusyhteystiedot[role]?.let { entity -> yhteystietoRequest.hasChanges(entity) } ?: true
+    }
```
corvidian, line 104: If we check that the keys match instead of sizes, we know `hakemusyhteystiedot[role]` can't be null.

It's also easier to see why this covers cases where hakemusyhteystiedot has information this doesn't.
```suggestion
    if (this.keys != hakemusyhteystiedot.keys) return true
    return this.entries.any { (role, yhteystietoRequest) ->
        yhteystietoRequest.hasChanges(hakemusyhteystiedot[role]!!)
    }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: Since the frontend needs to be changed for contacts and contact persons I think these other changes would go with the same effort. The data structures that UI uses internally may still remain as they are, only creating the update request would need updating.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: I implemented your suggestions for simplifications. And removed "tilaaja" field. Thanks!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: Thanks again! I use your suggestions.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestTest.kt
```diff
@@ -0,0 +1,153 @@
...
+        @Test
+        fun `returns true when name is changed`() {
+            val original =
+                ApplicationFactory.createApplicationEntity(
+                    applicationData = ApplicationFactory.createBlankCableReportApplicationData(),
+                    hanke = HankeFactory.createMinimalEntity()
+                )
+            val request =
+                "/fi/hel/haitaton/hanke/hakemus/hakemusUpdateRequest-with-name.json"
+                    .asJsonResource<JohtoselvityshakemusUpdateRequest>()
```
corvidian, line 47:
```suggestion
            val request =
                "/fi/hel/haitaton/hanke/hakemus/hakemusUpdateRequest-blank.json"
                    .asJsonResource<JohtoselvityshakemusUpdateRequest>()
                    .copy(name = "Testihakemus")
```

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestTest.kt
```diff
@@ -0,0 +1,153 @@
...
+        @Test
+        fun `returns true when work description is changed`() {
+            val original =
+                ApplicationFactory.createApplicationEntity(
+                    applicationData = ApplicationFactory.createBlankCableReportApplicationData(),
+                    hanke = HankeFactory.createMinimalEntity()
+                )
+            val request =
+                "/fi/hel/haitaton/hanke/hakemus/hakemusUpdateRequest-with-work-description.json"
+                    .asJsonResource<JohtoselvityshakemusUpdateRequest>()
```
corvidian, line 61:
```suggestion
            val request =
                "/fi/hel/haitaton/hanke/hakemus/hakemusUpdateRequest-blank.json"
                    .asJsonResource<JohtoselvityshakemusUpdateRequest>()
                    .copy(tyonKuvaus = "Lorem ipsum dolor sit amet, consectetur adipiscing elit.")
```

services/hanke-service/src/test/resources/fi/hel/haitaton/hanke/hakemus/hakemusUpdateRequest-blank.json
```diff
@@ -0,0 +1,14 @@
+{
```
corvidian, line 1: This could as well be a static object in Kotlin. But that way, any changes would cause compilation errors instead of errors trying to read the JSON during tests.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: That is a good and valid point. Would it be too much changes to both backend and frontend if the new HakemusController endpoints would return a flatter structure with Finnish naming instead of current Application+ApplicationData structure with English/Allu naming? Maybe refactoring to flat+Finnish is something for the future and I should revert my changes in this update request also to follow ApplicationData+English naming?

umeetiusbaar: I will revert these flat+Finnish changes at this point

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestTest.kt
umeetiusbaar: I removed jsons completely.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestTest.kt
umeetiusbaar: I removed jsons completely.

corvidian (approved):

PR #633, opened 2024-02-29
HAI-2338 Implement gdpr delete dry-run for kortisto

corvidian:
# Description

Refactor the GDPR service dry-run to not return a list of applications. Instead, return true if the information can be deleted and throw an exception if it can't. This also means that the delete call needs to be refactored to not receive a list of applications, but fetch them from the DB again.

Besides supporting both the old and Kortisto model of GDPR service, this also has the added benefit of loading the information to be deleted in the same transaction as they are deleted in, so there's no chance of the information changing between the loading and deleting.

Implement the canDelete method for `KortistoGdprService`. This checks that the user is not on any active applications, and is not the only admin user (user with KAIKKI_OIKEUDET privileges) in the hanke. If the user shouldn't be deleted, all the conflicting applications are gathered and thrown in an exception. The controller returns the errors in the correct format to Profiili.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Can be tested with the profile-gdpr-api-tester.

# Checklist:

 or other location: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8075476993/GDPR+API#Kortiston-my%C3%B6t%C3%A4

# Other relevant info
The `deleteInfo` method will be implemented in the next PR.

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
@@ -5,7 +5,9 @@ import assertk.all
 import assertk.assertFailure
 import assertk.assertThat
 import assertk.assertions.any
+import assertk.assertions.contains
 import assertk.assertions.containsExactlyInAnyOrder
+import assertk.assertions.first
 import assertk.assertions.hasClass
 import assertk.assertions.hasSize
 import assertk.assertions.isEmpty
@@ -13,6 +15,7 @@ import assertk.assertions.isEqualTo
 import assertk.assertions.isInstanceOf
 import assertk.assertions.isNotNull
 import assertk.assertions.isNull
+import assertk.assertions.isTrue
 import assertk.assertions.prop
 import assertk.assertions.single
 import fi.hel.haitaton.hanke.DatabaseTest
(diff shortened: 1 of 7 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
```diff
@@ -200,20 +207,122 @@ class KortistoGdprServiceITest(
...
+            val activeHakemus =
+                hakemusFactory.builder(USERID, hanke).inHandling().saveWithYhteystiedot {
+                    rakennuttaja()
+                }
+
+            val failure = assertFailure { gdprService.canDelete(USERID) }
+
+            failure.given { e ->
+                val messages = (e as DeleteForbiddenException).errors.map { it.message.fi }
+                assertThat(messages).single().contains(activeHakemus.applicationIdentifier!!)
```
umeetiusbaar, line 303: Maybe this test should assert that there is something different in the result compared to test `throws exception when there's an active application`? Or is there any difference between the results in these two tests?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/gdpr/GdprService.kt
```diff
@@ -80,7 +154,13 @@ class OldGdprService(private val applicationService: ApplicationService) : GdprS
     }
 
     @Transactional(readOnly = true)
-    override fun findApplicationsToDelete(userId: String): List<Application> {
+    override fun canDelete(userId: String): Boolean {
+        findApplicationsToDelete(userId)
+        // Will throw an exception if the information can't be deleted
```
umeetiusbaar, line 159: This comment might be better to be above the `findApplicationsToDelete` or on the same line with it.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
corvidian: The errors they return are exactly the same. So a request might fail on both for the same hakemus. The response is mapped to a set to get rid of the duplicates.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/gdpr/GdprService.kt
corvidian: I moved the comment.

umeetiusbaar (approved):

PR #636, opened 2024-02-29
HAI-2377 Add invitation date to hankekayttaja response

corvidian:
# Description

Add the date and time the latest invitation to a user was sent. If the invitation is resent, the date and time will reset as well.

Also, remove the deprecated `nimi` field from HankeKayttaja and the DTO. The frontend doesn't use that field anymore, so it's safe to delete it now.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
@@ -262,10 +262,10 @@ class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : C
                 prop(HankeKayttajaDto::sahkoposti).isEqualTo("email.1.address.com")
                 prop(HankeKayttajaDto::etunimi).isEqualTo("test1")
                 prop(HankeKayttajaDto::sukunimi).isEqualTo("name1")
-                prop(HankeKayttajaDto::nimi).isEqualTo("test1 name1")
                 prop(HankeKayttajaDto::puhelinnumero).isEqualTo("0405551111")
                 prop(HankeKayttajaDto::kayttooikeustaso).isEqualTo(Kayttooikeustaso.KATSELUOIKEUS)
                 prop(HankeKayttajaDto::tunnistautunut).isEqualTo(false)
+                prop(HankeKayttajaDto::kutsuttu).isEqualTo(HankeKayttajaFactory.INVITATION_DATE)
             }
             verifySequence {
                 authorizer.authorizeKayttajaId(kayttajaId, "VIEW")
@@ -303,10 +303,10 @@ class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : C
                 assertThat(id).isNotNull()
                 assertThat(etunimi).isEqualTo("test1")
                 assertThat(sukunimi).isEqualTo("name1")
-                assertThat(nimi).isEqualTo("test1 name1")
                 assertThat(puhelinnumero).isEqualTo("0405551111")
                 assertThat(sahkoposti).isEqualTo("email.1.address.com")
                 assertThat(tunnistautunut).isEqualTo(false)
+                assertThat(kutsuttu).isEqualTo(HankeKayttajaFactory.INVITATION_DATE)
                 assertThat(roolit).containsExactlyInAnyOrder(ContactType.OMISTAJA, ContactType.MUU)
             }
             assertThat(response.kayttajat).hasSameElementsAs(testData)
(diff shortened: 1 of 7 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
```diff
@@ -303,10 +303,10 @@ class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : C
                 assertThat(id).isNotNull()
                 assertThat(etunimi).isEqualTo("test1")
                 assertThat(sukunimi).isEqualTo("name1")
-                assertThat(nimi).isEqualTo("test1 name1")
                 assertThat(puhelinnumero).isEqualTo("0405551111")
                 assertThat(sahkoposti).isEqualTo("email.1.address.com")
                 assertThat(tunnistautunut).isEqualTo(false)
+                assertThat(kutsuttu).isNotNull().isEqualTo(HankeKayttajaFactory.INVITATION_DATE)
```
umeetiusbaar, line 309: Is `isNotNull()` check needed or does `isEqualTo` cover also null case?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
```diff
@@ -431,10 +431,10 @@ class HankeKayttajaServiceITest : DatabaseTest() {
                 prop(HankeKayttajaDto::sahkoposti).isEqualTo(email)
                 prop(HankeKayttajaDto::etunimi).isEqualTo("Joku")
                 prop(HankeKayttajaDto::sukunimi).isEqualTo("Jokunen")
-                prop(HankeKayttajaDto::nimi).isEqualTo("Joku Jokunen")
                 prop(HankeKayttajaDto::puhelinnumero).isEqualTo("0508889999")
                 prop(HankeKayttajaDto::kayttooikeustaso).isEqualTo(Kayttooikeustaso.KATSELUOIKEUS)
                 prop(HankeKayttajaDto::tunnistautunut).isFalse()
+                prop(HankeKayttajaDto::kutsuttu).isNotNull().isRecent()
```
umeetiusbaar, line 437: If I understand correctly `isRecent` assertion succeeds if the value is `null`. I wonder if that is the right way to assert the time? In some places, where `isRecent` is used there is no `isNotNull` assertions accompaning it and at least my intuition would add the null-check implicitely in `isRecent`. How can a null date/time be recent?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttaja.kt
```diff
@@ -41,13 +40,12 @@ data class HankeKayttajaDto(
...
     @field:Schema(description = "Last name") val sukunimi: String,
-    @Deprecated("Use etunimi and sukunimi instead.")
-    @field:Schema(description = "Full name", deprecated = true)
-    val nimi: String,
     @field:Schema(description = "Phone number") val puhelinnumero: String,
     @field:Schema(description = "Access level in Hanke") val kayttooikeustaso: Kayttooikeustaso?,
     @field:Schema(description = "User roles in Hanke") val roolit: List<ContactType>,
     @field:Schema(description = "Has user logged in to view Hanke") val tunnistautunut: Boolean,
+    @field:Schema(description = "When their latest invitation was sent")
+    val kutsuttu: OffsetDateTime?,
```
umeetiusbaar, line 48: I guess that `kutsuttu` is never `null` for any one else than the founder of the hanke? And the founder never needs a tooltip where this information would be shown. So, the UI does not need to bother with null values for users that should show the invitation date, right?

corvidian: The invitation is deleted when accepted, so `kutsuttu` will be null for all identified users. There should be a value for all unidentified users since it's a non-nullable field in the invitation and all users should have exactly one of either permissions or invitations.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
corvidian: Yes, `assertThat(null).isEqualTo(INVITATION_DATE)` fails, so the null check is not required. I removed it.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
corvidian: Yes, I agree that `assertThat(null).isRecent()` should fail. I changed all the `isRecent` type asserts to fail with nulls.

I also thought about making them work only on non-null values, but that would cause unnecessary bloat when used.

umeetiusbaar (approved):

PR #637, opened 2024-03-03
HAI-2253 Implementation of HakemusService.updateHakemus

umeetiusbaar:
# Description

HakemusService.updateHakemus() implementation

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
JUnit integration tests are currently only way to test the implementation.

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
@@ -15,22 +15,20 @@ import assertk.assertions.messageContains
 import assertk.assertions.prop
 import assertk.assertions.single
 import fi.hel.haitaton.hanke.DatabaseTest
-import fi.hel.haitaton.hanke.HankeEntity
 import fi.hel.haitaton.hanke.HankeNotFoundException
 import fi.hel.haitaton.hanke.allu.CustomerType
 import fi.hel.haitaton.hanke.application.ApplicationAlreadySentException
 import fi.hel.haitaton.hanke.application.ApplicationContactType
-import fi.hel.haitaton.hanke.application.ApplicationEntity
 import fi.hel.haitaton.hanke.application.ApplicationGeometryException
 import fi.hel.haitaton.hanke.application.ApplicationGeometryNotInsideHankeException
 import fi.hel.haitaton.hanke.application.ApplicationNotFoundException
 import fi.hel.haitaton.hanke.application.ApplicationRepository
-import fi.hel.haitaton.hanke.application.CableReportApplicationData
 import fi.hel.haitaton.hanke.asJsonResource
 import fi.hel.haitaton.hanke.factory.ApplicationFactory
 import fi.hel.haitaton.hanke.factory.GeometriaFactory
 import fi.hel.haitaton.hanke.factory.HakemusFactory
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory
+import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.toUpdateRequest
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withAreas
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withCustomerWithContactsRequest
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withWorkDescription
@@ -239,30 +237,29 @@ class HakemusServiceITest(
 
         @Test
         fun `throws exception when the application has been sent to Allu`() {
-            val application = createHakemus { alluid = 21 }
-            val request =
-                HakemusUpdateRequestFactory.createFilledJohtoselvityshakemusUpdateRequest()
+            val entity = hakemusFactory.builder(USERNAME).withStatus(alluId = 21).save()
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
+            val request = hakemus.toUpdateRequest()
 
             val exception = assertFailure {
-                hakemusService.updateHakemus(application.id!!, request, USERNAME)
+                hakemusService.updateHakemus(hakemus.id, request, USERNAME)
             }
 
             exception.all {
                 hasClass(ApplicationAlreadySentException::class)
-                messageContains("id=${application.id}")
+                messageContains("id=${hakemus.id}")
                 messageContains("alluid=21")
             }
         }
 
         @Test
         fun `does not create a new audit log entry when the application has not changed`() {
-            val application = createHakemus()
+            val entity = hakemusFactory.builder(USERNAME).save()
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
             val originalAuditLogSize = auditLogRepository.findByType(ObjectType.APPLICATION).size
-            val request =
-                HakemusUpdateRequestFactory
-                    .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(application)
+            val request = hakemus.toUpdateRequest()
 
-            hakemusService.updateHakemus(application.id!!, request, USERNAME)
+            hakemusService.updateHakemus(hakemus.id, request, USERNAME)
 
             val applicationLogs = auditLogRepository.findByType(ObjectType.APPLICATION)
             assertThat(applicationLogs).hasSize(originalAuditLogSize)
@@ -270,19 +267,17 @@ class HakemusServiceITest(
 
         @Test
         fun `throws exception when there are invalid geometry in areas`() {
-            val application = createHakemus()
-            val request =
-                HakemusUpdateRequestFactory
-                    .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(application)
-                    .withAreas(listOf(intersectingArea))
+            val entity = hakemusFactory.builder(USERNAME).save()
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
+            val request = hakemus.toUpdateRequest().withAreas(listOf(intersectingArea))
 
             val exception = assertFailure {
-                hakemusService.updateHakemus(application.id!!, request, USERNAME)
+                hakemusService.updateHakemus(hakemus.id, request, USERNAME)
             }
 
             exception.all {
                 hasClass(ApplicationGeometryException::class)
-                messageContains("id=${application.id}")
+                messageContains("id=${hakemus.id}")
                 messageContains("reason=Self-intersection")
                 messageContains(
                     "location={\"type\":\"Point\",\"coordinates\":[25494009.65639264,6679886.142116806]}"
@@ -292,20 +287,21 @@ class HakemusServiceITest(
 
         @Test
         fun `throws exception when the request has a persisted contact but the application does not`() {
-            val application = createHakemus()
+            val entity = hakemusFactory.builder(USERNAME).save()
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
             val requestYhteystietoId = UUID.randomUUID()
             val request =
-                HakemusUpdateRequestFactory
-                    .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(application)
+                hakemus
+                    .toUpdateRequest()
                     .withCustomerWithContactsRequest(CustomerType.COMPANY, requestYhteystietoId)
 
             val exception = assertFailure {
-                hakemusService.updateHakemus(application.id!!, request, USERNAME)
+                hakemusService.updateHakemus(hakemus.id, request, USERNAME)
             }
 
             exception.all {
                 hasClass(InvalidHakemusyhteystietoException::class)
-                messageContains("id=${application.id}")
+                messageContains("id=${hakemus.id}")
                 messageContains("role=${ApplicationContactType.HAKIJA}")
                 messageContains("yhteystietoId=null")
                 messageContains("newId=$requestYhteystietoId")
@@ -314,21 +310,22 @@ class HakemusServiceITest(
 
         @Test
         fun `throws exception when the request has different persisted contact than the application`() {
-            val application = createHakemusWithHakija()
+            val entity = hakemusFactory.builder(USERNAME).saveWithYhteystiedot { hakija() }
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
             val originalYhteystietoId = hakemusyhteystietoRepository.findAll().first().id
             val requestYhteystietoId = UUID.randomUUID()
             val request =
-                HakemusUpdateRequestFactory
-                    .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(application)
+                hakemus
+                    .toUpdateRequest()
                     .withCustomerWithContactsRequest(CustomerType.COMPANY, requestYhteystietoId)
 
             val exception = assertFailure {
-                hakemusService.updateHakemus(application.id!!, request, USERNAME)
+                hakemusService.updateHakemus(hakemus.id, request, USERNAME)
             }
 
             exception.all {
                 hasClass(InvalidHakemusyhteystietoException::class)
-                messageContains("id=${application.id}")
+                messageContains("id=${hakemus.id}")
                 messageContains("role=${ApplicationContactType.HAKIJA}")
                 messageContains("yhteystietoId=$originalYhteystietoId")
                 messageContains("newId=$requestYhteystietoId")
@@ -337,12 +334,13 @@ class HakemusServiceITest(
 
         @Test
         fun `throws exception when the request has a contact that is not a user on hanke`() {
-            val application = createHakemusWithHakija()
+            val entity = hakemusFactory.builder(USERNAME).saveWithYhteystiedot { hakija() }
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
             val yhteystieto = hakemusyhteystietoRepository.findAll().first()
             val requestHankekayttajaId = UUID.randomUUID()
             val request =
-                HakemusUpdateRequestFactory
-                    .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(application)
+                hakemus
+                    .toUpdateRequest()
                     .withCustomerWithContactsRequest(
                         CustomerType.COMPANY,
                         yhteystieto.id,
@@ -350,12 +348,12 @@ class HakemusServiceITest(
                     )
 
             val exception = assertFailure {
-                hakemusService.updateHakemus(application.id!!, request, USERNAME)
+                hakemusService.updateHakemus(hakemus.id, request, USERNAME)
             }
 
             exception.all {
                 hasClass(InvalidHakemusyhteyshenkiloException::class)
-                messageContains("id=${application.id}")
+                messageContains("id=${hakemus.id}")
                 messageContains("invalidHankeKayttajaIds=[$requestHankekayttajaId]")
             }
         }
@@ -363,19 +361,17 @@ class HakemusServiceITest(
         @Test
         fun `throws exception when area is not inside hanke area`() {
             val hanke = hankeFactory.builder(USERNAME).withHankealue().saveEntity()
-            val application = createHakemus(hanke)
-            val request =
-                HakemusUpdateRequestFactory
-                    .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(application)
-                    .withAreas(listOf(notInHankeArea))
+            val entity = hakemusFactory.builder(USERNAME, hanke).save()
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
+            val request = hakemus.toUpdateRequest().withAreas(listOf(notInHankeArea))
 
             val exception = assertFailure {
-                hakemusService.updateHakemus(application.id!!, request, USERNAME)
+                hakemusService.updateHakemus(hakemus.id, request, USERNAME)
             }
 
             exception.all {
                 hasClass(ApplicationGeometryNotInsideHankeException::class)
-                messageContains("id=${application.id}")
+                messageContains("id=${hakemus.id}")
                 messageContains("hankeId=${hanke.id}")
                 messageContains("geometry=${notInHankeArea.geometry.toJsonString()}")
             }
@@ -384,13 +380,12 @@ class HakemusServiceITest(
         @Test
         fun `saves updated data and creates an audit log`() {
             val hanke = hankeFactory.builder(USERNAME).withHankealue().saveEntity()
-            val application =
-                createHakemusWithHakija(hanke) {
-                    applicationData =
-                        (applicationData as CableReportApplicationData).copy(
-                            workDescription = "Old work description"
-                        )
-                }
+            val entity =
+                hakemusFactory
+                    .builder(USERNAME, hanke)
+                    .withWorkDescription("Old work description")
+                    .saveWithYhteystiedot { hakija() }
+            val hakemus = hakemusService.hakemusResponse(entity.id!!)
             val yhteystieto = hakemusyhteystietoRepository.findAll().first()
             val kayttaja =
                 hankekayttajaRepository
@@ -399,8 +394,8 @@ class HakemusServiceITest(
             val newKayttaja = hankeKayttajaFactory.saveUser(hanke.id)
             val originalAuditLogSize = auditLogRepository.findByType(ObjectType.APPLICATION).size
             val request =
-                HakemusUpdateRequestFactory
-                    .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(application)
+                hakemus
+                    .toUpdateRequest()
                     .withCustomerWithContactsRequest(
                         CustomerType.COMPANY,
                         yhteystieto.id,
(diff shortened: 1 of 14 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -1,21 +1,81 @@
...
 class HakemusService(
     private val applicationRepository: ApplicationRepository,
     private val hankeRepository: HankeRepository,
+    private val geometriatDao: GeometriatDao,
+    private val hankealueService: HankealueService,
+    private val applicationLoggingService: ApplicationLoggingService,
+    private val hankeKayttajaService: HankeKayttajaService
 ) {
+    @Transactional
+    fun createHakemus(hakemus: Hakemus, userId: String): Hakemus {
```
corvidian, line 45: Why is this added here? There are other tickets for re-implementing the hakemus creation. As-is, this method is useless and riddled with problems.

1. Production code shouldn't be added to make testing easier. When writing production code just for tests, the implementation tends to become skewed towards testing and it becomes hard to tell what code is actually needed and what is not. This method is only called from the 
2. This method would be the one we use for the `POST /hakemus` endpoint, for creating an application for an existing hanke. We agreed in https://helsinkisolutionoffice.atlassian.net/browse/HAI-2374 that it should only accept the fields that are on the first page of the application form.
3. Even if we hadn't agreed to do that, there should be a separate DTO for creating a hakemus, which would be quite close to the update DTO.
4. The internal status fields (`alluid`, `applicationIdentifier`, `allustatus`) should never be updated based on input.
5. Please add a commit message that describes the changes. Things like this should definitely be mentioned in the commit message, along with an explanation of why it was added.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -12,12 +12,12 @@ import java.time.ZonedDateTime
 import java.util.UUID
 
 sealed interface HakemusUpdateRequest {
-    val name: String
-    val postalAddress: PostalAdressRequest
-    val startTime: ZonedDateTime?
-    val endTime: ZonedDateTime?
-    val areas: List<ApplicationArea>?
-    val customerWithContacts: CustomerWithContactsRequest?
+    var name: String
```
corvidian, line 15: Why were all fields changed to vars? There's no reason for the request to change after creation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
corvidian: 6. There are no tests for the method. Tests should in place as soon as a method is added, not added at a later point.

umeetiusbaar: You're right of course. I tried to follow what has been done in Application and Hanke Service integration tests where the Service is used to create test data. But now, since there is no create method yet in this new HankeService, I should not have put it there but instead in HakemusBuilder or HakemusFactory. The creation needs to be done inside Transaction though and I had trouble to get it done without Service using only Repositories to persist the test data. I tried add @Transactional in the test method itself but that did not seem right either. So, in the end after trying many different paths I ended up putting it into HakemuService. Which was wrong as you  have pointed out. I will refactor and put the creation of test data in test factories / builders. Maybe also without any checks as the purpose is only have the database populated with test data. Or maybe I keep the checks (for geometries and hanke users etc) so that the test data is valid. Or should I do HAI-2374 first before this and use its create method after it has been done right? As have been done in ApplicationServiceITest.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: This is also something I did for testing purposes where I create the request from Hakemus data object and then change something in it. I could use vals and copy instead but I had trouble figuring out how to do it with nested fields, e.g. to change HakemusUpdateRequest.customerWithContacts.contact.id. using copy. Maybe it is easy and I have to just google a bit moore for the solution.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
corvidian: Do you need to do something you can't do with the existing HakemusFactory / ApplicationBuilder? ApplicationBuilder returns an Application, but it uses the new customers and contacts and it's of course saved in the same DB as any Hakemus would be saved to. So the first approach should be to expand on the existing builder.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
corvidian: Using copy on nested objects, especially deeply nested objects, can be annoying. But it can be made less annoying by using extension methods. It's also often easier to create the right kind of nested object than it is to modify an existing one.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: The problem is to have a non-entity object (Hakemus) from an entity object (ApplicationEntity) without a session. This is not possible due to ApplicationEntity.yhteystiedot being lazily fetched. So, I need a place where I can persist an ApplicationEntity and return it as a full domain object. E.g. HakemusBuilder.saveWithyhteystiedot returns the entity, not a domain object. If I try to access any lazy fields (e.g. yhteystiedot) of the entity, I get an error. But, now I got it to work by doing the persisting save in HakemusFactory.save with a transaction. Maybe, when there is a proper HakemusService.create we can use that in tests?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: Still trying to figure out a proper extension function mechanism for copy. But in the mean time I reverted the fields as vals.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
corvidian: Do you even need the Hakemus domain object in this PR? It's only used in generating test data, which can be generated by other means.

umeetiusbaar: Hmm.. yes.. good point. Why did I create it in the first place.. was my plan at some point to use it in Service layer and smaller Request/Response dtos in Controller only.. but that was not a good plan.. either. Or maybe I thought of doing some "modulith" things and keeping entities and such only for "hakemus" package and communication with other packages (modules) would be done with the Hakemus object (and perhaps hopefully make things more clear with clear boudaries). Hmm hmm.. I'll remove it because you're right, I do not need it in this PR. But, that does not mean there is no problem of creating a pojo (HakemusUpdateRequest in this case) using an entity when there are lazy fields.

umeetiusbaar: I got rid of Hakemus and got testing to work with just repositories. Test code could use some simplyfication with e.g. extension functions.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -113,38 +167,287 @@ class HakemusServiceITest : DatabaseTest() {
...
+                        workDescription = "New work description",
+                        // add a new contact
+                        customerWithContacts =
+                            HakemusUpdateRequestFactory.createCustomerWithContactsRequest(
+                                CustomerType.COMPANY,
+                                yhteystieto.id,
+                                kayttaja.id,
+                                newKayttaja.id
+                            )
+                    )
```
corvidian, line 417: This is one place where generation can be made simpler by using extension methods. The copy can be hidden inside the extension method, making the result a lot cleaner.
```suggestion
                    .withCustomer(CustomerType.COMPANY, yhteystieto.id, kayttaja.id, newKayttaja.id)
                    .withDescription("New work description")
```

The extension functions are simple and can be added to `HakemusUpdateRequestFactory`:
```kotlin
    fun JohtoselvityshakemusUpdateRequest.withCustomer(
        customerType: CustomerType = CustomerType.COMPANY,
        yhteystietoId: UUID? = UUID.randomUUID(),
        vararg hankekayttajaIds: UUID
    ) =
        copy(
            customerWithContacts =
                createCustomerWithContactsRequest(customerType, yhteystietoId, *hankekayttajaIds)
        )

    fun JohtoselvityshakemusUpdateRequest.withDescription(description: String) =
        copy(workDescription = description)
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -113,38 +167,287 @@ class HakemusServiceITest : DatabaseTest() {
...
+                        HakemusUpdateRequestFactory
+                            .createJohtoselvityshakemusUpdateRequestFromApplicationEntity(
+                                application
+                            )
+                            .copy(
+                                customerWithContacts =
+                                    HakemusUpdateRequestFactory.createCustomerWithContactsRequest(
+                                        yhteystietoId = requestYhteystietoId
+                                    )
+                            ),
```
corvidian, line 314: I think creating the request should be in the arrange part of the AAA pattern. Especially if creating it is complicated like this. The act part should be simple.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusUpdateRequestFactory.kt
```diff
@@ -25,4 +37,79 @@ object HakemusUpdateRequestFactory {
...
+            maintenanceWork = applicationData.maintenanceWork,
+            propertyConnectivity = applicationData.propertyConnectivity,
+            emergencyWork = applicationData.emergencyWork,
+            rockExcavation = applicationData.rockExcavation ?: false,
+            workDescription = applicationData.workDescription,
+            startTime = applicationData.startTime,
+            endTime = applicationData.endTime,
+            areas = applicationData.areas,
+        )
+    }
```
corvidian, line 97: If you get the HakemusResponse from HakemusService instead of using the entity directly, this can be simplified dramatically.

```suggestion
    fun HakemusResponse.toUpdateRequest(): JohtoselvityshakemusUpdateRequest =
        this.applicationData.toJsonString().parseJson()
```
This might seem like a hack at first glance, but this is exactly what the front end will do when they request an update. FE will read the DTO json, modify it and send the result as an update. So the response and update request need to be JSON-compatible in this direction. If this fails, it's a sign that the FE will fail here as well.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusUpdateRequestFactory.kt
```diff
@@ -25,4 +37,79 @@ object HakemusUpdateRequestFactory {
...
+            maintenanceWork = false,
+            propertyConnectivity = false,
+            emergencyWork = false,
+            rockExcavation = false,
+            workDescription = "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+            startTime = ZonedDateTime.now(TZ_UTC),
+            endTime = ZonedDateTime.ofInstant(Instant.now().plus(5, ChronoUnit.DAYS), TZ_UTC),
+            areas = listOf(ApplicationArea("Hankealue 1", GeometriaFactory.polygon)),
+            customerWithContacts =
+                CustomerWithContactsRequest(
```
corvidian, line 55: This could use `createCustomerWithContactsRequest`.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -113,38 +167,287 @@ class HakemusServiceITest : DatabaseTest() {
...
+            val applicationLogs = auditLogRepository.findByType(ObjectType.APPLICATION)
+            assertThat(applicationLogs).hasSize(originalAuditLogSize)
+        }
+
+        @Test
+        fun `throws exception when there are invalid geometry in areas`() {
+            val application = createHakemus()
+
+            val exception =
+                assertThrows<ApplicationGeometryException> {
```
corvidian, line 244: I think we should stick with one assertion library as much as possible.
```suggestion
                assertFailure {
```

The class of the exception can be checked with `exception.hasClass()`.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -113,38 +167,287 @@ class HakemusServiceITest : DatabaseTest() {
...
+                }
+
+            Assertions.assertEquals(
+                """Invalid geometry received when updating application id=${application.id},
+                | reason = Self-intersection,
+                | location = {"type":"Point","coordinates":[25494009.65639264,6679886.142116806]}"""
+                    .trimMargin()
+                    .replace("\n", ""),
+                exception.message
+            )
```
corvidian, line 263: Here as well it would be nice to stick with assertk.

Check the exception message character-by-character might not be the most productive approach, since it will break easily. Checking that the message contains the important details like the application ID, reason and maybe the coordinates is less fragile and easier to read.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -32,6 +58,264 @@ class HakemusService(
...
+    }
+
+    /** Update the hanke areas based on the update request work areas. */
+    private fun updateHankealueet(hankeEntity: HankeEntity, updateRequest: HakemusUpdateRequest) {
+        val hankealueet =
+            updateRequest.areas?.mapIndexed { i, area ->
+                area.toNewHankealue(i, updateRequest.startTime!!, updateRequest.endTime!!)
+            } ?: emptyList()
+        hankeEntity.alueet.clear()
+        if (hankealueet.isNotEmpty()) {
```
corvidian, line 253: Nothing will happen anyway if hankealueet is empty.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -32,6 +58,264 @@ class HakemusService(
...
+                throw ApplicationGeometryNotInsideHankeException(customMessageOnFailure(area))
+        }
+    }
+
+    /** Update the hanke areas based on the update request work areas. */
+    private fun updateHankealueet(hankeEntity: HankeEntity, updateRequest: HakemusUpdateRequest) {
+        val hankealueet =
+            updateRequest.areas?.mapIndexed { i, area ->
+                area.toNewHankealue(i, updateRequest.startTime!!, updateRequest.endTime!!)
+            } ?: emptyList()
```
corvidian, line 251: This, including `toNewHankealue`, is very close to `HankealueService.createHankealueetFromCableReport`. That could be refactored to take `areas: List<ApplicationArea>?` as its parameter and used in both cases.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -32,6 +58,264 @@ class HakemusService(
...
+                customerWithContacts?.customer?.yhteystietoId != null
+        ) {
+            // new yhteystieto has an id when it shouldn't
+            throw InvalidHakemusyhteystietoException(
+                applicationId,
+                rooli,
+                null,
+                customerWithContacts.customer.yhteystietoId
+            )
+        }
```
corvidian, line 215: I think this would be simpler if the positive cases were handled first. Then all others can throw an exception. The positive cases are:
1. CustomerWithContacts is missing.
2. The customer in customerWithContacts doesn't have an ID.
3. The IDs are the same.
```suggestion
        if (customerWithContacts == null || customerWithContacts.customer.yhteystietoId == null) {
            return
        }
        if (customerWithContacts.customer.yhteystietoId == hakemusyhteystietoEntity?.id) {
            return
        }
        throw InvalidHakemusyhteystietoException(
            applicationId,
            rooli,
            hakemusyhteystietoEntity?.id,
            customerWithContacts.customer.yhteystietoId
        )
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -32,6 +58,264 @@ class HakemusService(
...
+        when (applicationEntity.applicationData) {
+            is CableReportApplicationData ->
+                if (request !is JohtoselvityshakemusUpdateRequest) {
+                    throw IncompatibleHakemusUpdateRequestException(
+                        applicationEntity.id!!,
+                        applicationEntity.applicationData::class,
+                        request::class
+                    )
+                }
+        }
```
corvidian, line 135: This will be easier to extend if checking the condition and throwing the exception are separated. There's also less nesting this way, which is better for our brains.
```suggestion
        val expected =
            when (applicationEntity.applicationData) {
                is CableReportApplicationData -> request is JohtoselvityshakemusUpdateRequest
            }
        if (!expected) {
            throw IncompatibleHakemusUpdateRequestException(
                applicationEntity.id!!,
                applicationEntity.applicationData::class,
                request::class
            )
        }
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
umeetiusbaar: Thank you!

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
umeetiusbaar: You're right. This was also due to copy-pasting from ApplicationServiceITest. Will change.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusUpdateRequestFactory.kt
umeetiusbaar: I do not use HakemusService to create an Entity atm. I use HakemusFactory for that.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
umeetiusbaar: You're right. Thanks!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: Good call, thanks!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: Nicely spotted, thanks!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: I refactored this as you suggested but combined the first two ifs into one.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusUpdateRequestFactory.kt
umeetiusbaar: Changed, thanks!

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusUpdateRequestFactory.kt
umeetiusbaar: But, you're right that the JSON-compatibility is an issue that I need to solve when I work on the Controller. I guess it needs similar @JsonTypeInfo and @JsonSubTypes annotations as are in ApplicationData class. And they need applicationType to be added which I was hoping to awoid since it is not necessary in business-sense. The type is already in the parent Application and copying it to ApplicationData seemed unnecessary. Until I noticed those JSON-annotations.

corvidian: > I do not use HakemusService to create an Entity atm. I use HakemusFactory for that.
You can call HakemusService after the entity has been saved.
```
            val entity = hakemusFactory.builder(USERID).save()
            val hakemus = hakemusService.hakemusResponse(entity.id!!)
```

corvidian: Yes, copying the application type to application data didn't feel like a perfect solution, but I couldn't think of any way around it. We could also solve the JSON subtyping when we start working on kaivuilmoitus, so we have a second type of application.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
corvidian: The version taking a CableReportApplicationData could be left as an overloaded method since it's called from a couple of places already.
```
        fun createHankealueetFromCableReport(
            cableReportData: CableReportApplicationData
        ): List<NewHankealue> =
            createHankealueetFromApplicationAreas(
                cableReportData.areas,
                cableReportData.startTime,
                cableReportData.endTime
            )
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -182,4 +209,236 @@ class HakemusServiceITest(
...
+            val applicationLogs = auditLogRepository.findByType(ObjectType.APPLICATION)
+            assertThat(applicationLogs).hasSize(originalAuditLogSize + 1)
+        }
+
+        private fun createHakemus(
+            hankeEntity: HankeEntity = hankeFactory.builder(USERNAME).withHankealue().saveEntity(),
+            f: ApplicationEntity.() -> Unit = {}
+        ): ApplicationEntity {
+            return hakemusFactory.builder(USERNAME, hankeEntity).withModification(f).save()
+        }
```
corvidian, line 432: I'm not a big fan of adding private factory methods to test classes. Developing the shared factory classes seems more productive in the long run.

Creating a hakemus could be easier by adding a default value for the hankeEntity parameter or the `builder()` method.
```
    fun builder(
        userId: String,
        hankeEntity: HankeEntity = hankeFactory.builder(userId).withHankealue().saveEntity()
    ): HakemusBuilder {
```

Then the simplest way to create a hakemus would be easy enough not to require a helper method.
```
    val application = hakemusFactory.builder(USERNAME).save()
```

The function parameter here is used only once to set the alluId of an application, which could be done with the existing (though private) `.withStatus()` method:
```
    val application = hakemusFactory.builder(USERNAME).withStatus(alluId = 21).save()
```

umeetiusbaar: I refactored the creation of ApplicationEntity.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: Did that.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusUpdateRequestFactory.kt
umeetiusbaar: I refactored the creation of hakemus and request as you suggested.

corvidian (approved):

PR #640, opened 2024-03-05
HAI-2392 Add an API for checking if a hankekayttaja can be deleted

corvidian:
# Description

The API returns names, identifiers and statuses of all applications the hankekayttaja is a contact in, partitioned to active applications and drafts. It also returns a boolean that's true when the hankekayttaja is the only contact in one of the omistaja customers.

Add domain classes for Hakemus and HakemusData. The classes and methods for creating them from entities and in factories are loaned from #637.

Add a new permission code for deleting users. It's given to KAIKKI_OIKEUDET and KAIKKIEN_MUOKKAUS privileges.

## Type of change

- [X] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
The endpoint can be called from e.g. Swagger UI, but there's no method yet 

# Checklist:

 or other location: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8417968152/Uusi+malli+yhteystietoihin+ja+k+ytt+oikeuksiin#Henkil%C3%B6iden-poistaminen-kortistosta

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
@@ -2,36 +2,51 @@ package fi.hel.haitaton.hanke.permissions
 
 import assertk.all
 import assertk.assertThat
+import assertk.assertions.containsExactly
 import assertk.assertions.containsExactlyInAnyOrder
+import assertk.assertions.extracting
 import assertk.assertions.hasSize
 import assertk.assertions.isEmpty
 import assertk.assertions.isEqualTo
+import assertk.assertions.isFalse
 import assertk.assertions.isNotNull
 import assertk.assertions.isNull
+import assertk.assertions.isTrue
 import assertk.assertions.prop
+import assertk.assertions.single
 import fi.hel.haitaton.hanke.ContactType
 import fi.hel.haitaton.hanke.ControllerTest
 import fi.hel.haitaton.hanke.HankeError
 import fi.hel.haitaton.hanke.HankeNotFoundException
 import fi.hel.haitaton.hanke.HankeService
 import fi.hel.haitaton.hanke.IntegrationTestConfiguration
+import fi.hel.haitaton.hanke.allu.ApplicationStatus
 import fi.hel.haitaton.hanke.andReturnBody
+import fi.hel.haitaton.hanke.domain.Hanke
+import fi.hel.haitaton.hanke.domain.Yhteyshenkilo
+import fi.hel.haitaton.hanke.factory.HakemusFactory
 import fi.hel.haitaton.hanke.factory.HankeFactory
+import fi.hel.haitaton.hanke.factory.HankeFactory.Companion.withOmistaja
 import fi.hel.haitaton.hanke.factory.HankeIdentifierFactory
 import fi.hel.haitaton.hanke.factory.HankeKayttajaFactory
+import fi.hel.haitaton.hanke.factory.HankeYhteyshenkiloFactory
 import fi.hel.haitaton.hanke.factory.TestHankeIdentifier
 import fi.hel.haitaton.hanke.factory.identifier
+import fi.hel.haitaton.hanke.hakemus.HakemusService
 import fi.hel.haitaton.hanke.hankeError
 import fi.hel.haitaton.hanke.hasSameElementsAs
 import fi.hel.haitaton.hanke.logging.DisclosureLogService
+import fi.hel.haitaton.hanke.permissions.HankeKayttajaController.DeleteInfo
 import fi.hel.haitaton.hanke.permissions.HankeKayttajaController.Tunnistautuminen
 import fi.hel.haitaton.hanke.permissions.HankeKayttajaController.TunnistautuminenResponse
+import fi.hel.haitaton.hanke.permissions.PermissionCode.DELETE_USER
 import fi.hel.haitaton.hanke.permissions.PermissionCode.EDIT
 import fi.hel.haitaton.hanke.permissions.PermissionCode.MODIFY_EDIT_PERMISSIONS
 import fi.hel.haitaton.hanke.permissions.PermissionCode.MODIFY_USER
 import fi.hel.haitaton.hanke.permissions.PermissionCode.RESEND_INVITATION
 import fi.hel.haitaton.hanke.permissions.PermissionCode.VIEW
 import fi.hel.haitaton.hanke.profiili.VerifiedNameNotFound
+import fi.hel.haitaton.hanke.touch
 import io.mockk.Called
 import io.mockk.checkUnnecessaryStub
 import io.mockk.clearAllMocks
@@ -42,10 +57,15 @@ import io.mockk.verify
 import io.mockk.verifyOrder
 import io.mockk.verifySequence
 import java.util.UUID
+import java.util.stream.Stream
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest
 import org.springframework.context.annotation.Import
@@ -64,13 +84,15 @@ private const val HANKE_TUNNUS = HankeFactory.defaultHankeTunnus
 @Import(IntegrationTestConfiguration::class)
 @ActiveProfiles("test")
 @WithMockUser(USERNAME)
-class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : ControllerTest {
-
-    @Autowired private lateinit var hankeKayttajaService: HankeKayttajaService
-    @Autowired private lateinit var hankeService: HankeService
-    @Autowired private lateinit var permissionService: PermissionService
-    @Autowired private lateinit var disclosureLogService: DisclosureLogService
-    @Autowired private lateinit var authorizer: HankeKayttajaAuthorizer
+class HankeKayttajaControllerITest(
+    @Autowired override val mockMvc: MockMvc,
+    @Autowired private val hankeKayttajaService: HankeKayttajaService,
+    @Autowired private val hankeService: HankeService,
+    @Autowired private val hakemusService: HakemusService,
+    @Autowired private val permissionService: PermissionService,
+    @Autowired private val disclosureLogService: DisclosureLogService,
+    @Autowired private val authorizer: HankeKayttajaAuthorizer,
+) : ControllerTest {
 
     @BeforeEach
     fun clearMocks() {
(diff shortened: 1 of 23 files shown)


services/hanke-service/src/main/resources/db/changelog/changesets/071-grant-delete-user-permission-to-kayttooikeustasot.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+--changeset Topias Heinonen:071-grant-delete-user-permission-to-kayttooikeustasot
+--comment: Grant the MODIFY_USERS permission to all kayttooikeustasot except KATSELUOIKEUS
```
umeetiusbaar, line 3: Is this comment correct in this context?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
```diff
@@ -948,6 +976,179 @@ class HankeKayttajaControllerITest(@Autowired override val mockMvc: MockMvc) : C
...
+            every { hankeService.loadHankeById(hankeId) } returns
+                HankeFactory.create(id = hankeId)
+                    .withOmistaja(1, 1, HankeYhteyshenkiloFactory.create(id = kayttajaId))
+            every { hakemusService.getHakemuksetForKayttaja(kayttajaId) } returns listOf()
+
+            val response: DeleteInfo = get(url).andExpect(status().isOk).andReturnBody()
+
+            assertThat(response).all {
+                prop(DeleteInfo::onlyOmistajanYhteyshenkilo).isTrue()
+                prop(DeleteInfo::activeHakemukset).isEmpty()
```
umeetiusbaar, line 1026: In most of these tests only the thing mentioned in the test name is checked. E.g. onlyOmistajanYhteyshenkilo. Here, also activeHakemukset and draftHakemukset are checked. Should they be checked in every test or removed from here?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationData.kt
```diff
@@ -107,6 +111,29 @@ data class CableReportApplicationData(
...
+            maintenanceWork = maintenanceWork,
+            propertyConnectivity = propertyConnectivity,
+            emergencyWork = emergencyWork,
+            rockExcavation = rockExcavation,
+            workDescription = workDescription,
+            startTime = startTime,
+            endTime = endTime,
+            areas = areas,
+            customerWithContacts =
+                applicationEntity.yhteystiedot[ApplicationContactType.HAKIJA]?.toDomain(),
```
umeetiusbaar, line 129: These need a Hibernate Session (transaction) to work. One cannot access ApplicationEntity.yhteystiedot otherwise. Or at least so I figured after banging my head to that tree for too long.

umeetiusbaar: But now that I check the code further, I see that this is actually used in a `@Transactional HakemusService.getHakemuksetForKayttaja`. So, no problems. But one cannot use this in tests (without session/transaction).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusyhteystietoEntity.kt
```diff
@@ -57,7 +57,28 @@ class HakemusyhteystietoEntity(
...
+            id = id,
+            tyyppi = tyyppi,
+            rooli = rooli,
+            nimi = nimi,
+            sahkoposti = sahkoposti,
+            puhelinnumero = puhelinnumero,
+            ytunnus = ytunnus,
+            yhteyshenkilot =
+                yhteyshenkilot.map { yhteyshenkilo ->
+                    Hakemusyhteyshenkilo(
```
umeetiusbaar, line 71: Here is the same as with ApplicationEntity.yhteystiedot above - a session/transasction is needed. Maybe we could add some comments on these `toDomain()` functions about this?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -111,7 +115,7 @@ class HankeKayttajaController(
     )
     @PreAuthorize("@hankeKayttajaAuthorizer.authorizeKayttajaId(#kayttajaId, 'VIEW')")
     fun getHankeKayttaja(@PathVariable kayttajaId: UUID): HankeKayttajaDto =
-        hankeKayttajaService.getKayttaja(kayttajaId).also {
+        hankeKayttajaService.getKayttaja(kayttajaId).toDto().also {
```
umeetiusbaar, line 118: This the kind of functionality I thought to implement with HakemusService.update: the service could return the full Hakemus domain object and then the Controller would create the dto with only the data needed. Should I refactor my solution into this kind of implementation? It would feel better for me, I think - but then I would need to again have Hakemus which I just deleted yesterday from my branch...

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -434,6 +438,78 @@ Returns the updated hankekayttaja.
...
+                ),
+            ]
+    )
+    @PreAuthorize(
+        "@featureService.isEnabled('USER_MANAGEMENT') && " +
+            "@hankeKayttajaAuthorizer.authorizeKayttajaId(#kayttajaId, 'DELETE_USER')"
+    )
+    fun checkForDelete(@PathVariable kayttajaId: UUID): DeleteInfo {
+        val kayttaja = hankeKayttajaService.getKayttaja(kayttajaId)
+        val hanke = hankeService.loadHankeById(kayttaja.hankeId)!!
```
umeetiusbaar, line 480: This will throw NPE and return 500 error but that is okay, because unless the database is totally bonkers there should always be a Hanke?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
```diff
@@ -434,6 +438,78 @@ Returns the updated hankekayttaja.
...
+
+        return DeleteInfo(activeHakemukset, draftHakemukset, onlyOmistajanYhteyshenkilo)
+    }
+
+    data class DeleteInfo(
+        val activeHakemukset: List<HakemusDetails>,
+        val draftHakemukset: List<HakemusDetails>,
+        val onlyOmistajanYhteyshenkilo: Boolean,
+    ) {
+        data class HakemusDetails(
```
umeetiusbaar, line 498: This is also nicely done using inner class when it is only needed inside the parent.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeKayttajaFactory.kt
```diff
@@ -270,5 +271,23 @@ class HankeKayttajaFactory(
...
         ): List<HankeKayttajaDto> =
             (1..amount).map { createDto(it, roolit = roolit, tunnistautunut = (it % 2 == 0)) }
+
+        private fun createDto(
+            i: Int,
+            id: UUID = UUID.randomUUID(),
+            roolit: List<ContactType> = emptyList(),
+            tunnistautunut: Boolean = false
+        ) =
+            HankeKayttajaDto(
```
umeetiusbaar, line 281: I wonder if it would be better if createDto(id, etunimi, sukunimi, sahkoposti, puhelinnumero, kayttooikeustaso, roolit, tunnistautunut) would be called here instead of constructing HankeKayttajaDto in very similar manner as in that other createDto function?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
corvidian: You can refactor it in the next PR, which would add the controller?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationData.kt
corvidian: In production code, we shouldn't be passing entities over transaction boundaries.

In integration tests, the correct way to get a domain object is to call the service method that loads it from the DB.

In unit tests, there is no DB or entity manager, so there are no lazily loaded values.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
corvidian: I added these here so I wouldn't have to make a new test case that checks these are empty when there are no applications for the user. Integration tests are not free in running time, after all. Would it be better if that was a separate test?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationData.kt
corvidian: That being said, ApplicationData is not an entity, so I don't think it should be taking entities as parameters, either. The border between entities and domain objects is not clear.

I refactored this to take a map of Hankeyhteystieto domain objects instead.

services/hanke-service/src/main/resources/db/changelog/changesets/071-grant-delete-user-permission-to-kayttooikeustasot.sql
corvidian: Sorry, I forgot to edit the comment after copy-pasting it. Thank you for noticing. It should be correct now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusyhteystietoEntity.kt
corvidian: I think one should assume that any method in an entity can cause `LazyInitializationException`s, with the exception of simple values.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaControllerITest.kt
umeetiusbaar: I think this is fine as it is now.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaController.kt
corvidian: I added an elvis operator to throw an exception.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeKayttajaFactory.kt
corvidian: I changed this to use `createDto` instead.

umeetiusbaar (approved):

PR #641, opened 2024-03-05
HAI Allow wildcards in test environment email filters

corvidian:
# Description

Allow wildcards in the list of allowed emails, so it can have an address like `*@hel.fi`. This will reduce the amount of manual maintenance of the filter lists. It will also enable the use of `person+tag@domain.com` type emails for testing multiple users and contacts with a single email account.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/email/FilteredEmailSenderServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/email/FilteredEmailSenderServiceITest.kt
@@ -19,12 +19,12 @@ import org.springframework.test.context.ActiveProfiles
     properties =
         [
             "haitaton.email.filter.use=true",
-            "haitaton.email.filter.allow-list=test@test.test;something@mail.com",
+            "haitaton.email.filter.allow-list=test@test.test;something@mail.com;.*@wildcard.com",
             "haitaton.features.user-management=false"
         ]
 )
 @ActiveProfiles("test")
-class EmailSenderServicePropertiesITest : DatabaseTest() {
+class FilteredEmailSenderServiceITest : DatabaseTest() {
 
     companion object {
         @JvmField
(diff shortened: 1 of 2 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/email/EmailSenderService.kt
```diff
@@ -144,7 +144,8 @@ class EmailSenderService(
             "$templatePath/common/signature-en.mustache".getResourceAsText(),
         )
 
-    private fun emailNotAllowed(email: String) = !emailConfig.filter.allowList.contains(email)
+    private fun emailNotAllowed(email: String) =
+        !emailConfig.filter.allowList.any { it.toRegex().matches(email) }
```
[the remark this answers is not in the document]
corvidian: The full regex was probably a bit too much. I restricted it to just allowing asterisks.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/email/FilteredEmailSenderServiceITest.kt
```diff
@@ -46,19 +46,42 @@ class EmailSenderServicePropertiesITest : DatabaseTest() {
...
+
+    @Test
+    fun `sendJohtoselvitysCompleteEmail blocks send when dot is replaced`() {
+        emailSenderService.sendJohtoselvitysCompleteEmail("test@test-test", 13L, "JS2300001")
+
+        assertThat(greenMail.receivedMessages.size).isEqualTo(0)
+    }
+
+    @Test
+    fun `sendJohtoselvitysCompleteEmail sends email when email matches wildcard`() {
```
umeetiusbaar, line 70: Should there also be a test for email addresses like `test+suffix@wildcard.com`?

corvidian: I added a test.

umeetiusbaar (approved):

PR #644, opened 2024-03-08
HAI Unite username constants in tests

corvidian:
# Description

Almost every integration test file starts with something like:
```
private consta val USERNAME = "test-user"
```

Combine all these different username constants into one constant that's shared between different tests. This means that the user ID of the mocked user will be the same in all tests.

This enables things like using the constant as a default value for factories. This is outside the scope of this PR, however.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar (approved):
A very welcomed refactoring, thank you!

PR #650, opened 2024-03-11
HAI-2393 Refactor DatabaseTest to IntegrationTest

corvidian:
# Description

Add the annotations shared by integration tests to DatabaseTest and rename it to IntegrationTest. The annotations can be removed from the test classes since IntegrationTest adds them. Leave the `@SpringBootTest` annotation to the class, if it modifies system properties. The class-specific annotation overrides the one from the abstract class.

Also, add MockFileClientExtension to IntegrationTest, so the file client is always available. Remove the extension annotation from other places.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/IntegrationTest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/IntegrationTest.kt
@@ -1,6 +1,8 @@
 package fi.hel.haitaton.hanke
 
+import fi.hel.haitaton.hanke.attachment.common.MockFileClientExtension
 import fi.hel.haitaton.hanke.test.USERNAME
+import org.junit.jupiter.api.extension.ExtendWith
 import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.boot.testcontainers.service.connection.ServiceConnection
 import org.springframework.security.test.context.support.WithMockUser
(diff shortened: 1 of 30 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/IntegrationTest.kt
```diff
@@ -26,7 +28,8 @@ import org.testcontainers.utility.MountableFile
 @SpringBootTest
 @ActiveProfiles("test")
 @WithMockUser(USERNAME)
-abstract class DatabaseTest {
+@ExtendWith(MockFileClientExtension::class)
+abstract class IntegrationTest {
```
umeetiusbaar, line 32: Has it been considered that we would use Azurite as a Test Container like we do for the database?

corvidian: We use an Azurite container for `BlobClientITest` to test the client code against a close approximation of the real Blob Storage.

I think the mock implementation is enough for other tests. The difference is that Blob Storage is a lot less intricate than an SQL server, so it's not hard to mock it, while it would be impossible to mock an SQL server.

It might be OK to use Azurite everywhere. It shouldn't slow the tests down when running all tests, but it might slow down running single tests or test classes, though it might not even be noticeable. If I remember correctly, validating the uploaded files was harder with the Azurite approach, which was one of the reasons we use a mock implementation in most tests.

umeetiusbaar: Valid points and no need to change the current implementation. If I remember correctly, there was one issue with Azure Blob Storage that we didn't catch in unit tests. Was it something to do with the scandic letters or other symbols in attachment filenames? Perhaps that sort of errors would be caught using Azurite in a test container. But, no need to anything about it now.

umeetiusbaar (approved):

PR #655, opened 2024-03-14
HAI-2275 Add a new API for creating a stand-alone johtoselvityshakemus

corvidian:
# Description

The API takes in a name for the hakemus (and the generated hanke) and the founders contact info (email and phone number). It creates a hanke with the given name, adds the founder as a Kaikki Oikeudet level hankekayttaja and creates a new johtoselvityshakemus for the hanke with the same name as the hanke.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
It should be possible to create a johtoselvityshakemus from the Swagger UI now.

# Checklist:

 or other location: 

# Other relevant info
Audit logging is added in the next PR (#656).

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
@@ -31,6 +31,7 @@ import fi.hel.haitaton.hanke.allu.ApplicationStatus
 import fi.hel.haitaton.hanke.allu.CableReportService
 import fi.hel.haitaton.hanke.application.ALLU_USER_CANCELLATION_MSG
 import fi.hel.haitaton.hanke.application.Application
+import fi.hel.haitaton.hanke.application.ApplicationData
 import fi.hel.haitaton.hanke.application.ApplicationEntity
 import fi.hel.haitaton.hanke.application.ApplicationGeometryException
 import fi.hel.haitaton.hanke.application.ApplicationRepository
@@ -65,8 +66,12 @@ import fi.hel.haitaton.hanke.factory.HankeYhteyshenkiloFactory
 import fi.hel.haitaton.hanke.factory.HankeYhteystietoFactory
 import fi.hel.haitaton.hanke.factory.HankealueFactory
 import fi.hel.haitaton.hanke.factory.ProfiiliFactory
+import fi.hel.haitaton.hanke.factory.ProfiiliFactory.DEFAULT_GIVEN_NAME
+import fi.hel.haitaton.hanke.factory.ProfiiliFactory.DEFAULT_LAST_NAME
 import fi.hel.haitaton.hanke.factory.TEPPO_TESTI
 import fi.hel.haitaton.hanke.geometria.Geometriat
+import fi.hel.haitaton.hanke.hakemus.Hakemus
+import fi.hel.haitaton.hanke.hakemus.HakemusData
 import fi.hel.haitaton.hanke.logging.AuditLogEntryEntity
 import fi.hel.haitaton.hanke.logging.AuditLogRepository
 import fi.hel.haitaton.hanke.logging.AuditLogTarget
@@ -325,8 +330,8 @@ class HankeServiceITests(
             val hankeKayttajat = hankekayttajaRepository.findAll()
             assertThat(hankeKayttajat).single().all {
                 prop(HankekayttajaEntity::hankeId).isEqualTo(returnedHanke.id)
-                prop(HankekayttajaEntity::etunimi).isEqualTo(ProfiiliFactory.DEFAULT_GIVEN_NAME)
-                prop(HankekayttajaEntity::sukunimi).isEqualTo(ProfiiliFactory.DEFAULT_LAST_NAME)
+                prop(HankekayttajaEntity::etunimi).isEqualTo(DEFAULT_GIVEN_NAME)
+                prop(HankekayttajaEntity::sukunimi).isEqualTo(DEFAULT_LAST_NAME)
                 prop(HankekayttajaEntity::puhelin).isEqualTo(DEFAULT_HANKE_PERUSTAJA.puhelinnumero)
                 prop(HankekayttajaEntity::sahkoposti).isEqualTo(DEFAULT_HANKE_PERUSTAJA.sahkoposti)
                 prop(HankekayttajaEntity::kutsuttuEtunimi).isNull()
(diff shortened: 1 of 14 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -1201,6 +1204,90 @@ class HankeServiceITests(
...
+        @Test
+        fun `generates hankekayttaja for founder`() {
+            val request = CreateHankeRequest(hakemusNimi, DEFAULT_HANKE_PERUSTAJA)
+
+            val hakemus =
+                hankeService.generateHankeWithJohtoselvityshakemus(request, setUpProfiiliMocks())
+
+            val hanke = hankeRepository.findByHankeTunnus(hakemus.hankeTunnus)!!
+            val kayttajat = hankekayttajaRepository.findByHankeId(hanke.id)
+            assertThat(kayttajat.first()).all {
```
umeetiusbaar, line 1260: Does using `first()` instead of `single()` implicate that there are more than one hankekayttaja created? How do we then know for sure that the founder is the first in result list?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
```diff
@@ -1201,6 +1204,90 @@ class HankeServiceITests(
...
+                prop(HankekayttajaEntity::id).isNotNull()
+                prop(HankekayttajaEntity::sahkoposti).isEqualTo(DEFAULT_HANKE_PERUSTAJA.sahkoposti)
+                prop(HankekayttajaEntity::puhelin).isEqualTo(DEFAULT_HANKE_PERUSTAJA.puhelinnumero)
+                prop(HankekayttajaEntity::etunimi).isEqualTo(ProfiiliFactory.DEFAULT_GIVEN_NAME)
+                prop(HankekayttajaEntity::sukunimi).isEqualTo(ProfiiliFactory.DEFAULT_LAST_NAME)
```
umeetiusbaar, line 1265: Maybe these constants from `ProfiiliFactory` could be in import statements to make it a bit readable here?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
```diff
@@ -4,41 +4,92 @@ import fi.hel.haitaton.hanke.allu.ApplicationStatus
...
 
 data class JohtoselvityshakemusData(
+    override val applicationType: ApplicationType = ApplicationType.CABLE_REPORT,
     override val name: String,
     override val postalAddress: PostalAddress? = null,
-    val constructionWork: Boolean? = null,
-    val maintenanceWork: Boolean? = null,
-    val propertyConnectivity: Boolean? = null,
-    val emergencyWork: Boolean? = null,
+    val area: Double? = null,
```
umeetiusbaar, line 51: Do we need `area`?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -54,6 +55,37 @@ class HakemusService(
         )
     }
 
+    /** Create a johtoselvitys from a hanke that was just created. */
+    @Transactional
```
umeetiusbaar, line 59: I am not sure, but should this be `@Transactional(propagation = Propagation.MANDATORY)` because we pass an entity as a parameter?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankeService.kt
```diff
@@ -98,6 +101,21 @@ class HankeService(
         )
     }
 
+    /**
+     * Create application when no existing hanke. Autogenerates hanke and applies application to it.
+     */
+    @Transactional
+    fun generateHankeWithJohtoselvityshakemus(
+        request: CreateHankeRequest,
+        securityContext: SecurityContext,
```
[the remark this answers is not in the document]
corvidian: We also need the user's access token to get the verified name of the founder.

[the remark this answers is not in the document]
umeetiusbaar: Ah, yes. That's true.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
corvidian: Yes, using `single()` would be better here. I copied this from the old test and didn't notice this.

I changed this to `single()`. Thank you for the recommendation.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
corvidian: I imported the constants 'staticly'.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
corvidian: Good call. I don't think we need it here. Removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
corvidian: I think this would stop us from testing this in a separate integration test, or at least it would be a lot harder.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusResponse.kt
```diff
@@ -27,7 +27,7 @@ sealed interface HakemusDataResponse {
     val postalAddress: PostalAddress?
     val startTime: ZonedDateTime?
     val endTime: ZonedDateTime?
-    val areas: List<ApplicationArea>?
+    val areas: List<ApplicationArea>
```
corvidian, line 30: The UI seems to expect this array to be present, so I changed it to be non-nullable.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
corvidian: Adding `@Transactional` to test methods or classes means that the tests might not hit the DB at all. Or at least until the test and all its checks are completed. It's possible to start and stop the transaction manually with 

I'm not sure how necessary it is to have an existing transaction when calling the method. I mean, the service tests save the entity without a transaction and then call this method just fine. I tried to make this more explicit by creating a new hanke with just the mandatory fields and definitely without any magic, and the hakemus was created successfully.

```kotlin
val hanke = hankeFactory.saveMinimal(nimi = hakemusNimi)
val otherHanke =
    HankeEntity(id = hanke.id, nimi = hanke.nimi, hankeTunnus = hanke.hankeTunnus)

hakemusService.createJohtoselvitys(otherHanke, USERNAME)

assertThat(applicationRepository.findAll()).single().all {
    ...
}
```
I think if we follow the basic guideline of "Don't return entities from a transaction", this should be completely safe. This guideline is one of the reasons `createJohtoselvitys` returns a Hakemus, and not the ApplicationEntity it created.

umeetiusbaar: Ok, thank you for the clarification.

umeetiusbaar (approved):

PR #658, opened 2024-03-18
HAI-1539 kaivuilmoitus hakemus request and response

umeetiusbaar:
# Description

* Support for new ApplicationData type in new Hakemus request/response/service/controller.
* Validation for the new type

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
@@ -49,8 +49,10 @@ import fi.hel.haitaton.hanke.factory.HakemusFactory
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.toUpdateRequest
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withArea
+import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withAreas
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withContractor
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withCustomer
+import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withCustomerWithContactsRequest
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withWorkDescription
 import fi.hel.haitaton.hanke.factory.HankeFactory
 import fi.hel.haitaton.hanke.factory.HankeKayttajaFactory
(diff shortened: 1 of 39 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -72,380 +72,749 @@ class HakemusServiceITest(
 ) : IntegrationTest() {
 
     @Nested
-    inner class HakemusResponse {
-        @Test
-        fun `when application does not exist should throw`() {
-            assertThat(applicationRepository.findAll()).isEmpty()
+    inner class Common {
```
corvidian, line 75: When doing refactoring that changes entire files, it would be nice to do the refactoring separately without any other changes first. This reduces the changes per PR and makes it easy to see what meaningful changes there are. Now the only way to see it is to compare the files line-by-line.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -72,380 +72,749 @@ class HakemusServiceITest(
...
 
-            val exception = assertFailure { hakemusService.hakemusResponse(1234) }
+        @Nested
+        inner class HakemusResponse {
```
corvidian, line 78: I think having all the tests for a single method in the same place is better. So, like:

```
@Nested
inner class HakemusResponse {
    @Test
    fun `when application does not exist should throw`() {
    ...

    @Nested
    inner class WithJohtoselvityshakemus {
        @Test
        fun `returns yhteystiedot and yhteyshenkilot if they're present`() {
    ...
    }

    @Nested
    inner class WithKaivuilmoitus {
        @Test
        fun `returns yhteystiedot and yhteyshenkilot if they're present`() {
    ...
    }
}
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -72,380 +72,749 @@ class HakemusServiceITest(
...
-        fun `throws exception when the application does not exist`() {
-            assertThat(applicationRepository.findAll()).isEmpty()
-            val request =
-                HakemusUpdateRequestFactory.createFilledJohtoselvityshakemusUpdateRequest()
-
-            val exception = assertFailure { hakemusService.updateHakemus(1234, request, USERNAME) }
-
-            exception.hasClass(ApplicationNotFoundException::class)
-        }
+    inner class Kaivuilmoitushakemus {
```
corvidian, line 474: The application is called just `Kaivuilmoitus`. Legally, it's not an application, but more like a notification.

umeetiusbaar: Is `Johtoselvityshakemus` then just `Johtoselvitys`? It has sounded a bit redundant in my ears to have both substantives in the name. `Johtoselvitys` and `Kaivuilmoitus` sound more coherent.

umeetiusbaar: And does the same apply to `CableReportApplication`? Should there be just `Application` <-- `CableReport` / `ExcavationNotification` (and maybe later `PlacementContract`) and same with `Hakemus` <-- `Johtoselvitys` / `Kaivuilmoitus` (and maybe later `Sijoitussopimus`)?

corvidian: Johtoselvitys (cable report) and kaivuilmoitus (excavation notification) are not comparable terms in the application process.

Johtoselvitys is the map/document that tells the digger where the cables are underground. It's the "decision" of the process. The contractor asks for the city to produce that document. They send a johtoselvityshakemus, with which they order the johtoselvitys. The semantically correct name might be "johtoselvitystilaus", but everyone talks about applications, even the [city itself](https://www.suomi.fi/palvelut/verkkoasiointi/tilaa-johtoselvitys-tilaa-johtoselvitys-helsingin-kaupunki-kaupunkiympariston-toimiala/959df9bd-0801-4be9-9535-1b1fcdd4adc2).

Kaivuilmoitus is the application itself, which the contractor sends to the city. The city decides if the applicant has given all of the necessary information and makes a decision document. I don't think there's a special name for the decision document, so we can refer to it as a päätös, maybe kaivuilmoituspäätös or päätös kaivuilmoituksesta or kaivuilmoituksen päätös or something when we need to be more specific.

I haven't looked at placement contracts in any detail, but I think they're comparable to cable reports, i.e. sijoitussopimus is the "decision" of the process.

umeetiusbaar: Thank you for taking time to clarify this for me.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
umeetiusbaar: Sorry, I am still struggling sometimes to create a decent size PRs. I will try to do better in the future.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
umeetiusbaar: Refactored as suggested.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
umeetiusbaar: Though it still seems that all of these different types are actually applications. Even with excavation notification (or announcement or notice or what ever), it is not just a statement that "I will dig here next week, live with it" but more like an application for a permit to dig. As you said yourself, the "**applicant**" creates the excavation notification. The result of this excavation notification would be an excavation permit. Or in Finnish it would go like "Johtoselvityshakemus" (application) --> "Johtoselvitys" (result), "Kaivuilmoitus" --> "Kaivulupa" and "Sijoitushakemus" --> "Sijoitussopimus" (maybe, not sure about this, in someplaces it is called "sijoituslupa"). But you're right about the term "Kaivuilmoitushakemus", that is wrong. The applicant does not want "kaivuilmoitus" but "kaivulupa" or "kaivulupapäätös" or something.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/ApplicationValidator.kt
```diff
@@ -19,8 +19,7 @@ class ApplicationValidator : ConstraintValidator<ValidApplication, BaseApplicati
         val result =
             when (val data = application.applicationData) {
                 is CableReportApplicationData -> validate { data.validateForErrors() }
-                is ExcavationNotificationApplicationData ->
-                    TODO("Excavation notification not implemented")
+                is ExcavationNotificationData -> validate { data.validateForErrors() }
```
corvidian, line 22: Do we support excavation notifications somewhere where this validation is useful?

umeetiusbaar: Should I leave the actual validation code or delete that too? It will be under hakemus package in the future and you are atm doing that.

corvidian: I think we'll still need support for excavation notices for `ensureValidForSend()`, so the validation code shouldn't be removed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -13,6 +13,7 @@ private val logger = KotlinLogging.logger {}
 
 private val businessIdRegex = "^(\\d{7})-(\\d)\$".toRegex()
 private val businessIdMultipliers = listOf(7, 9, 10, 5, 8, 4, 2)
+private val ovtRegex = "^0037(\\d{8})[|\\w]{0,5}\$".toRegex()
```
corvidian, line 16: Why are `|` and `_` allowed but no other punctuation? Should the identifying part also be digits? Is there a source for the definition?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -120,3 +121,32 @@ fun String?.isValidBusinessId(): Boolean {
...
+        logger.warn { "OVT is null." }
+        return false
+    }
+
+    val matchResult = ovtRegex.find(this)
+    if (matchResult == null) {
+        logger.warn { "Invalid format." }
+        return false
+    }
+    if (businessId != null && matchResult.groups[1]!!.value != businessId.replace("-", "")) {
```
corvidian, line 146: Is it guaranteed that the OVT has the business ID of the company? So it's not possible to use a billing processor, or if a billing processor is used, the OVT is still for the main company?

corvidian: Looks like ProCounter only accepts numbers as the OVT.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
corvidian: Looks like ProCounter only accepts numbers as the OVT.

https://support.procountor.fi/hc/fi/articles/360000256417-Laskuaineiston-siirtotiedosto
Laskutietue -> Nro. 38

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
```diff
@@ -120,3 +121,32 @@ fun String?.isValidBusinessId(): Boolean {
         false
     }
 }
+
+/**
+ * Valid OVT ("verkkolaskuosoitteen vastaanottajan identifioiva osa") requirements:
+ * 1. format 0037NNNNNNNNXXXXX, where 0037 = "Suomen verohallinnon tunnus", NNNNNNNN = business id
```
corvidian, line 127: Are all the companies using OVT Finnish companies?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
umeetiusbaar: That sounds more accurate than the one I used (https://support.maventa.fi/fi/support/solutions/articles/77000468510-mik%C3%A4-on-ovt-tunnus-verkkolaskuoperaattori-ja-verkkolaskuosoite-) which only talks about "characters" in ovt.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
umeetiusbaar: Good point! I think I made unnecessary assumptions in my implementation. Maybe we need only to check that the length is correct (12-17 digits)?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusResponse.kt
```diff
@@ -29,7 +31,7 @@ sealed interface HakemusDataResponse {
     val areas: List<ApplicationArea>
     val customerWithContacts: CustomerWithContactsResponse?
 
-    fun customersByRole(): Map<ApplicationContactType, CustomerWithContactsResponse>
+    fun customersByRole(): List<Pair<ApplicationContactType, CustomerWithContactsResponse>>
```
corvidian, line 34: Why not a map?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
corvidian: Based on the Slack conversation, the only check could be that it's at least 12 characters long.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusResponse.kt
corvidian: A map is faster to understand while reading the code. And here it makes perfect sense since there can be only one customer per role.

umeetiusbaar: Changed it back to a map.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/Utils.kt
umeetiusbaar: I changed the validation to only check the length of the string.

corvidian (approved):

PR #659, opened 2024-03-19
HAI-2301 Remove HakemusyhteystietoBuilder

corvidian:
# Description

Remove HakemusyhteystietoBuilder and move the necessary methods for adding yhteystiedot to HakemusBuilder. Building the yhteystiedot for a hakemus can be done before saving the hakemus, as long as the yhteyshenkilot are explicitly saved during `save()`.

The yhteyshenkilot need to be saved explicitly, because `HankeyhteystietoEntity` only has `REMOVE` cascade for the `Hankeyhteyshenkilo` list. It can't be sent to `ALL`, because that would break removing `HankekayttajaEntity`, possible because it's referenced from both `HankekayttajaEntity` and `HankeyhteystietoEntity`, but I'm not sure.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar:
> The yhteyshenkilot need to be saved explicitly, because HankeyhteystietoEntity only has REMOVE cascade for the Hankeyhteyshenkilo list. It can't be sent to ALL, because that would break removing HankekayttajaEntity, possible because it's referenced from both HankekayttajaEntity and HankeyhteystietoEntity, but I'm not sure.

I do not fully understand this. 

Currently `HakemusyhteystietoEntity.yhteyshenkilot` has cascade type `REMOVE` but `HankeYhteystietoEntity.yhteyshenkilot` has cascade `ALL`. There are no cascade settings in either `HakemusyhteyshenkiloEntity` or `HankeYhteyshenkiloEntity`.  Why cannot `HakemusyhteystietoEntity.yhteyshenkilot` have cascade `ALL` as `HankeYhteystietoEntity.yhteyshenkilot` have?

umeetiusbaar:
In database there are a bit different cascading settings in these tables:
```
create table if not exists public.hankeyhteystieto
(
    id               integer generated by default as identity
        primary key,
    contacttype      text    not null,
    hankeid          integer not null
        constraint fk_hanke_hankeyhteystieto
            references public.hanke
            on delete cascade,
...
);

create table if not exists public.hankeyhteyshenkilo
(
    id                  uuid                                               not null
        primary key,
    hankekayttaja_id    uuid                                               not null
        constraint fk_hankeyhteyshenkilo_hankekayttaja
            references public.hankekayttaja
            on delete cascade,
    hankeyhteystieto_id integer                                            not null
        constraint fk_hankeyhteyshenkilo_hankeyhteystieto
            references public.hankeyhteystieto
            on delete cascade,
    created_at          timestamp with time zone default CURRENT_TIMESTAMP not null
);

create table if not exists public.hakemusyhteystieto
(
    id             uuid    not null
        primary key,
    application_id bigint  not null
        constraint fk_hakemusyhteystieto_applications
            references public.applications,
...
);

create table if not exists public.hakemusyhteyshenkilo
(
    id                    uuid    not null
        primary key,
    hakemusyhteystieto_id uuid    not null
        constraint fk_hakemusyhteyshenkilo_hakemusyhteystieto
            references public.hakemusyhteystieto,
    hankekayttaja_id      uuid    not null
        constraint fk_hakemusyhteyshenkilo_hankekayttaja
            references public.hankekayttaja,
    tilaaja               boolean not null
);

```

If these were more uniform, would it be possible then to use the same cascade `ALL` for HankeYhteystietoEntity.yhteyshenkilot` and `HakemusyhteystietoEntity.yhteyshenkilot`?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
```diff
@@ -187,13 +187,14 @@ class KortistoGdprServiceITest(
                     ytunnus = "3213212-0",
                 )
             val hanke = hankeFactory.saveMinimal(generated = true)
```
umeetiusbaar, line 189: Not actually part of this PR, but it caught my eye: `HankeFactory.saveMinimal` does not create a `HankeKayttaja` as e.g. `HankeFactoty.saveWithAlue(USERNAME)` does. Should we provide this kind of Hanke creation that does not create a `HankeKayttaja` at all but instead make `saveMinimal` to create `HankeKayttaja`? 

I made this comment on this PR because I wondered why there are different ways to create `hanke` and `kayttaja` in these tests here.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
```diff
@@ -240,8 +241,8 @@ class KortistoGdprServiceITest(
         fun `returns true when there are only draft applications`() {
             val hanke = hankeFactory.saveWithAlue(USERNAME)
             val kayttaja = hankeKayttajaService.getKayttajaByUserId(hanke.id, USERNAME)!!
```
umeetiusbaar, line 243: Also not really part of this PR, but as we discussed about service functions using entities as parameters, you mentioned that we should not return an entity from a service method. Maybe at some point we will refactor these `HankeKayttajaService` functions that currently return entities?

corvidian:
> Why cannot `HakemusyhteystietoEntity.yhteyshenkilot` have cascade `ALL` as `HankeYhteystietoEntity.yhteyshenkilot` have?

If I change `HakemusyhteystietoEntity.yhteyshenkilot`to have `cascade = [CascadeType.ALL]`, `HankekayttajaDeleteServiceITest.Delete#delete the user when there are no violations` gives me an error:
```
ERROR: update or delete on table "hankekayttaja" violates foreign key constraint "fk_hakemusyhteyshenkilo_hankekayttaja" on table "hakemusyhteyshenkilo"
  Detail: Key (id)=(f0e8efa8-28dc-494c-b1ed-04766a419895) is still referenced from table "hakemusyhteyshenkilo".
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
corvidian: I think that one needs to return an entity since it's needed as a reference in other tables. The alternative would be to copy-paste the code to the other services that need it, which wouldn't be ideal.

Maybe it would be useful to set the `@Transactional` propagation to `MANDATORY` here?

The uses in tests seem to all be for getting the founder of a hanke, so there could be a custom test method for that.

But I'm not sure how to test `getKayttajaByUserId` itself since using `@Transactional` in tests has big caveats.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
corvidian: Sometimes we want to create minimalistic data for tests. Like, more minimalistic than could be created with production code. In some cases, we need to test what happens when there is no user, and in others, it's clearer if the user is created separately. Having a convenient way to create those is better than forcing every test that needs something like that to call the repository directly.

corvidian:
> If these were more uniform, would it be possible then to use the same cascade `ALL` for `HankeYhteystietoEntity.yhteyshenkilot` and `HakemusyhteystietoEntity.yhteyshenkilot`?

I added the `ON DELETE CASCADE`s and it fixed the errors with using `ALL` cascade for `HakemusyhteystietoEntity.yhteyshenkilot`. Thank you for the suggestion. It didn't occur to me to look at the database cascades.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/KortistoGdprServiceITest.kt
umeetiusbaar: Maybe we don't change anything for this now. Just have to remember that in some cases returning Entity is a valid option. But usually not and it needs some consideration whether to use it or not.

umeetiusbaar (approved):

PR #661, opened 2024-03-21
HAI-2253 Add tests to reproduce the bugs

corvidian:
# Description

Two bugs were discovered in updating yhteyshenkilot in applications:

1. If adding a new yhteystieto and yhteyshenkilo at the same time, the backend throws an exception: `Unable to find fi.hel.haitaton.hanke.hakemus.HakemusyhteystietoEntity with id ca2105e8-2296-4a42-b1f4-6396050743ee`.

2. If removing all yhteyshenkilot from a yhteystieto, the response has an empty contact list, but the next GET still has the yhteyshenkilot. I.e. clearing the contact list is not persisted.

Add tests that reproduce these bugs.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar:
Added the bug fixes also in this PR and made it mergeable to branch https://github.com/City-of-Helsinki/haitaton-backend/tree/HAI-2301/remove-hakemusyhteystietobuilder

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusUpdateRequestFactory.kt
```diff
@@ -94,6 +94,16 @@ object HakemusUpdateRequestFactory {
                 createCustomerWithContactsRequest(type, yhteystietoId, *hankekayttajaIds)
         )
 
+    fun JohtoselvityshakemusUpdateRequest.withContractorWithContactsRequest(
```
corvidian, line 97: Longer names don't always mean clearer code. There's only one thing `withContractor` can do. Making the name longer makes it harder to parse without any added benefit.

Besides, this sets the `contractorWithContacts` field, not a `contractorWithContactsRequest` field.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -405,6 +406,7 @@ class HakemusService(
                 ytunnus = customer.registryKey,
                 application = applicationEntity,
             )
+            .also { hakemusyhteystietoRepository.save(it) }
```
corvidian, line 409: I get the same results when deleting both explicit saves.

umeetiusbaar (approved):

PR #667, opened 2024-03-25
HAI-2301 Add PDF generation

corvidian:
# Description

Add creating a PDF from the hakemus form data. The created PDF is identical to how it looked before Kortisto.

Create a separate HakemusPdfService for generating the PDF. It's very similar to ApplicationPdfService, but it can be removed soon when purging the old Application package.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
@@ -24,7 +24,6 @@ import assertk.assertions.single
 import fi.hel.haitaton.hanke.HankeNotFoundException
 import fi.hel.haitaton.hanke.HankeRepository
 import fi.hel.haitaton.hanke.IntegrationTest
-import fi.hel.haitaton.hanke.allu.AlluException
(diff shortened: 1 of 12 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -817,6 +825,28 @@ class HakemusServiceITest(
             }
         }
 
+        @Test
+        fun `saves the Allu ID and status even when sending the form data attachment fails`() {
+            val hakemus = hakemusFactory.builder().withMandatoryFields().save()
+            every { cableReportService.create(any()) } returns alluId
+            every { cableReportService.addAttachment(alluId, any()) } throws AlluException(listOf())
```
umeetiusbaar, line 832: Maybe `AlluException` should have a default (empty) value for the list to make this just `throws AlluException()`

corvidian: AlluException is only used in tests, so I moved it to test code. The parameter was never used, so I removed it.

umeetiusbaar (approved):

PR #676, opened 2024-04-04
HAI-2375 Improve logging of hakemus ID's

corvidian:
# Description

We have three different IDs for a hakemus: there's our database ID (`id`), Allu's ID (`alluid`) and the official identifier of the hakemus (`applicationIdentifier`). When logging application events, different combinations of these have been arbitrarily added to log rows. There has also been some variance in the formatting of the ID fields.

Add a new interface that groups these three values, and different views of applications can implement that. The interface provides a way to log all three IDs uniformly. It also provides a simple way of passing all the values needed for logging them in a single value.

Implement the interface for Hakemus and ApplicationEntity. Change all the logging where the logging parameter was calling the `id` field directly to use the new identifier. Just the ID is still logged in some places where e.g. the ID is passed as a method parameter. Also, in some places, logging all three IDs made no sense, since `alluid` and `applicationIdentifier` are always null.

Make the ApplicationEntity `id` field to be non-nullable. It's already `NOT NULL` in the database, so this only affects creating new applications. The `id` field is a primitive value, so the default value of `0` is considered to be empty, which is used as a placeholder before the application has been saved.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
@@ -97,7 +97,7 @@ class HakemusService(
             )
         val entity =
             ApplicationEntity(
-                id = null,
+                id = 0,
                 alluid = null,
                 alluStatus = null,
                 applicationIdentifier = null,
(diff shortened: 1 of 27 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -126,7 +126,7 @@ class HakemusService(
         }
 
         assertGeometryValidity(request.areas) { validationError ->
-            "Invalid geometry received when updating application id=${applicationEntity.id}, reason=${validationError.reason}, location=${validationError.location}"
+            "Invalid geometry received when updating application ${applicationEntity.logString()}, reason=${validationError.reason}, location=${validationError.location}"
```
umeetiusbaar, line 129: There are still some "application" among "hakemus" strings in log messages.

corvidian: I changed every log message in HakemusService to hakemus instead of application.

umeetiusbaar (approved):

PR #681, opened 2024-04-10
HAI-2370 Make kayttaja/deleteInfo return every hakemus just once

corvidian:
# Description

Currently, when the hankekayttaja is a contact in several roles of the same application, the delete info endpoint returns copies of those applications. This is not correct, and shows up as duplicates to the user.

Return each distinct application just once.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
1. When making johtoselvityshakemus, create a new contact and add it to several roles.
2. Save and go to user management.
3. Click on remove user for the user you created earlier.
4. See that the application is mentioned only once in the confirmation popup.

# Checklist:

 or other location:


umeetiusbaar (approved):

PR #684, opened 2024-04-12
HAI-2231 Add code for migrating a single hanke

corvidian:
# Description
Add code for migrating a single hanke.

Still needs:
- Logging
- Starting the migration at startup or with a timer

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Until there's a trigger, this can only be tested with integration and unit tests-

# Checklist:

 or other location: 

# Other relevant info
I'll add logging in a separate PR since this is getting very long.
The trigger for running the migration, which loads all hanke IDs and migrates them one-by-one will be added later.

### DIFF
--- a/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
+++ b/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
@@ -1,0 +1,238 @@
+package fi.hel.haitaton.hanke.hakemus
+
+import assertk.assertThat
+import assertk.assertions.isEmpty
+import assertk.assertions.isEqualTo
+import assertk.assertions.isNotNull
+import assertk.assertions.isNull
+import assertk.assertions.isSameInstanceAs
+import fi.hel.haitaton.hanke.application.Contact
+import fi.hel.haitaton.hanke.factory.ApplicationFactory
+import fi.hel.haitaton.hanke.factory.ApplicationFactory.Companion.withContact
+import fi.hel.haitaton.hanke.factory.ApplicationFactory.Companion.withContacts
+import fi.hel.haitaton.hanke.hakemus.HakemusMigrationService.Companion.contactWithLeastMissingFields
+import fi.hel.haitaton.hanke.hakemus.HakemusMigrationService.Companion.defaultIfNullOrBlank
+import fi.hel.haitaton.hanke.hakemus.HakemusMigrationService.Companion.mode
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
+import org.junit.jupiter.params.provider.NullSource
+import org.junit.jupiter.params.provider.ValueSource
+
+class HakemusMigrationServiceTest {
+
+    @Nested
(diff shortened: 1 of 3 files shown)


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
```diff
@@ -0,0 +1,238 @@
...
+                    ApplicationFactory.createContact(
+                        firstName = "Hannu",
+                        lastName = null,
+                        email = " "
+                    )
+                )
+
+            val response = contacts.contactWithLeastMissingFields()
+
+            assertThat(response).isNotNull().isSameInstanceAs(contacts[1])
```
umeetiusbaar, line 117: So, it is Pekka? Reading just this test doesn't make it very clear how we have came into that conclusion. The names have some priority over phone? None of the contacts have email address, should there be such a test object too?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
```diff
@@ -0,0 +1,238 @@
...
+            val contacts = arrayOf(nonOrderer, missingTwo, firstMissingOne, secondMissingOne)
+            val data =
+                ApplicationFactory.createCableReportApplicationData(
+                    customerWithContacts =
+                        ApplicationFactory.createCompanyCustomer().withContacts(*contacts),
+                )
+
+            val result = HakemusMigrationService.findOrderer(data)
+
+            assertThat(result).isSameInstanceAs(firstMissingOne)
```
umeetiusbaar, line 179: This is more clear than the previous test with the differences between test objects.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
umeetiusbaar: Well, nevermid, this is not so important and you can ignore this. I thought that starting reading from the tests would make the logic more clear but it turned out to be the other way round.

umeetiusbaar (approved):

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
corvidian: I changed the test to create a `Contact` directly, without using the factory method. I think this makes it easier to see which fields are missing and which are not. I also made the contacts named values, to give them some context.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
corvidian: I changed the test to create a `Contact` directly, without using the factory method. I think this makes it easier to see which fields are missing and which are not.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusMigrationServiceTest.kt
umeetiusbaar: Thank you!

umeetiusbaar (approved):

PR #686, opened 2024-04-16
HAI-2468 Add support for disabling the API

umeetiusbaar:
# Description

New parameter `haitaton.api.disabled` which controls whether the API is blocked or not. When the API is blocked all calls to other endpoints than listed below will give 503 error but only after the user is logged in first. If the user is not logged in then a normal 401 error will be given (if the endpoint needs authentication).

Also, added a new `getRaw` function in `ContollerTest` that does not assume a response of certain content type and refactored calls like `get(url, resultType = MediaType.APPLICATION_PDF)` to be simpler: `get(url, MediaType.APPLICATION_PDF)`.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Set `HAITATON_API_DISABLED=true` and run the application. All calls to other paths than listed here should give 503 error (when the user is authenticated):
```
/actuator/**
/status
/swagger-ui.html
/swagger-ui/**
/v3/api-docs**
/profiili/**
```

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/ControllerTest.kt
```diff
@@ -17,7 +17,14 @@ interface ControllerTest {
     /** Send a GET request to the given URL. */
     fun get(
         url: String,
-        resultType: MediaType = MediaType.APPLICATION_JSON,
+    ): ResultActions {
```
corvidian, line 20: This will stop almost all current controller tests from checking that the Content-Type header is set correctly.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/ApiDisabledConfiguration.kt
```diff
@@ -0,0 +1,56 @@
...
+@Configuration
+@ConditionalOnProperty(
+    name = ["haitaton.api.enabled"],
+    havingValue = "false",
+    matchIfMissing = false
+)
+class ApiDisabledConfiguration {
+
+    companion object {
+        val BLOCKED_PATHS =
```
corvidian, line 26: It would be better to whitelist the allowed paths. I don't expect anyone to remember to update this list when adding new endpoints.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/ControllerTest.kt
umeetiusbaar: Added `getRaw` instead of changing this method.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/configuration/ApiDisabledConfiguration.kt
umeetiusbaar: I did that. Even though when adding new endpoints the blocked/not-blocked check has to be done anyway.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilter.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    companion object {
+        val ALLOWED_PATHS =
+            setOf(
+                "/actuator/.*",
+                "/status",
+                "/swagger-ui.html",
+                "/swagger-ui/.*",
+                "/v3/api-docs.*",
+                "/profiili/.*",
```
corvidian, line 48: Why are these allowed? Especially `/profiili`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilter.kt
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.servlet.FilterChain
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import mu.KotlinLogging
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.core.annotation.Order
+import org.springframework.stereotype.Component
+import org.springframework.web.filter.OncePerRequestFilter
+
+private val log = KotlinLogging.logger {}
```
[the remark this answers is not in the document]
umeetiusbaar: `logger` is already in super-class `GenericFilterBean`. But perhaps we can use it?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilter.kt
umeetiusbaar: What should we allow?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/ControllerTest.kt
```diff
@@ -24,6 +24,13 @@ interface ControllerTest {
             .andExpect(content().contentType(resultType))
     }
 
+    /** Send a GET request to the given URL without expecting content of certain type. */
+    fun getRaw(
```
corvidian, line 28: `get` could also be modified to accept `null`.

```kotlin
    fun get(
        url: String,
        resultType: MediaType? = MediaType.APPLICATION_JSON,
    ): ResultActions {
        val actions =
            mockMvc.perform(MockMvcRequestBuilders.get(url).accept(MediaType.APPLICATION_JSON))

        return if (resultType != null) {
            actions.andExpect(content().contentType(resultType))
        } else {
            actions.andExpect(header().doesNotExist("content-type"))
        }
    }
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilterActuatorITests.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.security.test.context.support.WithAnonymousUser
+import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+
+@SpringBootTest(
+    properties = ["haitaton.api.disabled=true", "management.server.port="],
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
```
corvidian, line 15: What does this do? We have dropped this from other integration tests.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilter.kt
corvidian: We need to allow `/status`, `/actuator/health/liveness` and `/actuator/health/readiness` so the docker pods are considered active and we generate extra errors in monitoring.

I don't know of any reasons to allow anything else.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilter.kt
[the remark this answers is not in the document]
corvidian: Ok, didn't notice that. Maybe there should be a comment here?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilterITests.kt
```diff
@@ -0,0 +1,137 @@
...
+    controllers =
+        [
+            HakemusController::class,
+            HankeController::class,
+            HankeKayttajaController::class,
+            PublicHankeController::class,
+            StatusController::class,
+            ProfiiliController::class,
+        ],
+    properties = ["haitaton.api.disabled=true", "management.server.port="],
```
corvidian, line 35: The management port doesn't affect this test.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/ControllerTest.kt
umeetiusbaar: Yes, this is better than my choice. Will change.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilter.kt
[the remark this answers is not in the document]
umeetiusbaar: I tried the `logger` in super-class but it is not compatible with our logging system. I added a comment about why using `log` instead of `logger`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilter.kt
umeetiusbaar: Right you are. I changed the white-listing to allow only these endpoints.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilterActuatorITests.kt
umeetiusbaar: This will start the servlet engine in a random free port. Without using it I could not get these test with Actuator endpoints and RestTemplate to work. Also, it makes using `@LocalServerPort` to work and via it we have the actual port where the calls should be pointed at.

Setting `management.server.port=` is also necessary and without it the call gives an 404 error for Actuator endpoints and after that 401 error for trying to access `/error` without authorization.

Also, now I can test with this same live-like test the `/status` endpoint as well.

corvidian (approved):

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/ApiBlockingFilterActuatorITests.kt
umeetiusbaar: I changed it. Thank you for your help. I tried a lot of things but missed this one.

PR #694, opened 2024-04-19
HAI-2456 Send a notification email when a user is added on the application as a contact person

umeetiusbaar:
# Description
Send a notification email when a user is added on the application as a contact person. Since the email subject and text changed a bit they will need a translation (information about this was added in translation issue: https://helsinkisolutionoffice.atlassian.net/browse/HAI-2446).

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Create a new Johtoselvityshakemus, fill it up until Yhteystiedot, add a new contact person. It should receive an email notification.

# Checklist:

- [x] I (and PO) have made necessary changes to the documentation, link to confluence
 or other location: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8189902867/S+hk+posti-ilmoitukset+k+ytt+j+lle

# Other relevant info

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttaja.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttaja.kt
@@ -146,19 +146,6 @@ class HankekayttajaEntity(
         yhteyshenkilot.map { it.hankeYhteystieto.contactType }
 
     fun fullName() = listOf(etunimi, sukunimi).filter { it.isNotBlank() }.joinToString(" ")
-
-    override fun equals(other: Any?): Boolean {
-        if (this === other) return true
-        if (javaClass != other?.javaClass) return false
-
-        other as HankekayttajaEntity
-
-        return sahkoposti == other.sahkoposti
-    }
-
-    override fun hashCode(): Int {
-        return sahkoposti.hashCode()
-    }
 }
 
 data class HankeKayttaja(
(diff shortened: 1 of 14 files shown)


umeetiusbaar:
I noticed a bug in the email: it should show the name of the Hanke but that is blank.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttaja.kt
```diff
@@ -146,6 +146,19 @@ class HankekayttajaEntity(
...
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as HankekayttajaEntity
+
+        return sahkoposti == other.sahkoposti
+    }
+
+    override fun hashCode(): Int {
+        return sahkoposti.hashCode()
```
corvidian, line 160: This will fail if at some point there's a set of hankekayttaja from different hanke. The default hashCode (instance-based) is safer than this.

Or you could map the entities to domain objects before creating a set.

umeetiusbaar: I changed to use default hashCode/equals.

corvidian: This still makes me a bit uneasy. I would avoid using sets here. But if you're convinced this works correctly, you can keep this like it is.

corvidian (approved):

PR #707, opened 2024-04-30
HAI-1542 Add data fields for excavation notification area

umeetiusbaar:
# Description

Data for excavation notification area.

Also, a new data structure `Haitat` for holding the hindrance data. If suitable, perhaps can be used with Hanke also.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationArea.kt
```diff
@@ -10,5 +12,19 @@ sealed interface ApplicationArea {
...
-    ApplicationArea
+data class ExcavationNotificationArea(
+    override val name: String,
+    override val geometry: Polygon,
+    val hankealueId: Int,
+    val tyoalueet: List<Tyoalue>,
+    val katuosoite: String,
+    val tyonTarkoitukset: Set<TyomaaTyyppi>,
+    val lisatiedot: String,
+    val haitat: Haitat,
```
corvidian, line 23: On the ticket this was specified as
```kotlin
    meluhaitta: Meluhaitta
    polyhaitta: Polyhaitta
    tarinahaitta: Tarinahaitta
    kaistahaitta: VaikutusAutoliikenteenKaistamaariin
    kaistahaittojenPituus: AutoliikenteenKaistavaikutustenPituus
```
Not this is trying to implement the Haittojenhallintasuunnitelma?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationArea.kt
```diff
@@ -10,5 +12,19 @@ sealed interface ApplicationArea {
...
+) : ApplicationArea
+
+data class Tyoalue(
+    val geometry: Polygon,
+    val area: Double,
+    val haittaindeksi: Float,
```
corvidian, line 29: I think this should be a `TormaystarkasteluTulos`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationArea.kt
umeetiusbaar: I got carried away again when I tried to follow the ux design. Reverted to contain only those mentioned fields.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationArea.kt
umeetiusbaar: Yes, you are right. There should be more data than just the index. Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/application/ApplicationArea.kt
```diff
@@ -1,14 +1,43 @@
...
+) : ApplicationArea {
+    override fun geometries(): List<Polygon> = tyoalueet.map { it.geometry }
+}
 
-data class ExcavationNotificationArea(override val name: String, override val geometry: Polygon) :
-    ApplicationArea
+data class Tyoalue(
+    val geometry: Polygon,
+    val area: Double,
+    val tormaystarkasteluTulos: TormaystarkasteluTulos,
```
corvidian, line 42: This probably shouldn't be required for requests. The UI shouldn't need to enter dummy values when creating or updating the area. It also shouldn't be able to overwrite an existing result with an arbitrary one.

The 'correct' solution would be to have a separate class for the requests, but that might create a lot of complexity here?

corvidian: There's a separate ticket for calculating the tormaystarkastelutulos for the areas in the excavation notification (https://helsinkisolutionoffice.atlassian.net/browse/HAI-2315), so someof this could solved there as well.

umeetiusbaar: You're right. Thanks for noticing this aspect. I did not think so much about the fact that these data structures are now also used in requests and responses. I did try to create separate classes for requests now but it became too large and complex change for this so I only changed `tormaystarkasteluTulos` and `lisatiedot` as nullable for now.

corvidian (approved):

PR #712, opened 2024-05-06
HAI-2345 Only allow PDFs for attachments other than MUU

corvidian:
# Description

Only accept PDF attachments for `LIIKENNEJARJESTELY` and `VALTAKIRJA` type attachments. Traffic arrangements need to be PDFs since they are attached to the decision in Allu. The valtakirja needs to be a PDF since it needs to be signed.

Change the consumes parameters for the attachment controllers. This enables attachment uploading in SwaggerUI.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
http://localhost:3001/api/swagger-ui/index.html#/application-attachment-controller/postAttachment_1

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentValidatorTest.kt
+++ b/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentValidatorTest.kt
@@ -6,13 +6,19 @@ import assertk.assertThat
 import assertk.assertions.hasClass
 import assertk.assertions.hasMessage
 import assertk.assertions.isEqualTo
+import assertk.assertions.messageContains
+import fi.hel.haitaton.hanke.attachment.common.ApplicationAttachmentType
 import fi.hel.haitaton.hanke.attachment.common.AttachmentInvalidException
 import fi.hel.haitaton.hanke.attachment.common.AttachmentValidator
 import fi.hel.haitaton.hanke.attachment.common.validNameAndType
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
 import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
 import org.junit.jupiter.params.provider.CsvSource
+import org.junit.jupiter.params.provider.EnumSource
+import org.junit.jupiter.params.provider.MethodSource
 import org.junit.jupiter.params.provider.ValueSource
 import org.springframework.http.MediaType
 import org.springframework.mock.web.MockMultipartFile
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentValidatorTest.kt
```diff
@@ -116,6 +122,49 @@ class AttachmentValidatorTest {
...
+                    messageContains("filename=file.$extension")
+                    messageContains("attachmentType=$attachmentType")
+                }
+        }
+
+        private fun failingExtensions(): List<Arguments> {
+            val extensions = listOf("jpg", "jpeg", "png", "dgn", "dwg", "docx", "txt", "gt")
+
+            return ApplicationAttachmentType.entries
+                .filterNot { it == ApplicationAttachmentType.MUU }
```
umeetiusbaar, line 154: In the description it only states the two types that must be pdfs (`LIIKENNEJARJESTELY` and `VALTAKIRJA`). Here it tests that the type is not (currently) the last one (`MUU`). Should the filter here check for those two types instead of the one other? In case there is a 4th type at some point and it can only be pdf?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentValidatorTest.kt
```diff
@@ -116,6 +122,49 @@ class AttachmentValidatorTest {
...
+                .flatMap { type -> extensions.map { Arguments.of(type, it) } }
+        }
+
+        @ParameterizedTest
+        @ValueSource(strings = ["jpg", "jpeg", "png", "dgn", "dwg", "docx", "txt", "gt"])
+        fun `accepts other extensions for MUU type attachments`(extension: String) {
+            AttachmentValidator.validateExtensionForType(
+                "file.$extension",
+                ApplicationAttachmentType.MUU
```
umeetiusbaar, line 163: Also, here would it be better to test "other than LIIKENNEJARJESTELY or VALTAKIRJA" instead of "MUU"?

umeetiusbaar (commented):
Tested with Swagger and it works as described. Few comments for tests.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentValidatorTest.kt
corvidian: I'm betting on the potential new type having the same restrictions as `LIIKENNEJARJESTELY` and `VALTAKIRJA`. As the test is now, it tests that the new type can't have the other extensions.

umeetiusbaar: Thanks for the clarification. Yes, if adding a new type should allow other file types, this test will fail and needs to be fixed accordingly. And if it only allows pdfs, then this test will pass as it should.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/AttachmentValidatorTest.kt
umeetiusbaar: This test will never fail if new attachment types are added because it only tests the MUU type. New tests will need to be added for new types in case they are similar to MUU and allow different file types. And if they are similar to VALTAKIRJA and LIIKENNEJARJESTELY, then no new tests are needed. So, you're right.

umeetiusbaar (approved):

PR #715, opened 2024-05-07
HAI-2553 Allow deleting the application even when deleting the content of the application attachments fails

umeetiusbaar:
# Description

When deleting application, all its attachments are also deleted. If for some reason Azure Blob Storage cannot be reached the deletion of attachment content fails but the application is already cancelled in Allu. This leads to an inconsistency between Haitaton and Allu. 

This PR proposes a fix that allows running the deletion process through even if the content cannot be deleted due to an error in accessing Azure Blob Storage. This leaves the attachment content undeleted in blob storage but the attachment metadata in database is deleted.

This PR does not try to fix the other problem concering the allowance of deleting applications in CANCELLED state. There will be other PR for that fix.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/common/MockFileClient.kt
+++ b/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/common/MockFileClient.kt
@@ -16,6 +16,9 @@ class MockFileClient : FileClient {
 
     private val subfolderRegex = Regex("/.*/")
 
+    /** Mock client can simulate being connected or disconnected. */
+    var connected = true
+
     fun recreateContainers() {
         Container.entries.forEach { fileMap[it] = mutableMapOf() }
     }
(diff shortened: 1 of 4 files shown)


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/attachment/common/MockFileClient.kt
```diff
@@ -16,6 +16,9 @@ class MockFileClient : FileClient {
 
     private val subfolderRegex = Regex("/.*/")
 
+    /** Mock client can simulate being connected or disconnected. */
+    var connected = true
```
corvidian, line 20: Nice way to simulate connection troubles.

We might want to add `client.connected = true` to `MockFileClientExtension.beforeEach()` so failing tests can't leave the client in the disconnected setting.

umeetiusbaar: Good call, thanks! I did that now.

corvidian (approved):

PR #718, opened 2024-05-08
HAI-2552 Exclude inviter from hakemus emails

corvidian:
# Description

Don't send application notification emails when the recipient email would be the same as the inviter email. Effectively, this means that the user adding themselves as a new contact to an application won't receive an email about it.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Add yourself as a contact on a new application. Check [smtp4dev](http://localhost:3003/) that there are no notifications sent.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
@@ -51,7 +51,7 @@ import fi.hel.haitaton.hanke.permissions.HankeKayttajaService
 import fi.hel.haitaton.hanke.toJsonString
 import java.time.LocalDateTime
 import java.time.OffsetDateTime
-import java.util.*
+import java.util.UUID
 import kotlin.reflect.KClass
 import mu.KotlinLogging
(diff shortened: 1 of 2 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -847,25 +847,27 @@ class HakemusService(
...
+            applicationEntity.allContactUsers().filterNot { excludedUserIds.contains(it.id) }
+        if (newContacts.isEmpty()) {
+            return
+        }
+
+        val inviter =
+            hankeKayttajaService.getKayttajaByUserId(applicationEntity.hanke.id, userId)
+                ?: throw CurrentUserWithoutKayttajaException(userId)
+
+        for (newContact in newContacts.filter { it.sahkoposti != inviter.sahkoposti }) {
```
umeetiusbaar, line 860: Why not `newContacts.filter { ... }.forEach { newContact -> ...`?  But this works just as well, so this comment can be ignored.

umeetiusbaar (approved):

PR #723, opened 2024-05-14
HAI-2310 Clean up old contact model

corvidian:
# Description

Remove code that's no longer used after the kortisto model has been adapted in all environments.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


corvidian:
Split to smaller PRs

PR #740, opened 2024-05-23
HAI-1378 Calculate haittaindeksi separately for each hankealue

corvidian:
# Description

The calculated tormaystarkastelu results are saved and returned with each hankealue. One tormaystarkastelu is also returned for the whole hanke, for backwards compatibility. That's calculated by taking the worst cases from all of the areas.

Wipe all existing tormaystarkastelu, because they're not used in production and they can't be converted.

## Type of change

- [X] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Add several hankealue to a hanke.
2. Check the response from the browser's network tab to see that there are tormaystarkastelu under the hankealueet.
3. Check that the tormaystarkastelu are saved in DB.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeControllerITests.kt
@@ -287,13 +287,15 @@ class HankeControllerITests(@Autowired override val mockMvc: MockMvc) : Controll
                 SavedHankealue(
                     hankeId = hankeIds[0],
                     geometriat = Geometriat(1, FeatureCollection(), version = 1),
-                    nimi = "$HANKEALUE_DEFAULT_NAME 1"
+                    nimi = "$HANKEALUE_DEFAULT_NAME 1",
+                    tormaystarkasteluTulos = null,
                 )
             val alue2 =
                 SavedHankealue(
                     hankeId = hankeIds[1],
                     geometriat = Geometriat(2, FeatureCollection(), version = 1),
-                    nimi = "$HANKEALUE_DEFAULT_NAME 2"
+                    nimi = "$HANKEALUE_DEFAULT_NAME 2",
+                    tormaystarkasteluTulos = null,
                 )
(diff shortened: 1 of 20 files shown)


corvidian:
> One minor thing: decimal values do not seem to be stored accurately in database. For example, in my project one area got `autoliikenneindeksi=1.3` but in the database its value was `1.2999999523162842`. Maybe extension function `Float.roundToOneDecimal()` does not work as it should? This is something that has probably been as it is since the beginning and not related to this PR.

I think that's just floating point numbers and binary fractions at work. 1.3 has in infinite binary expansion in the same way as 1/3 has in decimal, so it can't be represenrted accurately.

We could use the decimal type in the DB to store the value accurately with a fixed point. Then we might have to use BigDecimal in Kotlin to represent the accurate value in code.

Or we could store the value as an integer of, say, 1/100s and convert the values when they are shown to users.

umeetiusbaar:
> > One minor thing: decimal values do not seem to be stored accurately in database. For example, in my project one area got `autoliikenneindeksi=1.3` but in the database its value was `1.2999999523162842`. Maybe extension function `Float.roundToOneDecimal()` does not work as it should? This is something that has probably been as it is since the beginning and not related to this PR.
> 
> I think that's just floating point numbers and binary fractions at work. 1.3 has in infinite binary expansion in the same way as 1/3 has in decimal, so it can't be represenrted accurately.
> 
> We could use the decimal type in the DB to store the value accurately with a fixed point. Then we might have to use BigDecimal in Kotlin to represent the accurate value in code.
> 
> Or we could store the value as an integer of, say, 1/100s and convert the values when they are shown to users.

How about changing the column type to `numeric`? 
```
alter table tormaystarkastelutulos
    alter column autoliikenne type numeric(2,1);

```

corvidian:
> How about changing the column type to `numeric`?

Yes, `numeric` is equivalent to `decimal`.

umeetiusbaar:
> > How about changing the column type to `numeric`?
> 
> Yes, `numeric` is equivalent to `decimal`.

Yes, but according to my tests this change will not affect how the fields are in Kotlin - they can remain as they are now.

corvidian:
I changed the index values to be numeric. This PR seemed like a good time to do it since we're wiping the data anyway.

PR #746, opened 2024-05-31
HAI-2589 Reject attachments without content

corvidian:
# Description

Reject attachments that have no content. Allu will not accept such attachments, so adding them to applications causes an error when sending the application to Allu. Attachments without content are very dubious anyways, so we can block them for hanke attachments as well.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Try to upload a file without content.

A file without content can be created from the shell with `touch test.txt`.

# Checklist:

 or other location:


umeetiusbaar (approved):
Tested and it works. Code looks good to me.

PR #747, opened 2024-06-03
HAI-2385 Refactor haittojenhallintasuunnitelma

corvidian:
# Description

- Use a typealias `Haittojenhallintasuunnitelma` for `Map<Haittojenhallintatyyppi, String>`.
- In tests, default to Hankealue having a haittojenhallintasuunnitelma.
- Update the factory to allow overriding or emptying single values from a haittojenhallintasuunnitelma.

Also, as an unrelated change, add the `tormays_cycle_infra_polys` table data to the postgresql container build script.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar (approved):
Nicely refactored with type alias.

PR #748, opened 2024-06-04
HAI-2385 Require nuisance control plan for public hanke

corvidian:
# Description

The nuisance control plan is always required for the YLEINEN and MUUT categories. For the transport categories (AUTOLIIKENNE, LINJAAUTOLIIKENNE, RAITIOLIIKENNE and PYORAILYLIIKENNE) it's required if the matching nuisance index in tormaystarkastelu is higher than zero. Currently, the car traffic index is always higher than zero if the index has been calculated, but it's checked the same way regardless.

If the tormaystarkastelu is not yet calculated, the transport categories are not checked because we don't know which categories are required. It's always calculated when all necessary information has been given, so either the validation will fail somewhere else or the tormaystarkastelu will be calculated.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Check that the project is not public until the nuisance control plan has been filled out.

# Checklist:

 or other location:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3"
 services:
   db:
     build:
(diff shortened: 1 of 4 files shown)


umeetiusbaar (approved):
Tested and it works. Code looks good to me.

PR #752, opened 2024-06-05
HAI Refactor TormaysService and remove unused test methods

corvidian:
# Description

- Restyle the older methods in TormaystarkasteluTormaysService.kt to match the newer ones.
- For `getIntersectingBusRoutes`
  - Implement the DISTINCT ON (route_id, direction_id)
  - Select only the columns that are used in the row mapper.
  - Share the row mapper with the GeoJSON version
- Remove unused methods from `HankeKayttajaServiceITest`.
- In the build script, replace the deprecated `kotlinOptions` with the newer `compilerOptions`.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTormaysService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTormaysService.kt
@@ -4,6 +4,7 @@ import fi.hel.haitaton.hanke.toJsonString
 import java.util.Collections
 import org.geojson.GeoJsonObject
 import org.springframework.jdbc.core.JdbcOperations
+import org.springframework.jdbc.core.RowMapper
 import org.springframework.stereotype.Service
 
 @Service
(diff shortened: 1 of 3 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTormaysService.kt
```diff
@@ -89,64 +87,40 @@ class TormaystarkasteluTormaysService(private val jdbcOperations: JdbcOperations
...
+            """
+            SELECT DISTINCT ON (buses.route_id, buses.direction_id)
+                buses.route_id,
+                buses.direction_id,
+                buses.rush_hour,
+                buses.trunk
+            FROM tormays_buses_polys buses, hankegeometria
+            WHERE hankegeometria.hankegeometriatid in ($placeholders)
+            AND ST_Intersects(buses.geom, hankegeometria.geometria)
+            ORDER BY buses.route_id, buses.direction_id, buses.trunk DESC, buses.rush_hour DESC
```
umeetiusbaar, line 100: Do we need ordering?

umeetiusbaar: Well, it does not make so much difference in performance that it would matter - I was mostly curious why use `ORDER BY` here when the order does not matter as the results are returned in an unordered `Set`.

corvidian: `DISTINCT ON (buses.route_id, buses.direction_id)` needs the rows to be ordered by `route_id` and `direction_id`, and it selects the first row with unique values of the parameters. So we sort them by `trunk` and `rush_hour`, selecting the worst case for each route and direction.

umeetiusbaar: I see. If I run these two sql clauses in our dev environment they both return the exactly same results except for one row:
```
SELECT DISTINCT ON (buses.route_id, buses.direction_id)
    buses.route_id,
    buses.direction_id,
    buses.rush_hour,
    buses.trunk
FROM tormays_buses_polys buses, hankegeometria
WHERE hankegeometria.hankegeometriatid in (select distinct hankegeometriatid from hankegeometria)
  AND ST_Intersects(buses.geom, hankegeometria.geometria) 
ORDER BY buses.route_id, buses.direction_id, buses.trunk DESC, buses.rush_hour DESC;
```

```
SELECT DISTINCT ON (buses.route_id, buses.direction_id)
    buses.route_id,
    buses.direction_id,
    buses.rush_hour,
    buses.trunk
FROM tormays_buses_polys buses, hankegeometria
WHERE hankegeometria.hankegeometriatid in (select distinct hankegeometriatid from hankegeometria)
  AND ST_Intersects(buses.geom, hankegeometria.geometria);
```

The difference is this:
```
19c19
<  1024     |            0 |         3 | no
---
>  1024     |            0 |         0 | no
```

Without `ORDER BY` it does not necessarily get the max `rush_hour`. 

Goog to know this, thanks!

umeetiusbaar (approved):

PR #753, opened 2024-06-05
HAI-2601 Accept filename that ends in a dot

corvidian:
# Description

Change the file name sanitazion to split the filename to base and extension and sanitize those separately. Then rejoin them and add a dot if extension is not empty.

This way, dot can be added as an invalid character and all other dots except the extension separator will be replaced with the underscore `_`. File names with path traversal will have their `../` segments changed to `___`.

Handle null filenames separately to make this process easier.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
1. Create a file with a name like `testi luon..txt`.
2. Add it to a hanke or hakemus.
3. After saving its filename should be shown as `testi luon_.txt`.

# Checklist:

 or other location:


umeetiusbaar (approved):

PR #757, opened 2024-06-12
HAI-2401 Add car traffic nuisance classifications

umeetiusbaar:
# Description

* Added car traffic nuisance classifications/sub-categories (API and database)
* Use Kaivokatu test geometry for testing instead of previous geometry that was outside of all the traffic geometries
* Made the change originally in https://helsinkisolutionoffice.atlassian.net/browse/HAI-2600 ("autoliikenteen haittaindeksin tulee olla nolla, jos alueelta ei löydy katuluokkaa eikä liikennemäärää")
* DB migration deletes all `tormaystarkastelutulos` rows and changes the status for all projects as `DRAFT`

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
* Use the Haitaton UI to test that the API returns `tormaystarkasteluTulos.autoliikenne` as a proper data structure with all the sub-categories. Same goes to the new API `/haittaindeksit` which should also include the proper values for `autoliikenne`.
* DB migration will clear all `tormaystarkastelutulos` rows - this means that the areas must be changed either by changing the geometry or changing some value: the dates or lane nuisance or lane length nuisance

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
```diff
@@ -30,7 +30,15 @@ class HankealueService(
...
-            .mapNotNull { it.geometriat }
+            .mapNotNull {
+                // If car traffic nuisance indexes are not yet calculated, calculate them now.
+                it.tormaystarkasteluTulos?.let { tt ->
+                    if (tt.haitanKesto == -1) {
+                        updateTormaystarkastelu(it)
+                    }
+                }
+                it.geometriat
+            }
```
corvidian, line 41: I can see what this is trying to do, but I think this is a lousy way of handling the data migration.

* We can't know if the migration will ever be completed. Someone needs to access the hanke to update its indices. For some draft hanke, this might never happen.
* This needs to use a hacky invalid value for `haitanKesto`.
* We're changing data in a mapper that's supposed to be a read-only method.
  - This is potentially called outside any transactions and the like.
  - This will make HTTP GET calls to cause data changes.

Since we have no indices calculated in production, we could delete all indices from the DB. They would be recalculated when someone edits the hanke. We might have to mark all hanke as drafts as well.

It should also be possible to recalculate the indices completely in the SQL migration, but that might be a painful SQL script to write and verify.

services/hanke-service/src/main/resources/db/changelog/changesets/077-add-car-traffic-classification.yml
```diff
@@ -0,0 +1,38 @@
...
+      changes:
+        - addColumn:
+            tableName: tormaystarkastelutulos
+            columns:
+              - column:
+                  name: haitan_kesto
+                  type: int
+                  constraints:
+                    nullable: false
+                  defaultValue: -1
```
corvidian, line 14: The default values should be removed after the columns have been added.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
```diff
@@ -46,13 +48,37 @@ data class TormaystarkasteluTulos(
...
+    @JsonView(ChangeLogView::class) val katuluokka: Int,
+    @JsonView(ChangeLogView::class) val liikennemaara: Int,
+    @JsonView(ChangeLogView::class) val kaistahaitta: Int,
+    @JsonView(ChangeLogView::class) val kaistapituushaitta: Int,
+) {
+    @get:JsonView(ChangeLogView::class)
+    val indeksi: Float by lazy {
+        if (katuluokka == 0 || liikennemaara == 0) {
+            0.0f
+        } else {
```
corvidian, line 71: Isn't this jumping the gun a bit? There's a separate task for this. Why implement it here?

Besides, the index should be zero when both are zeroes.

There are no tests for this logic.

umeetiusbaar: The logic is wrong, you're right about that. I checked it only from HAI-2600 where it was either one. I fixed the phrasing there. I thought to implement it here since it only needed this single if-clause. But, if we use the stored value in database and not this lazyly calculated value, then this calculation functionality would be placed elsewhere. The old implementation of using a map of weights was a bit clumsy imo. I will think of something else...

services/hanke-service/src/main/resources/db/changelog/changesets/077-add-car-traffic-classification.yml
umeetiusbaar: Should there be any default values at all? Since we are not using my clumsy and out-of-place migration logic.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
umeetiusbaar: Yes, you're right. I will rewrite the database migration to delete all indices. I can also look into the SQL way of calculating the indices in the migration script but as you said, it would be a not-too-easy task.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
```diff
@@ -46,13 +48,37 @@ data class TormaystarkasteluTulos(
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class Autoliikenneluokittelu(
+    @JsonView(ChangeLogView::class) val haitanKesto: Int,
+    @JsonView(ChangeLogView::class) val katuluokka: Int,
+    @JsonView(ChangeLogView::class) val liikennemaara: Int,
+    @JsonView(ChangeLogView::class) val kaistahaitta: Int,
+    @JsonView(ChangeLogView::class) val kaistapituushaitta: Int,
+) {
+    @get:JsonView(ChangeLogView::class)
+    val indeksi: Float by lazy {
```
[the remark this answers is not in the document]
umeetiusbaar: You're probably right. It should be either not having the calculated index in database or use the calculated value from database. Will think of something different for this...

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
corvidian: Sorry for misphrasing the ticket description.

Yes, I agree the old implementation with the LuokitteluType map was too complex. This simple calculation is a lot nicer.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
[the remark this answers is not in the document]
corvidian: I think it would be better to save the index to DB, so the indices of old hanke don't change suddenly if we adjust the weights at some point.

services/hanke-service/src/main/resources/db/changelog/changesets/077-add-car-traffic-classification.yml
corvidian: If you delete all rows first, the defaults are not needed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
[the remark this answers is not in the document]
umeetiusbaar: I changed this to use the value in database and added a function for calculating the index. This function is used in 2nd constructor that is used when new instance is created in calculation.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/HankealueService.kt
umeetiusbaar: I did not go for the SQL migration scripting but just simply delete all `tormaystarkastelutulos` rows and set project status to `DRAFT`.

services/hanke-service/src/main/resources/db/changelog/changesets/077-add-car-traffic-classification.yml
umeetiusbaar: No more default.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
umeetiusbaar: I left this if-clause in the new function which calculates the index, corrected its logic and added tests that checks this logic.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/tormaystarkastelu/TormaystarkasteluTulos.kt
```diff
@@ -46,13 +48,59 @@ data class TormaystarkasteluTulos(
...
+            katuluokka: Int,
+            liikennemaara: Int,
+            kaistahaitta: Int,
+            kaistapituushaitta: Int,
+        ): Float =
+            if (katuluokka == 0 && liikennemaara == 0) {
+                0.0f
+            } else {
+                (0.1f * haitanKesto +
+                        0.2f * katuluokka +
```
corvidian, line 80: I'm not really loving having important business logic inside a data class like this. Data classes should be simple and boring.

Maybe the calculation function could be moved to the companion object of `TormaystarkasteluLaskentaService`?

umeetiusbaar: Valid point. I moved the function and companion object into `TormaystarkasteluLaskentaService`. And the test. Also, added a separate test for "no street class nor traffic" logic.

PR #758, opened 2024-06-13
HAI-2416 Validate kaivuilmoitus for sending to Allu

corvidian:
# Description

Validate it according to the confluence page:
https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/8626798660/Kaivuilmoituksen+pakolliset+tiedot

Combine missing and error validation, since they were both only done when sending. Update requests have separate validation.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Try to send the application with different bits of information missing.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/Validators.kt
```diff
@@ -98,6 +98,25 @@ sealed class ValidationResult {
         return this
     }
 
+    /**
+     * If either this or the validated function is ok, return ok.
+     *
+     * If both have errors, return all errors.
```
umeetiusbaar, line 104: "return ok" and "return all errors" sound a bit misleading since the returned object is always `this` ValidationResult. Maybe "return without errors" and "return with all errors"?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/KaivuilmoitusSendValidationTest.kt
```diff
@@ -0,0 +1,610 @@
...
+        val customer =
+            HakemusyhteystietoFactory.create(tyyppi = tyyppi, ytunnus = null).withYhteyshenkilo()
+        val hakemus = hakemus.copy(customerWithContacts = customer)
+
+        assertThat(hakemus.validateForSend()).succeeds()
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = CustomerType::class, names = ["COMPANY", "ASSOCIATION"])
+    fun `succeeds when the company contractor has no registry key`(tyyppi: CustomerType) {
```
umeetiusbaar, line 199: This sounds a bit odd but it is according to the specification so it should be okay.

umeetiusbaar (approved):
Very thorough testing of the validation logic👍

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/validation/Validators.kt
corvidian: The returned object can also be the result of the given validation function, when `this` has errors, but the new result doesn't.

I made ValidationResult immutable. Now the comments should be correct?

PR #782, opened 2024-07-29
HAI-2609 Mark replaced decisions as `KORVATTU`

corvidian:
# Description

Add handling of `REPLACED` status updates. Don't update the application status or identifier when we receive one. Instead, update all decisions in the `paatos` table with the same hakemustunnus as the update and set their status to `KORVATTU`.

Whenever Allu marks a decision as `REPLACED`, there should be a matching `DECISION` update with a new identifier.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create an application and create a decision for it in Allu. You can also add an operational condition decision if you want.
2. Create a korvaava päätös in Allu and accept it.
3. Check that the `paatos` table has rows for both the replaced and the new decision and that their statuses are `KORVATTU` and `NYKYINEN`, respectively.
4. Check that Azure Storage Explorer shows files for both decisions.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/paatos/PaatosService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/paatos/PaatosService.kt
@@ -9,6 +9,7 @@ import fi.hel.haitaton.hanke.hakemus.HakemusIdentifier
 import mu.KotlinLogging
 import org.springframework.http.MediaType
 import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
 
 private val logger = KotlinLogging.logger {}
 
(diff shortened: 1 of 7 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/paatos/PaatosService.kt
```diff
@@ -21,6 +22,11 @@ class PaatosService(
     fun findByHakemusId(hakemusId: Long): List<Paatos> =
         paatosRepository.findByHakemusId(hakemusId).map { it.toDomain() }
 
+    @Transactional
```
umeetiusbaar, line 25: Why `@Transactional`? Why not also in `saveKaivuilmoituksenPaatos`? Or maybe without it in both? Calling this function originates (besides the unit tests) from HakemusService.handleHakemusUpdate which is tagged as `@Transactional` - isn't that enough?

corvidian: I made this differently at first. I loaded the päätökset with the given hakemustunnus from the DB, then had a for-loop updating the status of each päätös. This needed `@Transactional` to persist the changes. The integration tests didn't work without it. The annotation is a remnant of that.

Now there's an explicit update statement - and only one statement - so the transaction is no longer required.

I think public methods should be marked as `@Transactional` if they need to be run in a transaction. That way there are no surprises if the caller is changed or the method is called from someplace else. There's very little downside. The default propagation is `REQUIRED`, which means any existing transaction will be used.

It's not completely clear to me if we should use `@Transactional` in all methods that call the database. Some people think it's never harmful, so it should be used for safety. Marking all database methods as either `@Transactional` or `@Transactional(readOnly=true)`

I'll use this approach here, so I'll mark all the public methods as `@Transactional`.

PR #797, opened 2024-08-14
HAI-2556 Add API for reporting operational condition

corvidian:
# Description

Check for 3 things:
1. The application is an excavation notification.
2. The reported date is between the start of the application work time and today.
3. The application is in a status where reporting operational condition is allowed.

If everything is OK, send the date to Allu.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Use [Swagger-UI](http://localhost:3001/api/swagger-ui/index.html#/hakemus-controller/operationalCondition) to report a operational condition.

When successful, it should create a supervision task in Allu (Toiminnallisen kunnon valvonta under Valvonta). The operational condition date should also be on the basic info tab under Asiakkaan ilmoittama voimassaolo -> Talvityön toiminn. kunto.

# Checklist:

 or other location: API docs

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
@@ -43,8 +43,10 @@ import io.mockk.checkUnnecessaryStub
 import io.mockk.clearAllMocks
 import io.mockk.confirmVerified
 import io.mockk.every
+import io.mockk.justRun
 import io.mockk.verify
 import io.mockk.verifySequence
+import java.time.LocalDate
 import java.time.ZonedDateTime
 import java.util.UUID
 import org.junit.jupiter.api.AfterEach
@@ -66,6 +68,7 @@ import org.springframework.security.test.web.servlet.response.SecurityMockMvcRes
 import org.springframework.test.context.ActiveProfiles
 import org.springframework.test.web.servlet.MockMvc
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.content
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
 
@@ -563,8 +566,8 @@ class HakemusControllerITest(@Autowired override val mockMvc: MockMvc) : Control
             every { hakemusService.updateHakemus(id, request, USERNAME) } throws
                 IncompatibleHakemusUpdateRequestException(
                     HakemusFactory.create(id = id),
-                    JohtoselvityshakemusEntityData::class,
-                    KaivuilmoitusUpdateRequest::class,
+                    KaivuilmoitusEntityData::class,
+                    JohtoselvityshakemusUpdateRequest::class,
                 )
 
             put(url, request)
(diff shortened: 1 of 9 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
```diff
@@ -687,6 +690,128 @@ class HakemusControllerITest(@Autowired override val mockMvc: MockMvc) : Control
...
+            post(url, request)
+                .andExpect(status().isNotFound)
+                .andExpect(hankeError(HankeError.HAI2001))
+
+            verifySequence {
+                authorizer.authorizeHakemusId(id, PermissionCode.EDIT_APPLICATIONS.name)
+            }
+        }
+
+        private fun RuntimeException.whenThrown(f: () -> Unit) {
```
umeetiusbaar, line 744: Would it be better if this extension function was last and not here in the middle of test methods?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
```diff
@@ -687,6 +690,128 @@ class HakemusControllerITest(@Autowired override val mockMvc: MockMvc) : Control
...
+                authorizer.authorizeHakemusId(id, PermissionCode.EDIT_APPLICATIONS.name)
+                hakemusService.operationalCondition(id, date)
+            }
+        }
+
+        @Test
+        fun `returns 409 when the application is not yet in Allu`() {
+            val exception = HakemusNotYetInAlluException(HakemusFactory.create(id = id))
+
+            exception.whenThrown {
```
umeetiusbaar, line 762: For me this "whenThrown" reads in my mind that "when this exception is thrown, do the following things". Instead, the inner function call throws this exception. It is good that with this the test methods are more clean but this syntax is a bit hard to read. If it was "exception.thrownWhen" or "isThrownWhen" then perhaps its purpose was more clear?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/AlluClient.kt
```diff
@@ -206,6 +209,18 @@ class AlluClient(
         }
     }
 
+    fun operationalCondition(alluApplicationId: Int, operationalConditionDate: LocalDate) {
```
umeetiusbaar, line 212: Other methods have INFO logging at the beginning - should this too?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/AlluClient.kt
```diff
@@ -206,6 +209,18 @@ class AlluClient(
...
+        put(
+                "excavationannouncements/$alluApplicationId/operationalcondition",
+                operationalConditionDate.atStartOfDay(TZ_UTC).toJsonString())
+            .bodyToMono(Void::class.java)
+            .timeout(defaultTimeout)
+            .doOnError(WebClientResponseException::class.java) {
+                logError("Error canceling application in Allu", it)
```
umeetiusbaar, line 219: Copy-paste mistake with "cancelling"?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusController.kt
```diff
@@ -246,6 +246,39 @@ class HakemusController(
         return result
     }
 
+    @PostMapping("/hakemukset/{id}/toiminnallinen-kunto")
+    @Operation(
+        summary = "Report the operational condition date for a excavation notification",
+        description =
+            """
+Report the operational condition date for a excavation notification.
```
umeetiusbaar, line 254: Should it be "for AN excavation notification"?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusController.kt
```diff
@@ -395,4 +428,36 @@ class HakemusController(
...
+    @ExceptionHandler(HakemusNotYetInAlluException::class)
+    @ResponseStatus(HttpStatus.CONFLICT)
+    @Hidden
+    fun hakemusNotYetInAlluException(ex: HakemusNotYetInAlluException): HankeError {
+        logger.warn(ex) { ex.message }
+        return HankeError.HAI2013
+    }
+
+    @ExceptionHandler(WrongHakemusTypeException::class)
+    @ResponseStatus(HttpStatus.CONFLICT)
```
umeetiusbaar, line 441: I wonder whether this error - the given id is for an application that has a type that is not allowed for this operation - is more a 400 error than 409. 

The HTTP 409 Conflict status code indicates a request conflict with the current state of the target resource.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
```diff
@@ -446,6 +451,157 @@ class HakemusServiceTest {
...
+                messageContains("date=2024-08-07")
+                messageContains("id=${hakemus.id}")
+            }
+            verifySequence { hakemusRepository.findOneById(id) }
+        }
+
+        @Test
+        fun `throws exception when date is in the future`() {
+            every { hakemusRepository.findOneById(id) } returns
+                hakemus.copy(
```
umeetiusbaar, line 526: Why adjusting the dates in this test? Isn't it okay to use the same data as in previous test:
`every { hakemusRepository.findOneById(id) } returns hakemus`

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
```diff
@@ -446,6 +451,157 @@ class HakemusServiceTest {
         }
     }
 
+    @Nested
+    inner class OperationalCondition {
```
umeetiusbaar, line 455: Is a test for the application type missing here?

umeetiusbaar:
I tested this with an application in state `PENDING` (KP2400007) and it worked okay.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
corvidian: I was going for "When the mock service throws an exception".

`withMockedException(exception)`?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
corvidian: To check that the important part is being in the future, not being over the hakemus end date.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
corvidian: It's tested in the integration tests.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
corvidian: Yes, thank you. I moved it to behind the test methods.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
corvidian: I renamed this to `withMockedException(exception)`.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/AlluClient.kt
corvidian: Yes, I added the logging.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/allu/AlluClient.kt
corvidian: Yes, thank you. Wrote a proper error message.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusController.kt
corvidian: I only considered the date as a part of the request, so I considered this to be 409 Conflict. But the id is also a part of the request, so the user has a way to fix their request, which is an indication of 400 Bad Request.

Using 400 now.

corvidian:
I renamed the `operationalCondition` methods to `reportOperationalCondition` to make it clear that the call changes something.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
umeetiusbaar: Yes, it is, but why there and not here? I understand that in integration test one tests the ingration between services, the database and external services but checking the type of the application does not need any of those things, right? At least for me it seems a similar test than testing the dates or states which are tested here in the unit test.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
umeetiusbaar: But it can be over the end date. E.g. if the application dates are 8.8.2024 - 12.8.2024, it is perfectly okay to give a date of e.g. 15.8.2024 in the call, right?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
corvidian: It's in integration tests because currently, there's no easy way to create a hakemus with an arbitrary type without the builder, which needs database access.

I had all these tests in the integration tests at first, but with so many parametrized tests they were slow so I moved them to unit tests.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceTest.kt
corvidian: Yes, that case is tested in the parametrized date test.

umeetiusbaar: There the tests do not test a future date. I should have written above 16.8.2024 instead of 15.8.2024. What I mean is that there is no need to change the dates of the application in this test. Or maybe there can be a separate tests cases for application dates that are from past to future and dates that are all in the pasts?

corvidian: Technically, there's no need to change the dates, but I think this way it's clearer for the reader that the exception is thrown because the date is in the future, not because we're past the end date.

Is there a downside to changing the hakemus dates for this test?

umeetiusbaar: I guess not. On my first reading the change of dates struck me as something that has to be done in order to make the test pass. Then I realized that the test would pass with the same data as in previous test as well. That's why these questions. It is okay as it is.

umeetiusbaar (approved):

PR #798, opened 2024-08-15
HAI-2398 Add audit logging to accompanying johtoselvityshakemus

corvidian:
# Description

When an accompanying johtoselvityshakemus is created, write the creation to the audit logs.

Create a new entity for the johtoselvityshakemus instead of copying one from the excavation notification. Set the founder to the current user instead of the excavation notification founder.

Also, introduce convenience methods for asserting that the object before or after is null. Adopt their use wherever it was done manually.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
- Send an accompanying johtoselvityshakemus.
- Check the audit_logs database table for the CREATE entry.

# Checklist:

 or other location:


umeetiusbaar (approved):
Sorry about missed audit logging and bug in creation of the johtoselvityshakemus.

PR #800, opened 2024-08-19
HAI-1504 Add an API for downloading kaivuilmoitus decisions

corvidian:
# Description

Add an API for downloading decision and approval documents for operational conditions and work finished. Downloads are allowed to anyone with view access to the hanke.

Write the decision downloads to audit logs. The decisions contain personal information, so accessing them needs to be logged. Log paatos metadata, since we can't know what personal information is on the decisions.

Also, remove `APPLICATION` from audit log object types. It was only used in tests, and incorrectly in those. The tests were fixed as well.

Enable a recently added feature in Springdoc that trims indentation from Kotlin's multiline strings. Long descriptions and descriptions with markdown formatting can be written with natural indentations.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
- Find out the decisions for an application with decisions with get single hakemus endpoint: http://localhost:3001/api/swagger-ui/index.html#/hakemus-controller/getById or look in `paatos` table in DB.
- Call the download API. Swagger UI doesn't work correctly with files, so this can be done with cURL:
  - ```curl 'http://localhost:3001/api/paatokset/<id>' -H 'Authorization: Bearer <token>' -H 'Connection: keep-alive' -H 'accept: application/pdf' -OJ -w "Downloaded: %{filename_effective}" -v"```
  - `-OJ` makes cURL use the filename from the Content-Disposition header.
  - `-w` tells cURL to output the name of the downloaded file
  - `-v` makes cURL show headers
  - cURL will not overwrite files by default. To download the same file again, delete the file or add `--clobber` to make cURL overwrite the existing file.
- Check in DB that there's an entry in audit logs.
  - `select * from audit_logs where message->'audit_event'->'target'->>'type' = 'PAATOS';`

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusBuilder.kt
+++ b/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusBuilder.kt
@@ -210,6 +210,17 @@ data class HakemusBuilder(
                     .tyonSuorittaja(founder())
         }
 
+    /**
+     * Make the hakemus appear like it has been just sent. I.e. it has mandatory fields and allu
+     * status fields filled.
+     */
+    fun asSent(
+        status: ApplicationStatus? = ApplicationStatus.PENDING,
+        alluId: Int? = 1,
+        identifier: String? = "JS000$alluId",
+        hankealue: SavedHankealue? = null
+    ) = this.withMandatoryFields(hankealue).withStatus(status, alluId, identifier)
+
     private fun founder(): HankekayttajaEntity =
         hankeKayttajaService.getKayttajaByUserId(hankeId, userId)!!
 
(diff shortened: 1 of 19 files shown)


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusBuilder.kt
```diff
@@ -210,6 +210,17 @@ data class HakemusBuilder(
...
         }
 
+    /**
+     * Make the hakemus appear like it has been just sent. I.e. it has mandatory fields and allu
+     * status fields filled.
+     */
+    fun asSent(
+        status: ApplicationStatus? = ApplicationStatus.PENDING,
+        alluId: Int? = 1,
+        identifier: String? = "JS000$alluId",
```
umeetiusbaar, line 220: Maybe it would be better to use a type-specific identifier? E.g. `JS2400001` or `KP2400001`.

corvidian: I made the identifier depend on the application type.

umeetiusbaar (approved):

PR #802, opened 2024-08-21
HAI-2651 Configure Haitaton for Keycloak instead of tunnistamo

corvidian:
# Description

Use the default JWT flow in Spring Security, since the default is Authorization Code with PKCE, which is what's used for the new Keycloak authorization.

Update the authentication to the Profiili API. With keycloak, getting the API token needs a different kind of request than before.

The address for the API token request can be fetched from the OpenID configuration endpoint. Getting it from there means we have one less environment-specific thing to configure.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
- Start UI from the https://github.com/City-of-Helsinki/haitaton-ui/pull/828 branch.
- Login.
- See that you get the verified name on the header bar and the everything is working as it should.

# Checklist:

 or other location:

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,9 @@ Haitaton supports uploading of attachment files. Files are validated with ClamAV
 
 ## Authentication
 
-Authentication is done by calling Helsinki Profiili's userinfo-url with the Bearer-token.
+Authentication is handled by Spring Boot's default configuration. It defaults to the Authentication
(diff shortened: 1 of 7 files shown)


umeetiusbaar (approved):
Tested with UI and Swagger and everything works okay.

PR #803, opened 2024-08-22
HAI-2657 Send attachment type to Allu in description

corvidian:
# Description

There's no field in Allu that would say which kind of attachment we're sending there. For the handlers looking at the application, it's useful knowledge so they don't have to open each one to see which one it is. So we include the type in the description field of the attachment.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
1. Add attachments to an application
2. Send it to Allu
3. Look at the application in Allu and check that the attachments have descriptions with the attachment type
    - The generated form data PDF has a different description which doesn't include the type

# Checklist:

 or other location:


umeetiusbaar (approved):

PR #806, opened 2024-08-26
HAI-2655 GDPR-API changes for new Tunnistus

corvidian:
# Description

Fix authentication when GDPR API is enabled. Tunnistamo used a different authentication flow for regular API calls and GDPR API calls. Only the GDPR API used JWT, so we had a JWT decoder bean that handled that GDPR API authentication. After adopting the new Tunnistus (Keycloak) authentication, it would break when GDPR API was enabled since all incoming tokens would be decoded with the GDPR decoder. Fix this by making the GDPR flow call the proper function instead of relying on a special registered bean.

With Keycloak, the GDPR API authentication has a lot less environment-specific variables compared to Tunnistamo. The scope structure is always as same as the scopes themselves. The issuer and audience of the token are the same as the general purpose API, so the environment variables can be shared. If we didn't need gdpr-api-tester for testing the calls locally, we wouldn't need a separate authentication at all, just the subject and scope authorization in GdprController.

## Type of change

- [X] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Set `HAITATON_GDPR_DISABLED` to false in `.env` and then test using the profile-gdpr-api-tester as instructed in README. Also, check that Haitaton works in general.

# Checklist:

 or other location:

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,10 +244,11 @@ Set the API Tester configuration as follows (in profile-gdpr-api-tester/.env):
 
 ```
   ISSUER = http://gdpr-api-tester:8888/
-  GDPR_API_AUDIENCE = http://localhost:8080/haitaton
+  ISSUER_TYPE = keycloak
+  GDPR_API_AUDIENCE = haitaton-api-dev
   GDPR_API_AUTHORIZATION_FIELD = http://localhost:8080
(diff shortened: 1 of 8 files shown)


umeetiusbaar (approved):
Tested with profile-gdpr-api-tester and UI. It works as described.

PR #807, opened 2024-08-26
HAI-2674 Send johtoselvitys complete emails after commit

corvidian:
# Description

A bug has occurred a couple of times in dev environment where Haitaton starts spamming emails to the contacts of a hanke. This has happened because the database changes have been rolled back because of a key conflict. The emails were sent during the transaction before we were sure they could be completed successfully.

Fix this by using `@TransactionalEvenListener`, which listens to events sent through `ApplicationEventPublisher`, but waits (by default) for the database transaction to commit successfully before handling the event. If the commit fails, the event is not handled. This ensures that the changes have been committed to the database before we send the email, so we don't send the same email several times.

It's not really possible to test the case where the database commit fails, except by waiting x number of seconds and see that no email was sent. I don't want to add unnecessary delays to tests, so we only check the positive case in integration tests. The unit tests test that the email is sent through the event publisher, so things should work as long as Spring handles it's job correctly.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Check the ticket for reproduction instructions. Without this fix, there should be a steady stream of emails in smtp4dev (http://localhost:3003/). When restarting with this fix, the emails should stop.

# Checklist:

 or other location:


umeetiusbaar (approved):
Great thing to take this listener approach into action. Tested and it works as described.

PR #810, opened 2024-08-29
HAI-2667 Add additional info to Allu data mapper

corvidian:
# Description

When mapping kaivuilmoitus from Haitaton data to Allu data, move the additional info field along the other data.

Remove the default values from AlluApplicationData and its implementations. Having null default values makes mistakes like this easier since fields can be ignored without errors. Now, each time the class is created, there needs to be an explicit value for each field, even if that value is null.

Also, reorder the fields in HakemusDataMapper to match the order declared in the data classes.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
- Create a kaivuilmoitus with something written in additional info ("Lisätietoja hakemuksesta").
- Send it to Allu
- Check that Allu shows the additional information (Kaivuilmoituksen tiedot -> Lisätietoja).

# Checklist:

 or other location:


umeetiusbaar (approved):
Tested and it works as described. Code changes without default values are good.

PR #811, opened 2024-08-29
HAI-2600 Use only table for street class calculation

corvidian:
# Description

Only use the `tormays_street_classes_polys` for calculating the street class of an area. Ylre data will be merged into that table as needed.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
It's hard to find a place where this makes a difference. One such place is in Kalasatama / Suvilahti.

In dev, this shows green for a class 1 road because the planned Kaasutehtaankatu is in the YLRE material:
image

With this change, it's not recognized, because Liikenneväylät only have existing roads:
image

# Checklist:

 or other location:


umeetiusbaar (approved):
Tested with the area in description and it works as described. Code looks good and is more simple now.

PR #813, opened 2024-09-04
HAI-2697 Send all emails after transactions

corvidian:
# Description

Use `TransactionalEventListener` in all emails to send them after the database transaction has been closed. Sending the emails seems to be taking a while in production, so keeping the transaction active while the email is being sent can lead to problems.

There were cases in production where two application updates were sent back-to-back - probably because the user skipped the attachments page instinctively - which caused errors because the second update didn't see the database changes the first one had made, so it tried to create a second version of the same user.

This isn't a perfect solution for that problem, but keeping the transaction short should mitigate the problem.

One of the side effects of the problem was that the notification email for being added to an application was sent twice. This should be completely solved since the second update failed inside the transaction, so the email event will not be processed.

Also, add some extra logging to sending the emails, so we know how long it takes.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
The affected emails are:
- Hanke invitation
- Application notification
- Access update
- Removal from hanke

So to test all of them:
1. Create a new application (either a cable report or an excavation notification).
2. Fill it out while creating a new contact person and adding it to the application.
3. Save the application, don't send it.
4. Go to the project's User management and give the new user higher access rights.
5. Remove the user.

After these steps, 4 emails should have been sent to the new user, one for each on the previous list.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/email/EmailSenderServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/email/EmailSenderServiceITest.kt
@@ -159,10 +159,10 @@ class EmailSenderServiceITest : IntegrationTest() {
     @Nested
     inner class HankeInvitation {
         private val notification =
-            HankeInvitationData(
+            HankeInvitationEmail(
                 inviterName = INVITER_NAME,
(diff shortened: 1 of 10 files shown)


umeetiusbaar (approved):

PR #814, opened 2024-09-05
HAI-2661 Hanke into RAKENTAMINEN phase when new application is created

umeetiusbaar:
# Description

When creating a new johtoselvitys without existing hanke, the phase of generated hanke is set to RAKENTAMINEN. Also, when creating new johtoselvityshakemus or kaivuilmoitus for an existing hanke, the phase of the hanke is set to RAKENTAMINEN if not already set.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
For generated hanke:
* Create a new johtoselvityshakemus (without existing hanke)
* Fill enough data for johtoselvityshakemus in order to be able to save it
* Go to the hanke page of the generated hanke and see that its phase is Rakentaminen
* Check audit log rows that should be created for hanke updates

For other hanke:
* Create a new hanke and fill enough data for it in order it to public (ie. allow applications to be created for it) - also, SET THE PHASE TO SOMETHING ELSE THAN RAKENTAMINEN
* Create a new application for the hanke (johtoselvitys or kaivuilmoitus)
* Fill enough data for the application in order to be able to save it
* Go to the hanke page and see that its phase is Rakentaminen
* Check audit log rows that should be created for hanke updates

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
@@ -17,8 +17,8 @@ import fi.hel.haitaton.hanke.allu.Attachment
 import fi.hel.haitaton.hanke.allu.AttachmentMetadata
 import fi.hel.haitaton.hanke.attachment.application.ApplicationAttachmentService
 import fi.hel.haitaton.hanke.attachment.common.ApplicationAttachmentType
-import fi.hel.haitaton.hanke.email.ApplicationNotificationEmail
 import fi.hel.haitaton.hanke.domain.Hankevaihe
+import fi.hel.haitaton.hanke.email.ApplicationNotificationEmail
 import fi.hel.haitaton.hanke.email.JohtoselvitysCompleteEmail
 import fi.hel.haitaton.hanke.email.KaivuilmoitusDecisionEmail
 import fi.hel.haitaton.hanke.geometria.GeometriatDao
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
@@ -127,10 +127,9 @@ class HakemusService(
             hanke.version = hanke.version?.inc() ?: 1
             hanke.modifiedByUserId = userId
             hanke.modifiedAt = getCurrentTimeUTCAsLocalTime()
-            hanke.generated = false
 
-            hankeRepository.save(hanke)
-            val hankeAfterUpdate = HankeMapper.domainFrom(hanke, geometriatMap)
+            val savedHanke = hankeRepository.save(hanke)
+            val hankeAfterUpdate = HankeMapper.domainFrom(savedHanke, geometriatMap)
             hankeLoggingService.logUpdate(hankeBeforeUpdate, hankeAfterUpdate, userId)
         }
     }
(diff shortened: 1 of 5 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -109,9 +111,28 @@ class HakemusService(
...
+    /** Update the hankevaihe of a Hanke to RAKENTAMINEN if it's not already there. */
+    private fun updateHankevaiheToRakentaminen(hanke: HankeEntity, userId: String) {
+        if (hanke.vaihe != Hankevaihe.RAKENTAMINEN) {
+            val geometriatMap = hankealueService.geometryMapFrom(hanke.alueet)
+            val hankeBeforeUpdate = HankeMapper.domainFrom(hanke, geometriatMap)
+            hanke.vaihe = Hankevaihe.RAKENTAMINEN
+            hanke.version = hanke.version?.inc() ?: 1
+            hanke.modifiedByUserId = userId
+            hanke.modifiedAt = getCurrentTimeUTCAsLocalTime()
+            hanke.generated = false
```
corvidian, line 128: Why is the generated tag updated here?

umeetiusbaar: This comes from HankeService.updateHanke which I took as a rererence for these settings. There, when hanke is updated, the generated flag is set to false. I thought that it should be done here as well.

corvidian: There, the update is a deliberate action taken by the user so the hanke should no longer be considered an automatically generated one. Here, the update is a side-effect of creating an application.

Creating an application for a generated hanke is effectively prohibited in the frontend, so updating it is redundant. If at some point we allow creating a second cable report application using the same generated hanke, we want the hanke to remain as a generated one. So there's no reason to update the tag here.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -109,9 +111,28 @@ class HakemusService(
...
+
+            hankeRepository.save(hanke)
```
corvidian, line 130: Saving the hanke can change it, so the return value should be used for the logging.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeBuilder.kt
```diff
@@ -136,9 +135,7 @@ data class HankeBuilder(
                 Names(
                     firstName = perustaja.etunimi,
                     lastName = perustaja.sukunimi,
-                    givenName = perustaja.etunimi
-                )
-        )
+                    givenName = perustaja.etunimi))
```
corvidian, line 138: I've been fixing these by adding the missing trailing comma.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: Thank you for the clarification! I removed setting the flag here.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: You're right. Changed it.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HankeBuilder.kt
umeetiusbaar: I added some commas, not quite sure where they should be added and where not.

corvidian: If it's a list that spans many rows, a trailing comma should be added. Beyond that, it depends on the situation.

corvidian: ktfmt added automatic trailing commas, but only for the `google` style. I'm almost tempted to adapt it, but that would change nearly every line of code since it's so different, starting from having 2 space indentations instead of 4.

corvidian (approved):

PR #818, opened 2024-09-12
HAI-2625 Make HankeService return domain objects

corvidian:
# Description

Make HankeService return domain objects instead DTOs. Returning DTOs from services violates the separation of concerns between services and controllers.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar (approved):
Good change. Mentioning HankeService is a typo I guess. These changes are about HakemusService.

PR #820, opened 2024-09-13
HAI-2315 Calculate kaivuilmoitus traffic nuisance indexes when the application is updated

umeetiusbaar:
# Description

When kaivuilmoitus is updated, new traffic nuisance indexes are calculated for its work areas. Calculated results are stored in database and returned for API endpoint caller.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
* With Swagger UI call the update endpoint: http://localhost:3001/api/swagger-ui/index.html#/hakemus-controller/update
* You can use e.g. the same geometry which is used in the integration test:  https://github.com/City-of-Helsinki/haitaton-backend/blob/main/services/hanke-service/src/test/resources/fi/hel/haitaton/hanke/geometria/toinen_polygoni.json
* API call should return filled `tormaystarkasteluTulos` for each work areas 

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
@@ -68,6 +68,7 @@ import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withAreas
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withContractor
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withCustomer
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withCustomerWithContactsRequest
+import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withDates
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withName
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withRequiredCompetence
 import fi.hel.haitaton.hanke.factory.HakemusUpdateRequestFactory.withWorkDescription
@@ -112,6 +113,10 @@ import fi.hel.haitaton.hanke.test.AuditLogEntryEntityAsserts.withTarget
 import fi.hel.haitaton.hanke.test.TestUtils
 import fi.hel.haitaton.hanke.test.USERNAME
 import fi.hel.haitaton.hanke.toJsonString
+import fi.hel.haitaton.hanke.tormaystarkastelu.Autoliikenneluokittelu
+import fi.hel.haitaton.hanke.tormaystarkastelu.IndeksiType
+import fi.hel.haitaton.hanke.tormaystarkastelu.LiikennehaittaindeksiType
+import fi.hel.haitaton.hanke.tormaystarkastelu.TormaystarkasteluTulos
 import fi.hel.haitaton.hanke.valmistumisilmoitus.Valmistumisilmoitus
 import fi.hel.haitaton.hanke.valmistumisilmoitus.ValmistumisilmoitusType
 import io.mockk.Called
@@ -1164,6 +1169,7 @@ class HakemusServiceITest(
                             CustomerType.COMPANY, yhteystieto.id, kayttaja.id, newKayttaja.id)
                         .withWorkDescription("New work description")
                         .withRequiredCompetence(true)
+                        .withDates(ZonedDateTime.now(), ZonedDateTime.now().plusDays(1))
                         .withArea(area)
 
                 val updatedHakemus = hakemusService.updateHakemus(hakemus.id, request, USERNAME)
@@ -1332,6 +1338,53 @@ class HakemusServiceITest(
                 assertThat(hankeRepository.findAll().single()).all {
                     prop(HankeEntity::nimi).isEqualTo("New name")
                 }
+            }
+
+            @Test
+            fun `calculates tormaystarkastelu for work areas`() {
+                val hanke = hankeFactory.builder(USERNAME).withHankealue().save()
+                val hankeEntity = hankeRepository.findAll().single()
+                val entity =
+                    hakemusFactory
+                        .builder(USERNAME, hankeEntity, ApplicationType.EXCAVATION_NOTIFICATION)
+                        .saveEntity()
+                val hakemus = hakemusService.getById(entity.id)
+                val area =
+                    createExcavationNotificationArea(
+                        hankealueId = hanke.alueet.single().id!!,
+                        tyoalueet = listOf(createTyoalue(tormaystarkasteluTulos = null)))
+                val request =
+                    hakemus
+                        .toUpdateRequest()
+                        .withDates(ZonedDateTime.now(), ZonedDateTime.now().plusDays(1))
+                        .withArea(area)
+
+                val updatedHakemus = hakemusService.updateHakemus(hakemus.id, request, USERNAME)
+
+                assertThat(updatedHakemus.applicationData)
+                    .isInstanceOf(KaivuilmoitusData::class)
+                    .prop(KaivuilmoitusData::areas)
+                    .isNotNull()
+                    .single()
+                    .prop(KaivuilmoitusAlue::tyoalueet)
+                    .isNotNull()
+                    .single()
+                    .prop(Tyoalue::tormaystarkasteluTulos)
+                    .isNotNull()
+                    .isCorrectTormaystarkasteluTulos()
+
+                val persistedHakemus = hakemusService.getById(updatedHakemus.id)
+                assertThat(persistedHakemus.applicationData)
+                    .isInstanceOf(KaivuilmoitusData::class)
+                    .prop(KaivuilmoitusData::areas)
+                    .isNotNull()
+                    .single()
+                    .prop(KaivuilmoitusAlue::tyoalueet)
+                    .isNotNull()
+                    .single()
+                    .prop(Tyoalue::tormaystarkasteluTulos)
+                    .isNotNull()
+                    .isCorrectTormaystarkasteluTulos()
             }
         }
     }
@@ -2689,6 +2742,28 @@ class HakemusServiceITest(
                 else -> throw IllegalArgumentException()
             }
     }
+}
+
+private fun Assert<TormaystarkasteluTulos>.isCorrectTormaystarkasteluTulos():
+    Assert<TormaystarkasteluTulos> {
+    this.all {
+        prop(TormaystarkasteluTulos::liikennehaittaindeksi).all {
+            prop(LiikennehaittaindeksiType::indeksi).isEqualTo(5.0f)
+            prop(LiikennehaittaindeksiType::tyyppi).isEqualTo(IndeksiType.RAITIOLIIKENNEINDEKSI)
+        }
+        prop(TormaystarkasteluTulos::pyoraliikenneindeksi).isEqualTo(3.0f)
+        prop(TormaystarkasteluTulos::autoliikenne).all {
+            prop(Autoliikenneluokittelu::indeksi).isEqualTo(3.1f)
+            prop(Autoliikenneluokittelu::haitanKesto).isEqualTo(1)
+            prop(Autoliikenneluokittelu::katuluokka).isEqualTo(4)
+            prop(Autoliikenneluokittelu::liikennemaara).isEqualTo(5)
+            prop(Autoliikenneluokittelu::kaistahaitta).isEqualTo(2)
+            prop(Autoliikenneluokittelu::kaistapituushaitta).isEqualTo(2)
+        }
+        prop(TormaystarkasteluTulos::linjaautoliikenneindeksi).isEqualTo(3.0f)
+        prop(TormaystarkasteluTulos::raitioliikenneindeksi).isEqualTo(5.0f)
+    }
+    return this
 }
 
 private fun JohtoselvityshakemusData.setOrdererForCustomer(
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
@@ -17,6 +17,7 @@ import fi.hel.haitaton.hanke.allu.Attachment
 import fi.hel.haitaton.hanke.allu.AttachmentMetadata
 import fi.hel.haitaton.hanke.attachment.application.ApplicationAttachmentService
 import fi.hel.haitaton.hanke.attachment.common.ApplicationAttachmentType
+import fi.hel.haitaton.hanke.daysBetween
 import fi.hel.haitaton.hanke.domain.Hankevaihe
 import fi.hel.haitaton.hanke.email.ApplicationNotificationEmail
 import fi.hel.haitaton.hanke.email.JohtoselvitysCompleteEmail
@@ -35,6 +36,7 @@ import fi.hel.haitaton.hanke.pdf.KaivuilmoitusPdfEncoder
 import fi.hel.haitaton.hanke.permissions.CurrentUserWithoutKayttajaException
 import fi.hel.haitaton.hanke.permissions.HankeKayttajaService
 import fi.hel.haitaton.hanke.toJsonString
+import fi.hel.haitaton.hanke.tormaystarkastelu.TormaystarkasteluLaskentaService
 import fi.hel.haitaton.hanke.valmistumisilmoitus.ValmistumisilmoitusEntity
 import fi.hel.haitaton.hanke.valmistumisilmoitus.ValmistumisilmoitusType
 import java.time.LocalDate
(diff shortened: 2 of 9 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -911,10 +916,32 @@ class HakemusService(
...
 
+    /** Calculate the traffic nuisance indexes for each work area of an application. */
+    private fun updateTormaystarkastelut(hakemusEntityData: HakemusEntityData) {
+        if (hakemusEntityData is KaivuilmoitusEntityData &&
+            hakemusEntityData.startTime != null &&
+            hakemusEntityData.endTime != null) {
+            hakemusEntityData.areas?.forEach { area ->
+                area.tyoalueet.forEach { tyoalue ->
+                    tyoalue.tormaystarkasteluTulos =
+                        tormaystarkasteluLaskentaService.calculateTormaystarkastelu(
```
corvidian, line 932: This method is nested deep. It should be split.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemusalue.kt
```diff
@@ -39,7 +39,7 @@ data class KaivuilmoitusAlue(
 data class Tyoalue(
     val geometry: Polygon,
     val area: Double,
-    val tormaystarkasteluTulos: TormaystarkasteluTulos?,
+    var tormaystarkasteluTulos: TormaystarkasteluTulos?,
```
corvidian, line 42: I wish there were a way to avoid making this class mutable. The change is reflected down the class tree. This is part of the application entity, domain, request, and response data. With this change, we can no longer guarantee the immutability of any of those.

The impact of making the class mutable would be lower if the entity data had separate classes for areas.

Using `.copy()` starting from the HakemusEntityData root is not really viable since that would be messy.

If the tormaystarkastelu could be added in `KaivuilmoitusUpdateRequest.toEntityData()`, copying the areas with modifications would be doable, but we can't call a stateful service from a data class method. All the mapping logic would need to be moved to a new mapper service.

I'm not really sure what would be the best approach, or if we need to make any changes.

corvidian:
`KaivuilmoitusUpdateRequest.hasChanges` compares the areas to see if there are any changes in the request. The tormaystarkastelutulos is a generated value that will be overridden, so it shouldn't be a part of the comparison, especially since the client doesn't need to send the old tormaystarkastelutulos in an update request.

A ticket should be made from this if it's not solved in this PR.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: I splitted this.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemusalue.kt
umeetiusbaar: I reverted this to be `val` again and implemented the update via copy-constructors.

umeetiusbaar:
> `KaivuilmoitusUpdateRequest.hasChanges` compares the areas to see if there are any changes in the request. The tormaystarkastelutulos is a generated value that will be overridden, so it shouldn't be a part of the comparison, especially since the client doesn't need to send the old tormaystarkastelutulos in an update request.
> 
> A ticket should be made from this if it's not solved in this PR.

I updated `KaivuilmoitusUpdateRequest.hasChanges` to ignore changes in `Tyoalue.tormaystarkastelu`.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -1178,7 +1184,7 @@ class HakemusServiceITest(
                             .prop(Hakemusyhteystieto::yhteyshenkilot)
                             .extracting { it.hankekayttajaId }
                             .containsExactlyInAnyOrder(kayttaja.id, newKayttaja.id)
-                        prop(KaivuilmoitusData::areas).isNotNull().single().isEqualTo(area)
+                        prop(KaivuilmoitusData::areas).isNotNull().single()
```
umeetiusbaar, line 1187: Since the request no longer contains tormaystarkastelutulos, we cannot check the equality. For this test I think it is enough to check that there is a single area.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -301,6 +301,54 @@ data class InvoicingPostalAddressRequest(
...
+        this.lisatiedot != area.lisatiedot ||
+        this.name != area.name
+}
+
+@JvmName("hasChangesInTyoalueList")
+fun List<Tyoalue>.hasChanges(workAreas: List<Tyoalue>): Boolean {
+    if (this.size != workAreas.size) {
+        return true
+    }
+    for (i in this.indices) {
```
umeetiusbaar, line 340: Here we assume that if the order of tyoalue differes, then it can be marked as a change.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -301,6 +301,54 @@ data class InvoicingPostalAddressRequest(
...
+    if (this == null) {
+        return areas != null
+    }
+    if (areas == null) {
+        return true
+    }
+    if (this.size != areas.size) {
+        return true
+    }
+    return this.any { it.hasChanges(areas.find { area -> area.hankealueId == it.hankealueId }) }
```
umeetiusbaar, line 315: Checking against `hankealueId` seems like a good choice for finding changes.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/factory/HakemusBuilder.kt
```diff
@@ -208,6 +218,7 @@ data class HakemusBuilder(
                     .withRequiredCompetence()
                     .hakija()
                     .tyonSuorittaja(founder())
+                    .withInvoicingCustomer()
```
umeetiusbaar, line 221: Added this because it seems to be mandatory for sending.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
corvidian: If the list ordering matters, it should matter everywhere.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
corvidian: If the list ordering matters for tyoalueet, it should matter here as well.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
```diff
@@ -301,6 +301,54 @@ data class InvoicingPostalAddressRequest(
...
+}
+
+fun KaivuilmoitusAlue.hasChanges(area: KaivuilmoitusAlue?): Boolean {
```
corvidian, line 318: Could we make copies from the request and the existing areas without the tormaystarkastelutulos and compare those with the built-in compare?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -911,10 +918,67 @@ class HakemusService(
...
+                            hakemusEntityData.areas?.map { area ->
+                                updateTormaystarkastelutForArea(
+                                    area,
+                                    hakemusEntityData.startTime.toLocalDate(),
+                                    hakemusEntityData.endTime.toLocalDate(),
+                                )
+                            },
+                    )
+            }
+        }
```
corvidian, line 943: Nested blobs are hard to read. This can be simplified greatly by inverting the `if`s to guard clauses and using one or two variables for the innermost values.

You could also get fancy and combine the type check and cast in a when-statement:
```kotlin
        val hakemusEntityData =
            when (val data = hakemusEntity.hakemusEntityData) {
                is KaivuilmoitusEntityData -> data
                is JohtoselvityshakemusEntityData -> return
            }
```

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -911,10 +918,67 @@ class HakemusService(
...
+        area.copy(
+            tyoalueet =
+                area.tyoalueet.map { tyoalue ->
+                    updateTormaystarkasteluForTyoalue(
+                        tyoalue,
+                        startDate,
+                        endDate,
+                        area.kaistahaitta,
+                        area.kaistahaittojenPituus,
+                    )
```
corvidian, line 960: I'm not sure how useful the split between these two methods is. The call is almost as long as complicated as the method it's calling.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -911,10 +918,67 @@ class HakemusService(
...
+        kaistahaitta: VaikutusAutoliikenteenKaistamaariin,
+        kaistapituushaitta: AutoliikenteenKaistavaikutustenPituus,
+    ) =
+        tyoalue.copy(
+            tormaystarkasteluTulos =
+                tormaystarkasteluLaskentaService.calculateTormaystarkastelu(
+                    FeatureCollection().apply {
+                        add(Feature().apply { this.geometry = tyoalue.geometry })
+                    },
+                    daysBetween(startDate, endDate)!!,
```
corvidian, line 977: Overloading `daysBetween()` with a non-null version would eliminate the need for `!!`:
```kotlin
fun daysBetween(startDate: Temporal, endDate: Temporal): Int =
    ChronoUnit.DAYS.between(startDate, endDate).toInt() + 1
```

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -2691,6 +2741,28 @@ class HakemusServiceITest(
     }
 }
 
+private fun Assert<TormaystarkasteluTulos>.isCorrectTormaystarkasteluTulos():
```
corvidian, line 2744: You could create the expected object and assert that it equals the actual object.

corvidian: This could also be closer to the call site since it's only used in one method.

umeetiusbaar: You're right. It is much more clear to create the expected object and do a `isEqualTo` comparion against it. I have been reading your brilliant use of `Assert` extensions and wanted to try that and got carried away. I think I do the same with using trailing commas everywhere even when it makes the code less readable.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: Added the overloading.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: Refactored the code a bit. Hopefully it is now more readable.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
umeetiusbaar: Refactored the code a bit. Hopefully it is now more readable.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -911,10 +918,67 @@ class HakemusService(
...
+        endDate: LocalDate,
+        kaistahaitta: VaikutusAutoliikenteenKaistamaariin,
+        kaistapituushaitta: AutoliikenteenKaistavaikutustenPituus,
+    ) =
+        tyoalue.copy(
+            tormaystarkasteluTulos =
+                tormaystarkasteluLaskentaService.calculateTormaystarkastelu(
+                    FeatureCollection().apply {
+                        add(Feature().apply { this.geometry = tyoalue.geometry })
+                    },
```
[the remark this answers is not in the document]
umeetiusbaar: Done.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: I fixed this by overriding equals and hashCode in Tyoalue so that they exclude tormaystarkasteluTulos. Then there is no need to change anything in hasChanges.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestTest.kt
```diff
@@ -165,9 +177,45 @@ class HakemusUpdateRequestTest {
...
+                    .copy(
+                        areas =
+                            listOf(
+                                createExcavationNotificationArea(
+                                    tyoalueet =
+                                        listOf(
+                                            createTyoalue(
+                                                tormaystarkasteluTulos =
+                                                    TormaystarkasteluTulos(
+                                                        TORMAYSTARKASTELU_DEFAULT_AUTOLIIKENNELUOKITTELU,
```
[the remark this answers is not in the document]
umeetiusbaar: Did that. Sorry.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: I got rid off these by overloading equals and hashCode in Tyoalue.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
umeetiusbaar: I got rid off these by overloading equals and hashCode in Tyoalue.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
umeetiusbaar: Fixed this by overloading equals and hashCode of Tyoalue to exclude Tormaystarkastelutulos. Now this can be done as it was with isEqualTo(area).

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequest.kt
corvidian: Overriding the equals and hashCode makes the change global. This might be fine for now, since we don't compare them anywhere else or place them in sets. At some point, this might change. Also, if we add anything to TyoAlue, we need to remember to update the equals and hashCode methods.

The approach I suggested might be easier to maintain. It requires a just couple of changes and it's targeted at this specific problem.

In `KaivuilmoitusUpdateRequest.hasChanges`:
```kotlin
        val areas = areas?.map { it.withoutTormaystarkastelut() }
        val newAreas = applicationData.areas?.map { it.withoutTormaystarkastelut() }
        // ...
            areas != newAreas ||
```

In `KaivuilmoitusAlue`:
```kotlin
    fun withoutTormaystarkastelut() =
        copy(tyoalueet = tyoalueet.map { it.copy(tormaystarkasteluTulos = null) })
```

umeetiusbaar: Plus change HakemusServiceITest to not compare with isEqualTo directly but like this:
```
                        prop(KaivuilmoitusData::areas)
                            .isNotNull()
                            .single()
                            .transform { it.withoutTormaystarkastelut() }
                            .isEqualTo(area.withoutTormaystarkastelut())
```
Did the change.

corvidian (approved):

PR #821, opened 2024-09-13
HAI-2625 Add registry key hidden to customer request

corvidian:
# Description

Add the `registryKeyHidden` field to `CustomerRequest`. The value is read as false if it's null or missing from the request JSON. The value and the registry key are validated to be acceptable.

The logic for updating the registry key or keeping the old value will be added later in a separate PR. Anything related to invoicing customers will be added later as well.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Nothing should be very different yet.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestValidator.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestValidator.kt
@@ -3,13 +3,17 @@ package fi.hel.haitaton.hanke.hakemus
 import fi.hel.haitaton.hanke.allu.CustomerType
 import fi.hel.haitaton.hanke.isValidBusinessId
 import fi.hel.haitaton.hanke.isValidOVT
+import fi.hel.haitaton.hanke.validation.HenkilotunnusValidator.isValidHenkilotunnus
 import fi.hel.haitaton.hanke.validation.ValidationResult
 import fi.hel.haitaton.hanke.validation.ValidationResult.Companion.whenNotNull
 import fi.hel.haitaton.hanke.validation.Validators.given
 import fi.hel.haitaton.hanke.validation.Validators.isBeforeOrEqual
+import fi.hel.haitaton.hanke.validation.Validators.notBlank
 import fi.hel.haitaton.hanke.validation.Validators.notJustWhitespace
 import fi.hel.haitaton.hanke.validation.Validators.notNull
 import fi.hel.haitaton.hanke.validation.Validators.validate
+import fi.hel.haitaton.hanke.validation.Validators.validateFalse
+import fi.hel.haitaton.hanke.validation.Validators.validateNull
 import fi.hel.haitaton.hanke.validation.Validators.validateTrue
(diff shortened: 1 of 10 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestValidator.kt
```diff
@@ -60,18 +60,72 @@ fun CustomerRequest.validateForErrors(path: String): ValidationResult =
...
+        .andWhen(!registryKeyHidden) {
+            whenNotNull(registryKey) { notBlank(it, "$path.registryKey") }
+        }
+
+private fun validateRegistryKeyForPerson(
+    registryKey: String?,
+    registryKeyHidden: Boolean,
+    path: String,
+) =
+    validateRegistryKeyForOther(registryKey, registryKeyHidden, path).whenNotNull(registryKey) {
```
umeetiusbaar, line 104: I see some possibility of errors here: if there is a change in logic in `validateRegistryKeyForOther` this might break. Maybe just copy-paste the logic here or have some common function used in both validations?

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HenkilotunnusValidatorValidatorTest.kt
```diff
@@ -0,0 +1,63 @@
...
+    @ValueSource(
+        strings =
+            [
+                "260570V398", // Missing check digit
+                "080720W84T", // Missing one index digit
+                "140288700X", // Missing separator
+                "24041Y748N", // Missing one date digit
+                "080720W8942T", // Extra digit
+                "190949+016N", // Before 1850
+                "120333B054D", // In the future
```
umeetiusbaar, line 51: Hmm.. is 2033 too little in the future.. :-)

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusUpdateRequestValidator.kt
corvidian: Thank you for the suggestion. This turned out better after I inlined the `validateRegistryKeyForOther` method here. The henkilötunnus validation was done even when not necessary and running both `notBlank` and `isValidHenkilotunnus` was redundant. These also contributed to the double error paths you reported in the other PRs.

Solving this didn't get rid of all the double errors, so I changed the validation to return only unique error paths. Some changes were needed in other parts of the code because of this.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/HenkilotunnusValidatorValidatorTest.kt
corvidian: I would prefer this to be tomorrow, but then I would have to calculate the check digit every time.

I think it's acceptable if this fails once every 10 years :)

umeetiusbaar (approved):

PR #823, opened 2024-09-17
HAI-2625 Don't update registry key when it's hidden

corvidian:
# Description

Don't update registry key when it's hidden.

Also, refactor hakemus tests to use `HakemusBuilder.save()` instead of `.saveEntity()` and a separate load.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create and modify a kaivuilmoitus so there's a person as the customer (hakija). The customer needs to be created with Swagger UI, since the frontend doesn't allow person customers to have registry IDs.
2. Make an update with Swagger UI. Set `registryKeyHidden` to true and `registryKey` to null for the customer.
3. Check that the registry key didn't change.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
@@ -756,6 +756,25 @@ class HakemusControllerITest(@Autowired override val mockMvc: MockMvc) : Control
         }
 
         @Test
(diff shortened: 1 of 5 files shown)


umeetiusbaar (approved):

umeetiusbaar:
Small detail with error messages. When calling the API with `registryKeyHidden: true` and some bogus text in `registryKey` the returned error contained a duplicate path:
`Application contains invalid data. Errors at paths: applicationData.customerWithContacts.customer.registryKey, applicationData.customerWithContacts.customer.registryKey`

PR #827, opened 2024-09-20
HAI-2684 Clear inviter when removing a user for a GDPR request

corvidian:
# Description

When receiving a valid GDPR request to delete a user, clear the `kutsujaId` column of any users the deleted user so the foreign key reference doesn't stop the deletion.

Do this separately for the two cases:
- There are only unidentified users in a hanke besides the deleted user, so the whole hanke is deleted.
- There are identified and/or unidentified users in the hanke, so only the requested user is removed.

When a user requests their information should be deleted from Haitaton, being an inviter to another user is not a valid reason to stop the information from getting deleted.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
From the bug report:
1. Uudella käyttäjällä luo hanke
2. Kutsu sinne muita käyttäjiä, älä käy hyväksymässä kutsuja
3. Pyydä Profiilin UI:sta (tai paikallisesti profiile-gdpr-api-testerillä) käyttäjän tietojen poistoa

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
@@ -335,7 +335,7 @@ class GdprServiceITest(
         }
 
         @Test
-        fun `deletes hanke with applications when there only unidentified users`() {
+        fun `deletes hanke with applications when there are only unidentified users`() {
             val hanke = hankeFactory.builder(USERNAME).withHankealue().saveEntity()
             val founder = hankeKayttajaService.getKayttajaByUserId(hanke.id, USERNAME)!!
             hankeFactory.addYhteystiedotTo(hanke) { omistaja(founder) }
@@ -557,7 +557,7 @@ class GdprServiceITest(
         }
 
         @Test
-        fun `deletes the user sets the inviter to null when the user is an inviter for other users`() {
+        fun `deletes the user and sets the inviter to null when the user is an inviter for other users`() {
             val invitedUserId = "InvitedUser"
             val hanke = hankeFactory.builder(OTHER_USER_ID).withHankealue().saveEntity()
             val kayttaja =
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
@@ -335,6 +335,26 @@ class GdprServiceITest(
         }
 
         @Test
+        fun `deletes hanke with applications when there only unidentified users`() {
+            val hanke = hankeFactory.builder(USERNAME).withHankealue().saveEntity()
+            val founder = hankeKayttajaService.getKayttajaByUserId(hanke.id, USERNAME)!!
+            hankeFactory.addYhteystiedotTo(hanke) { omistaja(founder) }
+            hakemusFactory.builder(hanke).hakija(founder).save()
+            hankekayttajaFactory.saveUnidentifiedUser(
+                hankeId = hanke.id,
+                sahkoposti = "invited@user",
+                kayttooikeustaso = Kayttooikeustaso.HAKEMUSASIOINTI,
+                kutsujaId = founder.id,
+            )
+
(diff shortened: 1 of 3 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
```diff
@@ -334,6 +334,26 @@ class GdprServiceITest(
             assertThat(hakemusRepository.findAll()).isEmpty()
         }
 
+        @Test
+        fun `deletes hanke with applications when there only unidentified users`() {
```
umeetiusbaar, line 338: "when there are only"

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
```diff
@@ -535,6 +555,54 @@ class GdprServiceITest(
                 .extracting { it.permission!!.userId }
                 .containsExactly(HankeKayttajaFactory.FAKE_USERID)
         }
+
+        @Test
+        fun `deletes the user sets the inviter to null when the user is an inviter for other users`() {
```
umeetiusbaar, line 560: "deletes the user and sets..."

umeetiusbaar:
Tested with gdpr-api-tester and a user with a single hanke that had a unidentified contact person. Deleting the user deleted hanke and the unidentieid user as well as the deleted user.

Tested also with a user that had a single hanke which had an identified user. Deleting the user did not delete hanke or the added user. But the added user has only viewing rights to the hanke and therefore cannot do anything with it. I guess this is as it should be?

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
corvidian: Thank you. Typo fixed.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/gdpr/GdprServiceITest.kt
corvidian: Thank you. Typo fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/gdpr/GdprService.kt
```diff
@@ -189,16 +189,17 @@ class GdprService(
             "There are no other users for this hanke, so removing the hanke in response to the GDPR request. ${hanke.logString()}"
         }
 
+        deleteKayttaja(kayttaja)
```
corvidian, line 192: I changed this to reuse the deletion and kutsuja nullification logic.

corvidian:
> Tested also with a user that had a single hanke which had an identified user. Deleting the user did not delete hanke or the added user. But the added user has only viewing rights to the hanke and therefore cannot do anything with it. I guess this is as it should be?

Yes. In this case, we don't have a valid reason to deny the user their right to be deleted, so they need to be deleted. This will be written as an error to the logs, so this should come up in log monitoring. This should be a very rare occurrence.

umeetiusbaar:
> > Tested also with a user that had a single hanke which had an identified user. Deleting the user did not delete hanke or the added user. But the added user has only viewing rights to the hanke and therefore cannot do anything with it. I guess this is as it should be?
> 
> Yes. In this case, we don't have a valid reason to deny the user their right to be deleted, so they need to be deleted. This will be written as an error to the logs, so this should come up in log monitoring. This should be a very rare occurrence.

Ah, now I see the log messsage:
```
{"@timestamp":"2024-09-24T09:53:43.453181361Z","@version":"1","message":"Deleting the last remaining KAIKKI_OIKEUDET user from the hanke. Manual action will be needed to restore full access to the hanke. hankeId=20","logger_name":"fi.hel.haitaton.hanke.gdpr.GdprService","thread_name":"http-nio-8080-exec-3","level":"ERROR","level_value":40000,"userId":"bb6d2d8a-5654-4d15-a71b-faa3536aed29"}
```

umeetiusbaar (approved):

PR #833, opened 2024-10-03
HAI-2803 Add requesting paper decision to send API

corvidian:
# Description

Add an optional body for the application send API. The body can contain the receiver of a paper copy of the decision for the application.

The receiver is saved to the application data and returned from the application APIs. Adding the receiver is written as an update to audit logs.

The receiver can be set only with the send API, it can't be modified with a regular PUT update.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Prepare a hakemus ready for sending
2. Send it using http://localhost:3001/api/swagger-ui/index.html#/hakemus-controller/sendHakemus, add the receiver information to the body.
3. The response should be the receiver information.
4. The receiver information should be in database:
    - `select applicationdata->'paperDecisionReceiver' from applications where id = xx;`
5. The receiver should be in other hakemus requests like `GET /hakemus/xx`. This can be also checked with dev tools in 
    the browser.
7. Everything here should work for both johtoselvityshakemus and kaivuilmoitus.

# Checklist:

 or other location: API docs

# Other relevant info
Sending the paper decision information to Allu will be done separately.


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
```diff
@@ -1132,13 +1136,59 @@ class HakemusControllerITest(@Autowired override val mockMvc: MockMvc) : Control
...
+        }
+
+        @Test
+        fun `returns paper decision receiver when request body is present`() {
+            val paperDecisionReceiver = PaperDecisionReceiverFactory.default
+            val request = HakemusSendRequest(paperDecisionReceiver = paperDecisionReceiver)
+            every {
+                authorizer.authorizeHakemusId(id, PermissionCode.EDIT_APPLICATIONS.name)
+            } returns true
+            every { hakemusService.sendHakemus(id, any(), USERNAME) } returns
```
umeetiusbaar, line 1167: Should here be used `paperDecisionReceiver` as 2nd parameter instead of `any()`?

umeetiusbaar:
Tested with both johtoselvityshakemus and kaivulupailmoitus and also with johtoselvityshakemus without the new information. All worked ok.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
corvidian: I changed this to use `paperDecisionReceiver` to keep up the pattern in other tests.

umeetiusbaar (approved):

PR #834, opened 2024-10-04
HAI-2792 Send paper decision receiver to Allu

corvidian:
# Description

When the orderer asks for a paper of the decision to be sent to an address, send that address to Allu. The address is added to the form data PDF that's sent to Allu after every application.

To make the information easy to notice, add a system comment saying that a paper copy has been requested and that the address can be found in the PDF.

Also,
- Change the customer headers in johtoselvityshakemus PDF to singular, since there can be only one customer for each role.
- Fix the `as` keyword in the Dockerfiles. It should be capitalized.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Create and send an application. Check Allu for the new comment and addition to the form-data PDF.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
@@ -1957,6 +1957,7 @@ class HakemusServiceITest(
             val paperDecisionReceiver = PaperDecisionReceiverFactory.default
             every { alluClient.create(any()) } returns alluId
             justRun { alluClient.addAttachment(alluId, any()) }
+            every { alluClient.sendSystemComment(alluId, any()) } returns 4
             every { alluClient.getApplicationInformation(any()) } returns
                 AlluFactory.createAlluApplicationResponse(alluId)
             auditLogRepository.deleteAll()
@@ -1977,6 +1978,7 @@ class HakemusServiceITest(
             verifySequence {
                 alluClient.create(any())
                 alluClient.addAttachment(alluId, any())
+                alluClient.sendSystemComment(alluId, any())
                 alluClient.getApplicationInformation(alluId)
             }
         }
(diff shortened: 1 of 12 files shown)


umeetiusbaar:
Tested with johtoselvityshakemus and it works as described.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -2002,6 +2005,30 @@ class HakemusServiceITest(
             verifySequence {
                 alluClient.create(any())
                 alluClient.addAttachment(alluId, any())
+                alluClient.sendSystemComment(alluId, any())
+                alluClient.getApplicationInformation(alluId)
+            }
+        }
+
+        @Test
+        fun `sends a system comment to Allu when there is a paper decision receiver`() {
```
umeetiusbaar, line 2014: Where does the name "system comment" come from? In Allu API description they are called "application comments". Okay, now I see it: in Allu UI comments are categorized and external system is one category: "Haitaton järjestelmä 4.10.2024 9.27 Ulkoinen järjestelmä". Btw, "Haitaton järjestelmä" seems a bit faulty as Finnish term - maybe "Haitaton-järjestelmä" could be better? Or is "Haitaton järjestelmä" already a "standard"?

corvidian: It's called a system comment because it's created by Haitaton itself. The same commenting API is envisioned to be used for two-way communication between the applicant and the Allu handlers.

corvidian: I changed the sender to "Haitaton-järjestelmä".

umeetiusbaar (approved):

PR #839, opened 2024-10-08
HAI-2703 Refactor history handling to a separate service

corvidian:
# Description

Hakemusservice is getting pretty big, and this is a convenient independent piece to take out. There are no references between HakemusService and HakemusHistoryService in either direction, so it's safe to separate them.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusHistoryService.kt
```diff
@@ -0,0 +1,168 @@
...
+
+    @Transactional
+    fun handleHakemusUpdates(
+        applicationHistories: List<ApplicationHistory>,
+        updateTime: OffsetDateTime
+    ) {
+        applicationHistories.forEach { handleHakemusUpdate(it) }
+        val status = alluStatusRepository.getReferenceById(1)
+        status.historyLastUpdated = updateTime
+        alluStatusRepository.save(status)
```
umeetiusbaar, line 45: Is explicite save necessary for the entity inside a transaction?

corvidian: I guess it isn't. Removed.

umeetiusbaar (approved):

PR #841, opened 2024-10-09
HAI-2812 Remove henkilotunnus from accompanying johtoselvitys

corvidian:
# Description

When a kaivuilmoitus has a person customer, henkilotunnus is mandatory for them. When an accompanying johtoselvityshakemus is created, the henkilotunnus has been copied to the created application. This causes problems when sending the application to Allu, because johtoselvityshakemus doesn't support registry keys for person customers.

Solve this dilemma by removing the registry keys from any person customers.

Johtoselvityshakemus isn't invoiced and doesn't have invoicing information, so we don't have to care about the registry key there.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
1. Create a kaivuilmoitus with the apply for cable report set on.
2. For the customer (työstä vastaava) select a person and give them a henkilötunnus.
3. Send the application to Allu.
4. The send should succeed.
5. The created johtoselvityshakemus doesn't have a registry key for the customer.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
@@ -276,8 +276,7 @@ class HakemusService(
             hakemus.alluStatus = response.status
         }
 
-        logger.info(
-            "Sent hakemus ${hakemus.applicationType}. ${hakemus.logString()}, alluStatus = ${hakemus.alluStatus}")
+        logger.info("Sent hakemus. ${hakemus.logString()}, alluStatus = ${hakemus.alluStatus}")
         // Save only if sendApplicationToAllu didn't throw an exception
         return hakemusRepository.save(hakemus).toHakemus()
     }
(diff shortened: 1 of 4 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -279,7 +276,8 @@ class HakemusService(
             hakemus.alluStatus = response.status
         }
 
-        logger.info("Sent hakemus. ${hakemus.logString()}, alluStatus = ${hakemus.alluStatus}")
+        logger.info(
+            "Sent hakemus ${hakemus.applicationType}. ${hakemus.logString()}, alluStatus = ${hakemus.alluStatus}")
```
umeetiusbaar, line 280: Would it be good to include applicationType in logString() function?

umeetiusbaar:

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
corvidian: Yes, that would be something we're often interested in when diagnosing an application problem. Added.

umeetiusbaar (approved):

PR #845, opened 2024-10-10
HAI Remove archaic hankeyhteystieto locking system

corvidian:
# Description

There's an old system for disabling edits to hankeyhteystieto. There's a `datalocked` boolean flag in the hankeyhteystieto database table. When it's set to true (by someone manually editing the database), updates to the hanke are refused if they try to change that yhteystieto.

It seems like this system has been born out of a misunderstanding of what data processing restrictions mean in GDPR. It's been interpreted that it means that personal information cannot be edited. In reality, it's a broader concept. If someone asks us to restrict the processing of their personal data, we can't even store it without a valid reason. The GDPR API and the Profiili UI handle this, if someone wants to remove their information, we remove it unless we have a reason why we can't do so.

Since the system is unused and fundamentally flawed, it's best to remove it so it doesn't have to be maintained in the future.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeServiceITests.kt
@@ -829,25 +829,6 @@ class HankeServiceITests(
         }"""
 
     /**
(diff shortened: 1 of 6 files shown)


umeetiusbaar (approved):
Thank you for noticing this and taking good care of the codebase.

PR #847, opened 2024-10-11
HAI-2748 Refactor logging services

corvidian:
# Description

Create an abstract class for the change logging services. They have very similar methods with just a couple of changed parameters, so implementing them only once makes sense. Some have extra methods, so they are left in place. HankeLoggingService has a differing implementation for logging deletes, so it overrides the default.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar (approved):
Really nice refactoring.

PR #848, opened 2024-10-15
HAI-2748 Add API for creating a täydennys

corvidian:
# Description

Add a new API endpoint for creating a response to an information request, or a täydennys. The täydennys will initially have a copy of the application data.

If the application is not in WAITING_INFORMATION status or it doesn't have an open information request (täydennyspyyntö), the endpoint will return 409 Conflict and not do anything else.

Audit logs are written:
- The creation of a new täydennys is reported to the change logs.
- The response might contain personal information, so it's written to disclosure logs when necessary.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Send an application to Allu.
2. In Allu, send an information request about the application.
3. Use the Swagger UI (http://localhost:3001/api/swagger-ui/index.html#/taydennys-controller/create_1) to create a täydennys.
4. Check the response and the database that the täydennys was created.
5. Check the database for new audit logs.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/taydennys/YhteystietoEntity.kt
```diff
@@ -0,0 +1,41 @@
+package fi.hel.haitaton.hanke.taydennys
```
umeetiusbaar, line 1: Why put these interfaces in `taydennys` package? Maybe we could have `taydennys` under `hakemus` package and these common interfaces in `hakemus` package?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/ControllerExceptionHandler.kt
```diff
@@ -157,6 +158,14 @@ class ControllerExceptionHandler {
...
     }
 
+    @ExceptionHandler(HakemusInWrongStatusException::class)
+    @ResponseStatus(HttpStatus.CONFLICT)
+    @Hidden
+    fun hakemusInWrongStatusException(ex: HakemusInWrongStatusException): HankeError {
+        logger.warn(ex) { ex.message }
+        return HankeError.HAI2015
+    }
+
```
umeetiusbaar, line 168: Unrelated to this but maybe we could delete the obsolete exception handling of contact processing restrictions (lines 67-79 + `HankeError.HAI1029` enum):
```
    @ExceptionHandler(HankeYhteystietoProcessingRestrictedException::class)
    // Using 451 (since the restriction is typically due to legal reasons).
    // However, in some cases 403 forbidden might be considered correct response, too.
    @ResponseStatus(HttpStatus.UNAVAILABLE_FOR_LEGAL_REASONS)
    @Hidden
    fun hankeYhteystietoProcessingRestricted(
        ex: HankeYhteystietoProcessingRestrictedException
    ): HankeError {
        logger.warn { ex.message }
        // TODO: the response body SHOULD include an explanation and link to server;
        //  left as future exercise. See https://tools.ietf.org/html/rfc7725
        return HankeError.HAI1029
    }
```

umeetiusbaar:

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/taydennys/YhteystietoEntity.kt
corvidian: I don't really want to create nested packages like that. We'll end up having half of the codebase inside the `hakemus` package. Like `paatos`, `taydennys`, `muutosilmoitus`, `valmistumisilmoitus`, `allu` and `application.attachment` could all be moved / created under `hakemus`.

I combined the interfaces into a single file and moved it under `hakemus`.

PR #850, opened 2024-10-17
HAI-2770 Refactor disclosure logging using an aspect

corvidian:
# Description

Use an `@Aspect` to handle disclosure logging instead of calling it explicitly from controller methods.

I'm worried we haven't always remembered to save the disclosure logs when writing new APIs. It's easy to forget since we need to add an explicit call to `disclosureService` to the controller method. There's also a lot of deduction involved, e.g. we don't run the logging for the return value of create hakemus endpoint, because the endpoint doesn't accept any personal information, so the response can't have any either. But if this assumption changes in the future, it's easy to miss the need to add the logging.

This PR adds an aspect to returning from controller methods to make it implicit. After returning, the return value is checked, and if it's a type that can contain personal information, the value is logged appropriately. The return value is, after all, the deciding factor on what should be logged.

The aspect is attached to any method annotated with `@Operation`. This seemed like the best distinction to find the actual controller methods. We could add it to every public method in classes named `*Controller`, but I think that would include extra things like exception handlers.

To make it impossible to forget, an exception is thrown on unknown types. So when implementing something with a new return value, you have to add it to the logging aspect, either with the logging or without it.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLoggingAspect.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLoggingAspect.kt
@@ -1,0 +1,119 @@
+package fi.hel.haitaton.hanke.logging
+
+import fi.hel.haitaton.hanke.attachment.common.ApplicationAttachmentMetadataDto
+import fi.hel.haitaton.hanke.attachment.common.HankeAttachmentMetadataDto
+import fi.hel.haitaton.hanke.banners.BannerResponse
+import fi.hel.haitaton.hanke.currentUserId
+import fi.hel.haitaton.hanke.domain.Hanke
+import fi.hel.haitaton.hanke.domain.PublicHanke
+import fi.hel.haitaton.hanke.gdpr.CollectionNode
+import fi.hel.haitaton.hanke.hakemus.HakemusDeletionResultDto
+import fi.hel.haitaton.hanke.hakemus.HakemusResponse
+import fi.hel.haitaton.hanke.hakemus.HakemusWithExtrasResponse
+import fi.hel.haitaton.hanke.hakemus.HankkeenHakemuksetResponse
+import fi.hel.haitaton.hanke.permissions.HankeKayttajaController
+import fi.hel.haitaton.hanke.permissions.HankeKayttajaDto
+import fi.hel.haitaton.hanke.permissions.HankeKayttajaResponse
+import fi.hel.haitaton.hanke.permissions.HankekayttajaDeleteService
+import fi.hel.haitaton.hanke.permissions.WhoamiResponse
+import fi.hel.haitaton.hanke.profiili.Names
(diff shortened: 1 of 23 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLoggingAspect.kt
```diff
@@ -0,0 +1,116 @@
...
+                result.taydennys?.let {
+                    disclosureLogService.saveDisclosureLogsForTaydennys(it, userId)
+                }
+            }
+            is Hanke -> disclosureLogService.saveDisclosureLogsForHanke(result, userId)
+            is HankeKayttajaDto ->
+                disclosureLogService.saveDisclosureLogsForHankeKayttaja(result, userId)
+            is HankeKayttajaResponse ->
+                disclosureLogService.saveDisclosureLogsForHankeKayttajat(
+                    result.kayttajat, currentUserId())
```
umeetiusbaar, line 57: copy-paste mishap with using `currentUserId()` instead of `userId`?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLoggingAspect.kt
```diff
@@ -0,0 +1,116 @@
...
+import fi.hel.haitaton.hanke.taydennys.TaydennysResponse
+import fi.hel.haitaton.hanke.tormaystarkastelu.TormaystarkasteluTulos
+import org.aspectj.lang.annotation.AfterReturning
+import org.aspectj.lang.annotation.Aspect
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Component
+
+@Aspect
+@Component
+class DisclosureLoggingAspect(private val disclosureLogService: DisclosureLogService) {
```
umeetiusbaar, line 29: Maybe this class could be unit-tested?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/logging/DisclosureLoggingAspect.kt
```diff
@@ -0,0 +1,116 @@
...
+    @AfterReturning(
+        pointcut = "@annotation(io.swagger.v3.oas.annotations.Operation)", returning = "result")
+    fun logResponse(result: Any?) {
```
umeetiusbaar, line 32: Maybe this function could be unit-tested?

corvidian: I added unit tests. I also added specific exceptions to replace the generic RuntimeExceptions.

umeetiusbaar (approved):

PR #851, opened 2024-10-17
HAI-2820 Add query param for retrieving application areas

umeetiusbaar:
# Description

Add query param 'areas=true/false' for retrieving application areas for all project applications.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
With Swagger UI call GET `/hankkeet/{hankeTunnus}/hakemukset` with and without new query parameter `areas`

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
@@ -479,10 +479,12 @@ class HakemusControllerITest(@Autowired override val mockMvc: MockMvc) : Control
             val response: HankkeenHakemuksetResponse =
                 get("$url?areas=true").andExpect(status().isOk).andReturnBody()
 
-            assertThat(response.applications).isNotEmpty()
-            val expected =
-                HankkeenHakemuksetResponse(hakemukset.map { HankkeenHakemusResponse(it, true) })
-            assertThat(response).isEqualTo(expected)
+            for (i in 0..3) {
+                assertThat(response.applications[i].applicationData.areas)
+                    .isNotNull()
+                    .single()
+                    .isEqualTo(hakemukset[i].applicationData.areas?.single())
+            }
             verifySequence {
                 authorizer.authorizeHankeTunnus(HANKE_TUNNUS, PermissionCode.VIEW.name)
                 hakemusService.hankkeenHakemukset(HANKE_TUNNUS)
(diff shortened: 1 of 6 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
```diff
@@ -462,6 +462,32 @@ class HakemusControllerITest(@Autowired override val mockMvc: MockMvc) : Control
...
+                authorizer.authorizeHankeTunnus(HANKE_TUNNUS, PermissionCode.VIEW.name)
+            } returns true
+
+            val response: HankkeenHakemuksetResponse =
+                get("$url?areas=true").andExpect(status().isOk).andReturnBody()
+
+            assertThat(response.applications).isNotEmpty()
+            val expected =
+                HankkeenHakemuksetResponse(hakemukset.map { HankkeenHakemusResponse(it, true) })
+            assertThat(response).isEqualTo(expected)
```
corvidian, line 485: This doesn't actually test if there are any areas in the response. The `HankkeenHakemusResponse` can ignore the parameter completely and always set the areas to null and the test will still pass.

Maybe check that the response has the same areas as `hakemukset`?

```suggestion
            for (i in 0..3) {
                assertThat(response.applications[i].applicationData.areas)
                    .isNotNull()
                    .single()
                    .isEqualTo(hakemukset[i].applicationData.areas?.single())
            }
```

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HankkeenHakemuksetResponseDeserializer.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonDeserializer
+import com.fasterxml.jackson.databind.node.ObjectNode
+import fi.hel.haitaton.hanke.OBJECT_MAPPER
+import fi.hel.haitaton.hanke.allu.ApplicationStatus
+import java.time.ZonedDateTime
+
+class HankkeenHakemuksetResponseDeserializer : JsonDeserializer<HankkeenHakemuksetResponse>() {
```
corvidian, line 12: You don't have to create a custom deserializer for the whole response, just for the part Jackson can't deserialize without help. The class it's having trouble with is the Hakemusalue interface, but that might not be the best pick, since it's used elsewhere and it has no explicit type information. But you could take a step back and only implement a deserializer for the singular `HankkeenHakemusResponse`.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusControllerITest.kt
umeetiusbaar: Thanks. Fixed this.

services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/hakemus/HankkeenHakemuksetResponseDeserializer.kt
umeetiusbaar: You're right. I changed it as you suggested.

PR #852, opened 2024-10-18
HAI-2749 Use data instead of entity for hakemus update change check

corvidian:
# Description

Compare the hakemus update to a ` HakemusData` instead of an `ApplicationEntity` when checking if the update has any changes.

This allows the same method to be used to check if a täydennys has any changes over the original hakemus or the previously saved täydennys.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Should work like before.

# Checklist:

 or other location:


umeetiusbaar (approved):
Changes make the code more clear and robust.

PR #853, opened 2024-10-21
HAI-2749 Add API for updating täydennys

corvidian:
# Description

Add API for updating a täydennys. The API has most of the same checks and functionalities as the hakemus update. I tried to reuse as much of the code in HakemusService as I could. However, some methods that alter the database entities have to access a specific entity, so they can't be shared easily.

One check removed from the täydennys update is checking if the hakemus has been sent to Allu. It always has. We don't check the hakemus status at all, since the täydennys shouldn't exist unless the hakemus is in a compatible state. Even if it's not due to a bug, there's no harm in updating the täydennys. When sending the täydennys and merging the data back to the hakemus we need to be more careful with the status.

Updating the name of a generated hanke will be done when the täydennys is sent and merged to the hakemus.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create a täydennys.
2. Call the update endpoint in Swagger UI (http://localhost:3001/api/swagger-ui/index.html#/hakemus-controller/update) with different kinds of updates.

# Checklist:

 or other location:


umeetiusbaar (approved):
Tested with `HAI-2747/johtoselvitys-taydennys-form` UI branch and it seems to work okay.

PR #859, opened 2024-10-25
HAI-2749 Refactor hakemus service for reusability

corvidian:
# Description

Make several methods in HankeService public and change their parameters from hakemus-specific objects to more generic data, but only the data they need. E.g., take `HakemusEntityData` instead of a `HakemusEntity`. Or use the generic interface `YhteystietoEntity<H>` to handle customer updates in a compatible way.

These changes allow us to reuse the same methods for validating and executing updates for täydennykset and hopefully for muutosilmoitukset.

Simplify exceptions in hakemus update. Don't worry about passing a hakemusIdentfier through the validating functions. Info about the hakemus we're updating is logged as an info-level log, so it doesn't need to be logged every time there's an error. When examining a problem, the reader can look at the previous lines to find the identifying info about the hakemus.

Otherwise, we would need a new set of exceptions for täydennys and some complex logic for passing the type of exception in the method parameters.

The `errorCode` value in HankeError made ktfmt crash after there were any changes in the file. I changed it to a getter function so that Jackson still serializes it as an `errorCode` field.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/UpdateHakemusITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/UpdateHakemusITest.kt
@@ -51,7 +51,6 @@ import fi.hel.haitaton.hanke.factory.HankeFactory
 import fi.hel.haitaton.hanke.factory.HankeKayttajaFactory
 import fi.hel.haitaton.hanke.findByType
 import fi.hel.haitaton.hanke.firstReceivedMessage
-import fi.hel.haitaton.hanke.hakemus.ApplicationContactType.HAKIJA
 import fi.hel.haitaton.hanke.hakemus.ApplicationContactType.TYON_SUORITTAJA
 import fi.hel.haitaton.hanke.logging.AuditLogRepository
 import fi.hel.haitaton.hanke.logging.ObjectType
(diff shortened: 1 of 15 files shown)


services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/UpdateHakemusITest.kt
```diff
@@ -506,26 +502,55 @@ class UpdateHakemusITest(
...
         }
 
+        @Test
+        fun `throws exception when area references non-existent hankealue`() {
+            val hanke = hankeFactory.builder(USERNAME).withHankealue().save()
+            val hankeEntity = hankeRepository.findAll().single()
+            val hakemus =
+                hakemusFactory
+                    .builder(USERNAME, hankeEntity, ApplicationType.EXCAVATION_NOTIFICATION)
+                    .withWorkDescription("Old work description")
```
umeetiusbaar, line 535: Is work description needed in this test?

corvidian: It's not. The hakija isn't either. I removed them.

umeetiusbaar (approved):

PR #863, opened 2024-11-04
HAI-2759 Add changes in täydennys to hakemus response

corvidian:
# Description

If there are changes in the täydennys when compared to the original hakemus, return the changed fields. The changes are only checked at the top level so any change in e.g. an yhteystieto or in any of it's yhteyshenkilot will tag the yhteystieto as having changed.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create a täydennys and make changes to it.
2. Check that the response for the GET /hakemukset/:id endpoint contains the changed fields.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
@@ -5,11 +5,16 @@ import fi.hel.haitaton.hanke.domain.HasId
 import fi.hel.haitaton.hanke.valmistumisilmoitus.Valmistumisilmoitus
 import fi.hel.haitaton.hanke.valmistumisilmoitus.ValmistumisilmoitusType
 import java.time.ZonedDateTime
+import java.util.UUID
+import kotlin.reflect.KClass
+import kotlin.reflect.KProperty1
 
 enum class ApplicationType {
     CABLE_REPORT,
     EXCAVATION_NOTIFICATION,
 }
+
+private val ZERO_UUID = UUID(0, 0)
 
 data class Hakemus(
     override val id: Long,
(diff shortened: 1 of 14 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
```diff
@@ -52,6 +57,33 @@ data class Hakemus(
...
+
+private fun <D : HakemusData> D.checkChangeInYhteystieto(
+    property: KProperty1<D, Hakemusyhteystieto?>,
+    second: D,
+): String? =
+    if (hasChanges(property.get(this), property.get(second))) {
+        property.name
+    } else null
+
+/** Compare two yhteystieto ignoring differences IDs and yhteyshenkilo IDs. */
```
umeetiusbaar, line 71: "differences in"

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
```diff
@@ -64,12 +96,43 @@ sealed interface HakemusData {
...
+        if (otherAreas != null) {
+            if (areas == null) {
+                changes.addAll(otherAreas.withIndex().map { (i, _) -> "areas[$i]" })
+            } else {
+                otherAreas.withIndex().drop(areas.size).forEach { (i, area) ->
+                    checkChange(area, areas.getOrNull(i)) { "areas[$i]" }?.let { changes.add(it) }
+                }
+            }
+        }
+
```
umeetiusbaar, line 127: This logic might be better to be in its own function perhaps?

Also, the logic is a bit hard to follow for me, but it seems that if areas contains areas A, B and C and otherAreas contains areas D, B, C and F, the result would contains two times "areas[0]"; 1st comes from area A and 2nd from area F after dropping first 3 elements in otherAreas. Is this so and does it matter if it is so?

umeetiusbaar: No, I was wrong. `withIndex` makes sure that the index of the second change is 3.

umeetiusbaar:

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
corvidian: I think `areas.getOrNull(i)` on line 123 is always null since we dropped all common indexes. This means the `changeChange` always returns the given name string. It should be possible to simplify this greatly.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
corvidian: Fixed.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/Hakemus.kt
corvidian: Moving this to a separate method made handling the null lists a lot easier since they can be replaced by empty lists when calling.

I made some other simplifications after realizing that the call to `chckChange` is unnecessary, like directly filtering out the unchanged elements instead of mapping to `String?` and then filtering out the nulls.

The `checkChange` with a function parameter was only called from a single anymore, so I inlined the method.

I think this should be simpler to read now.

umeetiusbaar (approved):

PR #878, opened 2024-11-22
HAI-750 API for adding täydennys attachments

umeetiusbaar:
# Description

POST API for adding täydennys attachments.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. Create a täydennyspyyntö and then a täydennys for a cable report application
2. Add an attachment file with Swagger UI: http://localhost:3001/api/swagger-ui/index.html#/taydennys-attachment-controller/postAttachment
3. Check the database, there should be a row in `taydennys_liite` table
4. Check Azurite blob storage, there should be a file in haitaton-hakemusliitteet-local container inside a folder with täydennys id

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/taydennys/TaydennysAttachmentContentService.kt
```diff
@@ -0,0 +1,49 @@
...
+                .also { logger.info { "Generated blob path: $it" } }
+
+        /**
+         * Prefix derived from täydennys each attachment of that täydennys should have in the cloud
+         * storage.
+         *
+         * Used to distinguish the attachments of different täydennys from each other in the cloud
+         * storage and to enable deleting all of them at once.
+         */
+        fun prefix(taydennysId: UUID): String = "$taydennysId/"
```
corvidian, line 47: The [specs](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/HAI/pages/9033580626/T+ydennyspyynt#POST-%2Ftaydennykset%2F%3Aid%2Fliitteet) say that these attachments should be in the same folder as the attachments of the hakemus. So that when the hakemus is removed, all the attachments related to it are removed as well.

If the folder is for the täydennys, we would need to add logic for deleting the folder to the täydennys removal, but it can also happen when the hakemus or the täydennyspyyntö is removed, so it would get more complicated.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/taydennys/TaydennysAttachmentService.kt
```diff
@@ -0,0 +1,98 @@
...
+                throw e
+            }
+        logger.info {
+            "Added attachment metadata ${newAttachment.id} and content $blobPath for täydennys. ${taydennys.logString()}"
+        }
+        return newAttachment
+    }
+
+    private fun findTaydennys(taydennysId: UUID): TaydennysIdentifier =
+        taydennysRepository.findByIdOrNull(taydennysId)?.toMetadata()
```
corvidian, line 89: `toMetaData()` isn't called under a transaction, but it uses `taydennyspyynto`, which is a lazily loaded field. Is there a risk this might fail at some point?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/taydennys/TaydennysAttachmentContentService.kt
corvidian: It could had been described more clearly. I have a bad habit of being terse.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/taydennys/TaydennysAttachmentService.kt
corvidian: I see. It's clearly in the [javadocs](https://jakarta.ee/specifications/persistence/3.2/apidocs/jakarta.persistence/jakarta/persistence/fetchtype). The persistence provider i.e. Hibarnate has to obey when the fetch type is `EAGER`, but `LAZY` is just a hint. So it can do eager fetches even when the fetch type is `LAZY`. When I turned SQL logging on, it clearly did an eager fetch here.

Since we're excepting the fetch to be eager here, maybe it would be better to set the `taydennyspyynto` relationship to be explicitly `EAGER`?

corvidian: Can't we just return the `TaydennysEntity`, since it already implements the `TaydennysIdentifier` interface?

umeetiusbaar: Interestingly, if I change `TaydennysEntity.taydennyspyynto` to be `EAGER`. it breaks integration tests on the other relationship `yhteystiedot`: 
```
org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: fi.hel.haitaton.hanke.taydennys.TaydennysyhteystietoEntity.yhteyshenkilot: could not initialize proxy - no Session

	at org.hibernate.collection.spi.AbstractPersistentCollection.throwLazyInitializationException(AbstractPersistentCollection.java:636)
	at org.hibernate.collection.spi.AbstractPersistentCollection.withTemporarySessionIfNeeded(AbstractPersistentCollection.java:219)
	at org.hibernate.collection.spi.AbstractPersistentCollection.readSize(AbstractPersistentCollection.java:150)
	at org.hibernate.collection.spi.PersistentBag.size(PersistentBag.java:353)
	at kotlin.collections.CollectionsKt__IterablesKt.collectionSizeOrDefault(Iterables.kt:39)
	at fi.hel.haitaton.hanke.hakemus.YhteystietoEntity$DefaultImpls.toDomain(YhteystietoEntity.kt:47)
	at fi.hel.haitaton.hanke.taydennys.TaydennysyhteystietoEntity.toDomain(TaydennysyhteystietoEntity.kt:22)
	at fi.hel.haitaton.hanke.taydennys.TaydennysEntity.toDomain(TaydennysEntity.kt:52)
	at fi.hel.haitaton.hanke.factory.TaydennysFactory.updateTaydennys(TaydennysFactory.kt:110)
	at fi.hel.haitaton.hanke.factory.TaydennysFactory.updateJohtoselvitysTaydennys(TaydennysFactory.kt:118)
	at fi.hel.haitaton.hanke.taydennys.TaydennysServiceITest$SendTaydennys.sends the data to Allu with list of changed fields(TaydennysServiceITest.kt:386)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

corvidian: We split the attachment services into several services to avoid having the database transaction active while we're uploading the attachment file to the antivirus service and to Blob Storage.

Now you're undoing that work for the täydennys attachment because there's a failing test data factory method. Maybe you could try to fix the factory instead?

corvidian: Looks like the only required changes are adding `@Transactional` to `TaydennysFactory.updateTaydennys()` and `TaydennysFactory.updateJohtoselvitysTaydennys()`.

This breaks the `save()` method. It's probably because it uses a dependency in the parameter default value. The `taydennyspyynto` parameter is not used anywhere so it can be moved to a value in the method body.

umeetiusbaar: Thank you for taking time to advice me in this.

I did the modifications you mentioned except that `TaydennysFactory.updateTaydennys()` does not need to have `@Transactional` because it is only called from `updateJohtoselvitysTaydennys()`.

umeetiusbaar: And `TaydennysEntity.taydennyspyynto` is now `EAGER`.

corvidian: `TaydennysFactory.updateTaydennys()` either needs to be private or have `@Transactional` as well.

umeetiusbaar: Made it private.

corvidian (approved):

PR #880, opened 2024-11-25
HAI-2888 Read verified name form the token for AD users

corvidian:
# Description

Modify the verified name API to handle users who are logged in through Helsinki AD. They don't have a full profile in Helsinki Profiili, so can't call the Profiili API to get a verified name for them. The name we get in the token is reliable, however. It's read from AD, so it's the official the person is known as an employee.

ProfiiliService was created between the controller and the ProfiiliClient to decide how we should get the name of that particular user. We can see from the AMR (Authentication method reference) claim whether the user was authenticated through Suomi.fi or Helsinki AD. The claim is technically an array, but in Profiili there's always a single value. If for some reason the user has both, we default to Suomi.fi method.

Getting a token from the SecurityContext was moved from ProfiiliClient to ProfiiliService since we now have one. I think this is a better separation of concerns. This caused a bunch of changes in tests. Though a lot of the tests in this PR are from ktfmt reformatting closing parentheses and trailing commas.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. When logging in, select City of Helsinki employee.
2. Login with your Gofore email and password.
3. Your name should show up in the user menu after logging in.

# Checklist:

 or other location:


umeetiusbaar (approved):
Tested and it works as described. Code looks good to me. Maybe at some point ProfiiliController and ProfiiliService could be renamed to e.g. VerifiedNameController and Service?

PR #881, opened 2024-11-26
HAI-2794 Don't send an email if täydennyspyyntö not found in Allu

corvidian:
# Description

Currently, if the handler in Allu creates a täydennyspyyntö and cancels it before Haitaton can read it during the next history update, we don't save the täydennyspyyntö, but we still send an email that a täydennyspyyntö was created.

Fix this by only sending the email when the täydennyspyyntö has been found and saved.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
1. Create a johtoselvityshakemus and send it to Allu.
2. In Allu, start to handle the application.
3. Wait for Haitaton to update the application statuses in the background.
4. Create a täydennyspyyntö in Allu and then immediately cancel it.
5. Check [smtp4dev](http://localhost:3003/) that no email was sent about the täydennyspyyntö.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/allu/AlluClientITests.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/allu/AlluClientITests.kt
@@ -372,6 +372,16 @@ class AlluClientITests {
         }
 
         @Test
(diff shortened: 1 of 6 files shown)


umeetiusbaar (approved):

PR #885, opened 2024-11-29
HAI-2794 Send email when a täydennyspyyntö is canceled

corvidian:
# Description

Send emails when a täydennyspyyntö is cancelled. The emails are sent to the hakemus contacts with edit permissions for the hakemus.

Also, refactor the factory methods for ApplicationHistory for easier use.

## Type of change

- [X] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create a täydennyspyyntö in Allu for a hakemus.
2. Wait a minute for Haitaton to get the status update.
3. Cancel the täydennyspyyntö in Allu.
4. Check [Smtp4dev](http://localhost:3003/) for the email.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/resources/email/template/taydennyspyynto-peruttu.subject.mustache
+++ b/services/hanke-service/src/main/resources/email/template/taydennyspyynto-peruttu.subject.mustache
@@ -1,0 +1,1 @@
+Haitaton: Hakemustasi koskeva täydennyspyyntö on peruttu / Hakemustasi koskeva täydennyspyyntö on peruttu / Hakemustasi koskeva täydennyspyyntö on peruttu
--- a/email/taydennyspyynto-peruttu.mjml
+++ b/email/taydennyspyynto-peruttu.mjml
@@ -1,0 +1,44 @@
+<mjml>
+    <mj-head>
+        <mj-include path="common/attributes.partial"/>
+    </mj-head>
+    <mj-body>
+        <mj-include path="common/header-content.partial"/>
+        <mj-section>
+            <mj-column>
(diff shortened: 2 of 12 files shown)


services/hanke-service/src/main/resources/email/template/taydennyspyynto-peruttu.subject.mustache
```diff
@@ -0,0 +1 @@
+Haitaton: Hakemustasi ({{hakemustunnus}}) koskeva täydennyspyyntö on peruttu / Hakemustasi ({{hakemustunnus}}) koskeva täydennyspyyntö on peruttu / Hakemustasi ({{hakemustunnus}}) koskeva täydennyspyyntö on peruttu
```
umeetiusbaar, line 1: This is as it is described in the specs but I noticed that the email for information request does not have `hakemustunnus` in its subject.

corvidian: I think the subject should be as short as possible since it's repeated three times for different languages. I removed the hakemustunnus from the subject line.

umeetiusbaar: Okay. But maybe then the specs should be updated as well?

corvidian: The specs have been updated and they match the current implementation.

umeetiusbaar (approved):

PR #891, opened 2024-12-03
HAI-1331 Remove attachments when täydennyspyyntö is canceled

corvidian:
# Description

When a täydennyspyyntö is cancelled in Allu, remove any täydennys attachment contents from Blob storage.

Also, add missing audit logging for the removal of the täydennys and täydennyspyyntö when they are deleted after the täydennyspyyntö is cancelled in Allu.  Allu is marked as the actor. Even though technically the data is deleted by Haitaton itself, Allu initiates the process by cancelling the täydennyspyyntö.

Add hakemus ID to the täydennyspyyntö domain class so that it's logged as well.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create a täydennys and add attachments to it.
2. Cancel the täydennyspyyntö in Allu.
3. Check that Haitaton removed the attachments with Azure Storage Explorer.
4. Check that the audit log table in the database has entries for the deletion.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/taydennys/TaydennysService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/taydennys/TaydennysService.kt
@@ -119,6 +119,13 @@ class TaydennysService(
         return taydennys
     }
 
+    /**
+     * Delete the täydennyspyyntö related to the given application if the application has one. The
+     * related objects (täydennys and it's attachments, customers and contacts) are removed as well.
+     *
+     * The deletions are logged to have been done by Allu, since this is called from the Allu event
+     * handler.
+     */
     @Transactional
     fun removeTaydennyspyyntoIfItExists(application: HakemusEntity) {
         logger.info {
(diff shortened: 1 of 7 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/taydennys/TaydennysService.kt
```diff
@@ -125,8 +125,17 @@ class TaydennysService(
             "A hakemus has has entered handling. Checking if there's a täydennyspyyntö for the hakemus. ${application.logString()}"
         }
 
-        taydennyspyyntoRepository.findByApplicationId(application.id)?.let {
+        taydennysRepository.findByApplicationId(application.id)?.also {
```
umeetiusbaar, line 128: This function is called `removeTaydennyspyyntoIfItExists` but it does a little more than that. Also, could the name indicate that this is called only by Allu event handler? I cannot give any example of good name candidates though. Maybe adding some comment for the function is sufficient?

corvidian: What more does it do?

umeetiusbaar: It removes täydennys as well.

corvidian: The täydennys is always deleted when the täydennyspyyntö is deleted. The database cascades take care of that. It's done explicitly to log the deletion and delete the attachments.

What makes this limited in use is that it assumes Allu initiated the deletion. Usually, we pass a string for the current user that's used as the actor in the audit logging, but here don't pass anything and log everything to have been done by Allu.

I added a comment to highlight this.

PR #894, opened 2024-12-10
HAI-3326 Use builder pattern for saving test täydennys

corvidian:
# Description

There are now several methods of creating a täydennys for test purposes. There's some overlap between them, so using fewer methods is better.

Replace the TaydennysFactory save methods with the builder pattern since it's the most versatile.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar (approved):
Thank you for keeping the codebase healthy and clean.

PR #897, opened 2024-12-12
HAI-753 Send täydennys attachments to Allu.

umeetiusbaar:
# Description

If there are new attachments added to täydennys, they are sent before sending the application data one-by-one. After each attachment is sent to Allu its metadata is transferred from täydennys to hakemus. This is done in separate transaction for each attachment. This ensures that no data is lost if e.g. sending the application data to Allu fails.

For transfer there is a new separate TaydennysAttachmentTransferService for using transaction propagation REQUIRES_NEW.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
1. Create a täydennys for johtoselvityshakemus
2. Add some attachment(s) to täydennys: http://localhost:3001/api/swagger-ui/index.html#/taydennys-attachment-controller/postAttachment
3. Send täydennys to Allu
4. Check the Allu that the attachment is there
5. Check the database that the attachment metadata has moved from `taydennys_liite` table to `application_attachment` table
6. Check Azure Blob Storage that the content of the attachment is still in place
7. Check that the new attachment can be loaded in application view

# Checklist:

 or other location: 

# Other relevant info


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/Domain.kt
```diff
@@ -87,6 +87,18 @@ data class TaydennysAttachmentMetadata(
             taydennysId = taydennysId,
         )
     }
+
+    fun toAlluAttachment(content: ByteArray): AlluAttachment =
```
corvidian, line 91: To avoid copy-pasting, you could define an interface for AttachmentMetadata with an attachment type, like:
```kotlin
sealed interface AttachmentMetadataWithType : AttachmentMetadata {
    val attachmentType: ApplicationAttachmentType
```
Then if both `ApplicationAttachmentMetadata` and `TaydennysAttachmentMetadata` implement the new interface, `toAlluAttachment()` can be moved to the interface.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/taydennys/TaydennysAttachmentTransferService.kt
```diff
@@ -0,0 +1,39 @@
...
+
+@Service
+class TaydennysAttachmentTransferService(
+    private val taydennysAttachmentRepository: TaydennysAttachmentRepository,
+    private val hakemusAttachmentRepository: ApplicationAttachmentRepository,
+) {
+
+    /** Transfers täydennys attachment to hakemus. */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    fun transferAttachmentToHakemus(
```
corvidian, line 20: Couldn't this be in the metadata service?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/attachment/common/Domain.kt
umeetiusbaar: Good suggestion, thank you! I did just that.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/taydennys/TaydennysAttachmentTransferService.kt
umeetiusbaar: Yes, it could and it would be better place for it. Moved it there.

corvidian (approved):

PR #904, opened 2024-12-18
HAI-1008 AD group checks for authentication

corvidian:
# Description

When a user has authenticated with the Helsinki AD, check that they are in at least one of the groups allowed to access Haitaton. This is done mainly to stop consultants and other non-employees who happen to have access to the Helsinki AD from using Haitaton with that AD login. They should use Suomi.fi instead.

The AD groups are configurable and there's an option to disable the group checks. The disabling is needed, because developers are consultants who don't have the correct AD groups. We can disable the group checks from e.g. the dev environment so we can still test the AD login process.

The constants describing JWT claims and their values were moved from under ProfiiliService to shared objects.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Login to Haitaton with Helsinki AD.
2. If your name shows up at the user icon in the top right corner, backend requests are authenticated.
3. Change `haitaton.ad.filter.use` to true in `application.yml`. Restart the backend.
4. Refreshing Haitaton, your name shouldn't show up anymore since the verified-name endpoint should return 401.
5. Get an AD group from your own bearer token, I like to use https://github.com/mike-engel/jwt-cli for decoding the JWT.
6. Add the AD group to `haitaton.ad.filter.allowed-groups`. Restart the backend.
7. Refreshing Haitaton again, the backend should respond to your queries again, so the name should show up once more.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/resources/application.yml
+++ b/services/hanke-service/src/main/resources/application.yml
@@ -1,4 +1,9 @@
 haitaton:
+  ad:
+    filter:
+      use: ${HAITATON_USE_AD_FILTER:false}
+      # List of allowed groups, seperated by ;
+      allowed-groups: ${HAITATON_ALLOWED_AD_GROUPS:}
   allu:
     baseUrl: ${ALLU_BASEURL:/}
     concurrentUploads: ${ALLU_CONCURRENT_UPLOADS:1}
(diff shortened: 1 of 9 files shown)


services/hanke-service/src/main/resources/application.yml
```diff
@@ -1,4 +1,9 @@
 haitaton:
+  ad:
+    filter:
+      use: ${HAITATON_USE_AD_FILTER:false}
+      # List of allowed groups, seperated by ;
+      allowed-groups: ${HAITATON_ALLOWED_AD_GROUPS:}
```
umeetiusbaar, line 6: Should there also be envs for these new properties in docker-compose.yml for haitaton-hanke:
```
      HAITATON_USE_AD_FILTER: ${HAITATON_USE_AD_FILTER:-false}
      HAITATON_ALLOWED_AD_GROUPS: ${HAITATON_ALLOWED_AD_GROUPS:-}
```

umeetiusbaar (commented):
Tested with docker-compose and it works as described.

services/hanke-service/src/main/resources/application.yml
corvidian: Adding those would allow us to modify the values without rebuilding the backend service, but I don't think we'll need to change these. We don't have the required AD groups anyway.

umeetiusbaar (approved):

PR #909, opened 2024-12-31
HAI-3325 Send a haittojenhallintasuunnitelma PDF to Allu

corvidian:
# Description

When sending a kaivuilmoitus to Allu, also send a haittojenhallintasuunnitelma along with it. There's a separate task for handling the haittojenhallintasuunnitelma for täydennys (HAI-3299).

The PDF is not complete. Currently, it only has basic information about the application.

The typography uses the Liberation Sans font since we can't distribute the Helsinki Grotesk font embedded in a PDF.

Use the new typography and styles also in the form data PDF for johtoselvityshakemus and kaivuilmoitus. The PDF section element was changed to stay together on one page, so the manual page breaks were removed from these reports.

In tests, add a matcher that checks that the name of an Attachment or a list of ApplicationAttachmentMetadata has the correct filename(s). Use the matcher in verifications to make it clear which Allu attachment calls are for what kinds of attachments.

Change the PDF file used in tests to be shorter. When tests fail, they sometimes print the PDF data as a byte array to the console. If the file is long, this can overflow the buffer, so the reason for the failure is lost. With the shorter file, this isn't likely. However, it will still happen with the generated PDF files.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
There's a test that generates a PDF in the local filesystem. This can be used to quickly check the PDF styles.

For more complete testing, send a kaivuilmoitus to Allu. The generated PDF will be sent to Allu as an attachment.

The layouts of form data PDFs have also changed, so create a new johtoselvityshakemus with the kaivuilmoitus to check those attachments as well.

# Checklist:

 or other location:


umeetiusbaar:
Are the margins in the new layout a bit too small? They are significally smaller than in earlier pdf version.

New layout:

image

Old layout:

image

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusServiceITest.kt
```diff
@@ -1214,8 +1218,8 @@ class HakemusServiceITest(
             inner class AccompanyingJohtoselvitys {
 
                 private var kaivuilmoitusHakemusId: Long = 0
-                private val cableReportAlluId: Int = alluId
-                private val excavationNotificationAlluId: Int = alluId + 1
+                private val cableReportId: Int = alluId
```
umeetiusbaar, line 1221: Would `johtoselvitysId` be more suitable as there is `kaivuilmoitusId` now as well in Finnish?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -649,23 +653,7 @@ class HakemusService(
...
-                    val hankealueNames = hankealueet?.associate { it.id to it.nimi } ?: mapOf()
-                    val alueet =
-                        data.areas?.map {
-                            EnrichedKaivuilmoitusalue(
-                                areas[it.hankealueId],
-                                hankealueNames[it.hankealueId] ?: "Nimetön hankealue",
-                                it,
-                            )
-                        }
+                    val alueet = data.areas?.let { enrichKaivuilmoitusalueet(it, hankeTunnus) }
```
umeetiusbaar, line 656: This is now named `alueet` but for johtoselvitys the similar name is still `areas`. Would it be better to have both named as `alueet`?

umeetiusbaar: Well, actually, `areas` in for johtoselvitys means the geometrical area, so, nevermind. It is better to have different names as they are different concepts.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/KaivuilmoitusPdfEncoder.kt
```diff
@@ -150,7 +118,8 @@ object KaivuilmoitusPdfEncoder {
         builder.append("Pinta-ala: ${totalArea.format()} m²\n")
         builder.append("Katuosoite: ${alue.katuosoite}\n")
         builder.append(
-            "Työn tarkoitus: ${alue.tyonTarkoitukset.map { it.format() }.joinToString ( ", " )}\n")
+            "Työn tarkoitus: ${alue.tyonTarkoitukset.map { it.format() }.joinToString ( ", " )}\n"
```
umeetiusbaar, line 121: This could be written as `"Työn tarkoitus: ${alue.tyonTarkoitukset.joinToString(", ") { it.format() }}\n"`

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/PdfHelpers.kt
```diff
@@ -4,37 +4,122 @@ import com.lowagie.text.Chunk
...
+private val sectionTitleFont = Font(baseBoldFont, pxToPt(20), Font.NORMAL, Color.BLACK)
+
+/** Font used for the left header column of a data section. */
+private val rowHeaderFont = Font(baseBoldFont, pxToPt(14), Font.NORMAL, Color.BLACK)
+/** Font used for the left header column of a data section. */
+private val textFont = Font(baseRegularFont, pxToPt(14), Font.NORMAL, Color.BLACK)
+
+fun pxToPt(px: Int): Float = px * PX_TO_PT
 
 fun Document.newline() {
```
umeetiusbaar, line 69: `newline()` is no longer used

corvidian:
> Are the margins in the new layout a bit too small? They are significally smaller than in earlier pdf version.

They are exactly like in Figma. They feel a little cramped, but I'm not an expert in layouts.

umeetiusbaar:
> > Are the margins in the new layout a bit too small? They are significally smaller than in earlier pdf version.
> 
> They are exactly like in Figma. They feel a little cramped, but I'm not an expert in layouts.

You're right, they are as they are suppossed to be.

umeetiusbaar (approved):
Tested with local unit test and with UI and Allu and it works as described.

PR #914, opened 2025-01-07
HAI-3327 Set cableReportDone=true for kaivuilmoitus with accompanying johtoselvityshakemus

umeetiusbaar:
# Description

When sending kaivuilmoitus, set `cableReportDone=true` for kaivuilmoitus with accompanying johtoselvityshakemus.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Other

# Instructions for testing
1. Create and fill kaivuilmoitus with "Hae uusi johtoselvitys" selected
2. Send kaivuilmoitus
3. Check that in response (or in database) `cableReportDone` is `true`

# Checklist:

 or other location: 

# Other relevant info

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
@@ -246,7 +246,8 @@ class HakemusService(
             hakemus.hakemusEntityData =
                 hakemusEntityData.copy(
                     cableReports =
-                        cableReports.plus(sentJohtoselvityshakemus.applicationIdentifier!!)
+                        cableReports.plus(sentJohtoselvityshakemus.applicationIdentifier!!),
+                    cableReportDone = true,
                 )
         }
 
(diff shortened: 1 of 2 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/hakemus/HakemusService.kt
```diff
@@ -246,7 +246,8 @@ class HakemusService(
             hakemus.hakemusEntityData =
                 hakemusEntityData.copy(
                     cableReports =
-                        cableReports.plus(sentJohtoselvityshakemus.applicationIdentifier!!)
+                        cableReports.plus(sentJohtoselvityshakemus.applicationIdentifier!!),
+                    cableReportDone = true,
```
corvidian, line 250: Should this happen 5 lines later, after the application has been sent? The difference would be that whether the user ordered a new johtoselvitys at the same time would be stored in the form data PDF sent to Allu.

umeetiusbaar: Yesn you're right again! Thank you for noticing this. I moved it to its own function after the send.

corvidian (approved):

PR #916, opened 2025-01-08
HAI-3353 Add map of all areas to haittojenhallintasuunnitelma PDF

corvidian:
# Description

Add a map with all työalueet near the top of the PDF. Other maps for the actual haittojenhallintasuunnitelma will be added later.

The base map is fetched from Helsinki Karttapalvelu, by calling their Web Map Service with GeoTools. The areas a rendered on top as polygons, also using GeoTools for the rendering.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Some PDFs with differently shaped work areas are created by `HhsManualTest`. You can also create and send your own kaivuilmoitus and check the attachment in Allu.

# Checklist:

 or other location:


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/HaittojenhallintasuunnitelmaPdfEncoder.kt
```diff
@@ -30,7 +35,18 @@ object HaittojenhallintasuunnitelmaPdfEncoder {
         document.subtitle(data.name)
 
         document.mapHeader("Alueiden sijainti", locationIcon)
-        document.placeholderImage()
+        // The image is better quality, if it's rendered at a higher resolution and then scaled down
```
umeetiusbaar, line 38: Would it be better to have this image generation in its own function?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/MapBounds.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    private fun fixedPoint(a: Double): String = String.format(Locale.UK, "%.3f", a)
+
+    override fun toString(): String = "(x=${fixedPoint(x)}, y=${fixedPoint(y)})"
+}
+
+data class MapBounds(val min: Point, val max: Point) {
+    val xSize: Double = max.x - min.x
+    val ySize: Double = max.y - min.y
+    val center: Point = Point(y = (min.y + max.y) / 2.0, x = (min.x + max.x) / 2.0)
```
umeetiusbaar, line 19: A suggestion:
```
fun Pair<Point, Point>.center(): Point = Point(
    x = (first.x + second.x) / 2,
    y = (first.y + second.y) / 2,
)
```

and then 
`val center: Point = Pair(min, max).center()`

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/MapGenerator.kt
```diff
@@ -0,0 +1,197 @@
...
+
+        renderer.paint(gr, imageBounds, envelope)
+        val baos = ByteArrayOutputStream()
+        ImageIO.write(cropped, "png", baos)
+        map.dispose()
+        return baos.toByteArray()
+    }
+
+    companion object {
+        val blueish: Style by lazy { buildAreaStyle(Color(0, 98, 185)) }
```
umeetiusbaar, line 111: Is the `ish` postfix really necessary for these colours?

umeetiusbaar (commented):
Superb work with this! I tested both with the manual unit test and with UI and Allu. Both worked well and produced nice maps in pdf.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/MapBounds.kt
corvidian: What would be the motivation for this change? This would create a Pair object and instantly throw it away, so there needs to be some benefit to offset the cost.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/MapGenerator.kt
corvidian: Not really necessary. I wanted to distinguish them from `Color.BLUE` etc.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/MapBounds.kt
umeetiusbaar: Maybe you're right and it does not bring enough benefits. If you want to extract the calculation logic it could be done in a function instead of Pair.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/MapGenerator.kt
corvidian: I was adding an enum for the colors in the next task. I added it here already.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/HaittojenhallintasuunnitelmaPdfEncoder.kt
corvidian: I extracted the image handling to a `Document.map()` function. I didn't add the spacer to this function yet. I want to see if I can use the same spacer for the other maps.

umeetiusbaar (approved):

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/MapBounds.kt
corvidian: I extracted a function for calculating the center point.

PR #922, opened 2025-01-13
HAI-3339: Add project area history table and triggers

petri-timlin:
# Description

Added project area history table(=hankealue_historia) and triggers for hankealue, hankegeometria and tormaystarkastelutulos tables. Triggers will add a new row to hankealue_historia table when hankealue, hankegeometria or tormaystarkastelutulos tables have changes.

### Jira Issue: 
https://helsinkisolutionoffice.atlassian.net/browse/HAI-3339

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Other

# Instructions for testing
Run database migration and after that when modifiying or creating "hankealue" objects then there will be new rows in the table "hankealue_historia".

# Checklist:

 or other location: 

# Other relevant info


umeetiusbaar (commented):
Changes only in geometry does not seem to add a new row in `hankealue_historia`. Changing name or other fields of the area adds a new row.

petri-timlin:
> Changes only in geometry does not seem to add a new row in `hankealue_historia`. Changing name or other fields of the area adds a new row.

There should be lines where dml_type='UPDATE_HANKEGEOMETRIA'

umeetiusbaar:
> > Changes only in geometry does not seem to add a new row in `hankealue_historia`. Changing name or other fields of the area adds a new row.
> 
> There should be lines where dml_type='UPDATE_HANKEGEOMETRIA'

Hmm.. I got new rows for one area, but not all. When I change only the geometry. My hanke has 6 areas but it seems that geometry changes only in the first one creates a new row in hankealue_historia. Changes in other fields work okay.

petri-timlin:
> > > Changes only in geometry does not seem to add a new row in `hankealue_historia`. Changing name or other fields of the area adds a new row.
> > 
> > 
> > There should be lines where dml_type='UPDATE_HANKEGEOMETRIA'
> 
> Hmm.. I got new rows for one area, but not all. When I change only the geometry. My hanke has 6 areas but it seems that geometry changes only in the first one creates a new row in hankealue_historia. Changes in other fields work okay.

I have same situation that 6 areas;
image
and then modified all those 6 areas and having this kind results in hankealue_historia table:
image

petri-timlin:
All areas need to be created after having this history functionality enabled in database. If there is areas which have been created before this, those will never get history lines because there is no initial history line where to compare.

umeetiusbaar (approved):
Tested with new Hanke and it works as described.

PR #925, opened 2025-01-15
HAI-3355 Add haittojenhallintasuunnitelma from hankealue to the PDF

corvidian:
# Description

Add haittojenhallintasuunnitelma from hankealue to the PDF

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Send a kaivuilmoitus to Allu and check the attachment in Allu. There's also a temporary test that creates example PDFs.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/HaittojenhallintasuunnitelmaPdfEncoder.kt
+++ b/services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/HaittojenhallintasuunnitelmaPdfEncoder.kt
@@ -6,6 +6,9 @@ import com.lowagie.text.Image
 import com.lowagie.text.ImgTemplate
 import com.lowagie.text.Paragraph
 import com.lowagie.text.Phrase
+import com.lowagie.text.Rectangle
+import com.lowagie.text.pdf.PdfPCell
+import com.lowagie.text.pdf.PdfPTable
 import fi.hel.haitaton.hanke.domain.Haittojenhallintatyyppi
 import fi.hel.haitaton.hanke.domain.Hanke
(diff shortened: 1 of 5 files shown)


services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/HaittojenhallintasuunnitelmaPdfEncoder.kt
```diff
@@ -77,105 +81,164 @@ class HaittojenhallintasuunnitelmaPdfEncoder(private val mapGenerator: MapGenera
             val worstIndexes = kaivuilmoitusalue.worstCasesInTormaystarkastelut()
 
             document.section(hankealue.nimi) {
-                row("Toimet työalueiden haittojen hallintaan, ${hankealue.nimi}", toimetTitle())
-                row("", nuisanceScore(worstIndexes?.liikennehaittaindeksi?.indeksi))
+                hankealueRow(
```
umeetiusbaar, line 84: Could all these sections be in separate functions in order to shorten the `formatPdf` function?

umeetiusbaar (commented):
Tested with UI and Allu and the pdf looks as it should. Really nice!

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/pdf/HaittojenhallintasuunnitelmaPdfEncoder.kt
corvidian: They could, but there would be so many parameters, that the function wouldn't be much shorter.

umeetiusbaar (approved):

PR #929, opened 2025-01-21
HAI-3365 Reset generated hankealueet when cancelling a täydennys

corvidian:
# Description

When cancelling a täydennys on a stand-alone johtoselvityshakemus (so the hanke is generated based on the application), reset the geometries of the hanke to match the original application. The hanke geometries are automatically changed if the user makes changes to the täydennys geometries, so they should be returned to the previous state if the täydennys is then cancelled.

Do this whether the täydennys is cancelled by the user or by Allu.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
1. Create a täydennyspyyntö on a stand-alone johtoselvityshakemus with a generated hanke.
2. Edit the täydennys and add some new work areas.
3. The hanke should be updated with these areas.
4. Cancel the täydennys.
5. The hanke should be showing the original areas, and not the ones added in the täydennys.

# Checklist:

 or other location:


umeetiusbaar (approved):

PR #940, opened 2025-01-31
HAI-2089 Use a mock map generator in tests

corvidian:
# Description

Use a mock map generator in integration tests. The mock always returns a blank PNG image. Depend on a MapGenerator interface instead of the implementation and decide whether to use the mock or the actual implementation with Spring profiles.

Currently, running the tests causes the Haitaton to contact the Helsinki karttapalvelu. This is bad in several ways. The most immediate downside is that the tests take longer since the karttapalvelu takes some seconds to answer each request. Replacing the map generator with a mock service means the karttapalvelu is never called.

Also, remove the disabled tests for manually verifying the Haittojenhallintasuunnitelma PDFs.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other


umeetiusbaar:
Build takes still over 15 mins. There is a PR for increasing timeout: https://dev.azure.com/City-of-Helsinki/haitaton/_git/haitaton-hanke-service-pipelines/pullrequest/10620

umeetiusbaar (approved):

PR #941, opened 2025-02-03
HAI-2089 Protect hakemusalue in hanke udpates

corvidian:
# Description

When updating a hanke, check that all hakemusalue are still inside valid hankealue. For johtoselvityshakemus this means that every hakemusalue needs to be inside some hankealue. For kaivuilmoitus, each työalue is inside the hankealue it's created for.

This also solves HAI-917 and HAI-2091.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Create a hanke with several hankealue. Add a johtoselvityshakemus and a kaivuilmoitus to the hankealue. You should be able to expand the hankealue, but not delete or make them smaller.

# Checklist:

 or other location:


umeetiusbaar (approved):

PR #961, opened 2025-02-18
HAI-3438 Refuse AD tokens without names

corvidian:
# Description

Add a check to AD token validation that checks that the token has non-blank claims for given and family names. Without them, we don't know who the user is and Haitaton can't add them to a hanke.

Also, add a check to getting a user's validated name that checks that the given and family names are not blank. There was already a check that they are present. The check shouldn't come up now that we don't even allow access to AD users without name claims, but I added it as a defence-in-depth thing.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
We don't have an account to test this with, so I trust the unit tests.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/security/AdGroupValidatorTest.kt
+++ b/services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/security/AdGroupValidatorTest.kt
@@ -12,12 +12,16 @@ import fi.hel.haitaton.hanke.test.AuthenticationMocks.TOKEN_VALUE
 import fi.hel.haitaton.hanke.test.USERNAME
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
 import org.junit.jupiter.params.ParameterizedTest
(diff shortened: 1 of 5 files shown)


services/hanke-service/src/test/kotlin/fi/hel/haitaton/hanke/security/AdGroupValidatorTest.kt
```diff
@@ -58,6 +62,47 @@ class AdGroupValidatorTest {
...
+        assertThat(result).hasError("Missing given_name")
+    }
+
+    @ParameterizedTest
+    @MethodSource("emptyNames")
+    fun `fails when AD token has no family name information`(
+        useAdFilter: Boolean,
+        familyName: String?,
+    ) {
+        val validator = validator(true)
```
umeetiusbaar, line 89: Should this be `val validator = validator(useAdFilter)`?

corvidian: Yes, it should. Thank you for noticing. Fixed now.

umeetiusbaar (approved):

PR #964, opened 2025-02-19
HAI Add test API for triggering Allu updates

corvidian:
# Description

Add an API that, when called, immediately retrieves the latest application history from Allu and updates the application statuses accordingly. The update is executed using the same code as the timed updates, ensuring that the last update time is refreshed and that the mutex safeguards against simultaneous updates.

Waiting for the next Allu history update can be frustrating when testing the Haitaton code. This functionality allows us to update and observe the effects in Haitaton instantly.

The endpoint is a GET, making it easy to call. You may use a shell alias that invokes it with cURL, a bookmark in your browser, or a custom shortcut in your IDE.

The endpoint can also be invoked from E2E tests, which often have to wait for a status update from Allu. In the johtoselvitys täydennys test alone, calling this endpoint after each Allu operation saved an average of just over 30 seconds.

This endpoint is only accessible in test environments, similar to other TestData endpoints.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
1. Do something visible in Haitaton after an Allu update, like a decision or a täydennyspyyntö.
2. Call [http://localhost:3001/api/testdata/trigger-allu](http://localhost:3001/api/testdata/trigger-allu)
3. Check Haitaton to see the result.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/configuration/MockAlluUpdateService.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/configuration/MockAlluUpdateService.kt
@@ -1,0 +1,11 @@
+package fi.hel.haitaton.hanke.configuration
+
+import fi.hel.haitaton.hanke.hakemus.AlluUpdateService
+import io.mockk.mockk
+import org.springframework.context.annotation.Bean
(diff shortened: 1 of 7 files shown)


umeetiusbaar (approved):
Works like a charm!

PR #988, opened 2025-03-07
HAI-3455 Add attachment to changed fields with other changes (#986)

corvidian:
# Description

Täydennys changed fields have included attachments only when it's the only change on the täydennys. (Except for when a kaivuilmoitus has changes in its haittojenhallintasuunnitelma, but that is a separate mechanism.)

This means that the handler in Allu might not notice that an attachment has been added, since it's not shown when processing the täydennys.

Fix this by including 'attachment' in the changed fields also when there are other fields.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Other

# Instructions for testing
Replication instructions are on the ticket.

# Checklist:

 or other location:


corvidian:
The changes were approved in #986

PR #997, opened 2025-03-18
HAI-3412 Share attachment creation between services

corvidian:
# Description

Attachment creation has some logic to it that needs to happen in a specific way every way. I.e. the uploaded file needs to be deleted if saving the attachment info to the database fails for some reason. This wasn't done with hanke attachments.

Extract this logic to an interface. Wrap the service calls to abstract interface methods, so the interface doesn't have to know anything about which services are involved. Use the proper logic also for hanke attachments.

Make hanke attachment services return domain objects instead of DTOs. To make the conversion of domain to DTO possible, add hanketunnus to the domain object. The hankeId from the domain object can be removed since it's not used anywhere and the hanketunnus serves the same purpose.

Also, move the `scanAttachment` method to `FileScanClient`, since it was literally identical in every service.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Attachments should still work the same way for hanke, hakemus and täydennyspyyntö.

# Checklist:

 or other location:


umeetiusbaar (approved):
Tested and attachment handling works as before. Code looks good to me.

PR #1009, opened 2025-03-24
HAI-3420 Update public hankkeet API

corvidian:
# Description

- Add all omistajat, but only add their names when they are businesses or organisations. A private person will have null in `organisaatioNimi`.
- Remove `osasto` from public customers, since it's always null for omistajat, the only customer added to public hanke.
- Use the common pattern for the PublicHanke conversion methods.
- Remove the unused audit fields from PublicGeometriat, leaving only the feature collection.
- Remove `tormaystarkasteluTulos` from PublicHanke.
- Add `tormaystarkasteluTulos` to PublicHankeAlue.

Also, remove `tormaystarkasteluTulos` from Hanke. The public hanke was the last using it in backend. It shouldn't be used anymore on frontend side.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Other

# Instructions for testing
Open the public hanke map and look at the response from the network tab of developer tools.

# Checklist:

 or other location:


umeetiusbaar (approved):

PR #1013, opened 2025-03-31
HAI-169 Mark hanke completed periodically

corvidian:
# Description

Every now and then (daily by default), check for any hanke that need to be completed. It needs to be completed, if the end dates of all its areas are gone and it has no open applications.

First, get a list of potential hanke IDs that need to be completed. Try to pre-filter the hanke IDs in the database call as much as possible, so there's no need to go through every hanke in Haitaton. Filtering out the hanke with active applications might be possible, but it would lead to a complicated JPQL call.

Secondly, check each hanke in its own transaction. Check that the hanke passes the requirements for completing i.e. it's public, all of the hankealue are over and that there are no active applications. If it checks all the boxes, it's marked as completed.

Change the name of the `ENDED` status to `COMPLETED`, since that's more descriptive of what the status means now.

Use locking to make sure two instances aren't trying to complete the same group of hanke at the same time.  When getting the lock, wait for it to become available instead of skipping the operation if it's locked like we do with the Allu status updates.

The amount of hanke to check per run and the scheduling can be configured with environment variables, but the defaults should be fine.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Make sure you have a hanke that fits the criteria for completion.
2. Either change the `haitaton.hanke.completions.cron` property to run more often, like `0 * * * * *` for running it every minute, or add `@EventListener(ApplicationReadyEvent::class)` to `completeHankkeet()` to run it at application start-up.
3. Check that the correct hanke were marked as completed.

# Checklist:

 or other location:


umeetiusbaar (approved):
Tested and it works as described. Nice work with the locking and scheduling making it possible to have two instances to run the task for different set of hankes.

PR #1018, opened 2025-04-07
HAI-167 Send reminders when a hanke is ending

corvidian:
# Description

Send reminder emails when a hanke is about to end, 5 and 14 days before the hanke end date. The reminders are sent so the user has a chance to extend the hanke period if the work is still continuing.

The reminders are not sent if the hanke if the end date has already passed. The 14-day won't be sent if the 5-day reminder instead.

The reminders are sent to all hanke users with EDIT permissions to the hanke, i.e. the people who can change the hanke end date.

If a reminder is sent, it's marked as sent on the hanke, so it's not sent again.

Also, update Spring Boot to the latest batch version.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
- Create some hankkeet:
  - One with the end date 6-14 days in the future.
  - One with the end date 1-5 days in the future.
- Add some users with different permissions to those hanke.
- Run the by changing the `reminderCron` value or adding `@EventListener(ApplicationReadyEvent::class)` in front of `sendCompletionReminders`.
- Check [smtp4dev](http://localhost:3003) that the correct emails have been sent. For both hanke, there should be emails to users with edit permissions.
- Check from the DB, that the hanke have their `sent_reminders` set. The hanke with its end date 6-14 days in the future, the column should contain only `COMPLETION_14` and for the other one, it should contain both `COMPLETION_14` and `COMPLETION_5`.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
@@ -283,6 +283,80 @@ class HankeKayttajaServiceITest : IntegrationTest() {
     }
 
     @Nested
+    inner class GtHankeKayttajatWithPermission {
+        @Test
+        fun `returns empty list when there's no hanke `() {
+            val result =
+                hankeKayttajaService.getHankeKayttajatWithPermission(42, PermissionCode.VIEW)
+
+            assertThat(result).isEmpty()
+        }
+
+        @Test
+        fun `returns empty list if no one has the permission`() {
+            val hanke =
+                hankeFactory.builder().saveWithYhteystiedot {
+                    omistaja(Kayttooikeustaso.KATSELUOIKEUS)
+                    rakennuttaja(Kayttooikeustaso.HANKEMUOKKAUS)
+                    toteuttaja(Kayttooikeustaso.KAIKKIEN_MUOKKAUS)
+                }
+            val perustajaPermission =
+                permissionService.findByHankeId(hanke.id).first {
+                    it.kayttooikeustaso == Kayttooikeustaso.KAIKKI_OIKEUDET
+                }
+            permissionService.updateKayttooikeustaso(
+                perustajaPermission,
+                Kayttooikeustaso.KATSELUOIKEUS,
+                USERNAME,
+            )
+
+            val result =
+                hankeKayttajaService.getHankeKayttajatWithPermission(
+                    hanke.id,
+                    PermissionCode.DELETE,
+                )
+
+            assertThat(result).isEmpty()
+        }
+
+        @Test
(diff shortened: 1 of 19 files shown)


corvidian:
> I get an error when trying to start the backend (either by `./gradlew bootRun` or by docker-compose):

I removed the Spring Boot update from this PR. I guess I only ran tests after updating it.

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
```diff
@@ -282,6 +282,80 @@ class HankeKayttajaServiceITest : IntegrationTest() {
         }
     }
 
+    @Nested
+    inner class GtHankeKayttajatWithPermission {
```
umeetiusbaar, line 286: Typo in "GtHankeKayttajat..."?

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
```diff
@@ -182,3 +182,8 @@ enum class TyomaaTyyppi {
     YLEISOTILAISUUS,
     VAIHTOLAVA,
 }
+
+enum class HankeReminder {
```
umeetiusbaar, line 186: I wonder if it would be better to name these enums without number of days in the name? So that if customer wants to change the 1st reminder to be sent when completion is in 21 days instead of 14, we would not need to change database as well as code. Could these be just ordinals as REMINDER_1 (was COMPLETION_14) and REMINDER_2 (was COMPLETION_5)? Of course, if there would be a new reminder between 1 and 2 then we would need to change the database as well. So, never mind, I think this is okay as it is.

umeetiusbaar (commented):

services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/permissions/HankeKayttajaServiceITest.kt
corvidian: Thank you, I fixed the typo.

services/hanke-service/src/main/kotlin/fi/hel/haitaton/hanke/domain/Hanke.kt
corvidian: Yeah, I don't think we can come up with names that would make sense after being redefined. I think if we would need to change the reminder from 14 to 21 days, the correct way would be to add a new reminder to the enum with some logic that we don't send the 21-day reminder if the 14-day reminder has been sent.

Btw, the reminders are named `COMPLETION_X` because there will also be a reminder when the hanke is close to being deleted.

umeetiusbaar (approved):

PR #1020, opened 2025-04-08
HAI-156 Delete hanke after being completed for 6 months

corvidian:
# Description

Once per day, delete all hanke that were completed 6 months ago. The hanke is not deleted if it has an active application for some unexpected reason.

All information related to the hanke is deleted, like users, applications and attachments.

Haitaton is not an archive, so we need to delete old hanke after they are no longer useful. Six months after completion was decided to be a enough time for users to download any decisions and other information that the hanke and its applications might still have.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Complete some hanke like in #1013.
2. Change their `completedAt` value in the database to be in the past.
3. Change the `deletionCron` value or add `@EventListener(ApplicationReadyEvent::class)` to `deleteCompletedHanke`.
4. Check that the hanke were deleted and the `audit_logs` table has rows about the deletion.

# Checklist:

 or other location:

### DIFF
--- a/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeCompletionServiceITest.kt
+++ b/services/hanke-service/src/integrationTest/kotlin/fi/hel/haitaton/hanke/HankeCompletionServiceITest.kt
@@ -721,6 +721,10 @@ class HankeCompletionServiceITest(
                 .save()
             hakemusFactory
                 .builder(hanke, ApplicationType.EXCAVATION_NOTIFICATION)
+                .withNoAlluFields()
+                .save()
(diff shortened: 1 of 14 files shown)


umeetiusbaar:
I tested this with a hanke that has 3 applications:
```
CABLE_REPORT,DECISION
EXCAVATION_NOTIFICATION,FINISHED
CABLE_REPORT,null
```
So, the last application is still a draft without Allu status. This hanke can be marked as completed, but it cannot be deleted because of this last application with `null` status. Is this as it should be?

umeetiusbaar (approved):

PR #1024, opened 2025-04-09
HAI-1150 Add deletion date to hanke responses

corvidian:
# Description

Add the date a completed hanke will be deleted to hanke API responses.

Fix a timezone issue with calculating the deletion date. Calculate it using Helsinki time, so the date doesn't shift backwards if the hanke is completed between midnight and 2-3 am, depending on daylight saving time.

Remove `HankeControllerTest`. It had a unique pattern of being a unit test for a controller, but while being a unit test, it used some Spring injection functionality in the tests. Most of the tests were duplicates of `HankeControllerITest` tests, so they could be deleted without missing anything important. The one validation test didn't have a counterpart in the integration tests, so I added one.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
Check the API response from a completed hanke.

# Checklist:

 or other location:


umeetiusbaar (approved):

PR #1032, opened 2025-04-15
HAI-3447 Complete draft hanke with all end dates

corvidian:
# Description

When a draft hanke has end dates in every hankealue, complete it like a public hanke. If a draft hanke has no areas or some areas don't have end dates, return without doing anything with it.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Other

# Instructions for testing
1. Create a new hanke.
2. On the form, fill out only the hankealue. Add it with dates from the past.
3. Change the `haitaton.hanke.completions.cron` value or add `@Eventlistener(ApplicationReadyEvent::class)` to completeHankkeet.
4. The hanke should be completed.

# Checklist:

 or other location: 

# Other relevant info
There will be another PR about completing drafts without end dates (based on the last time they were updated).


umeetiusbaar (approved):

